PR #7, opened 2021-09-28
Expose the simplest node+Elm HTTP API

n1k0:
This patch is mainly an experimental investigation at this point, it may never be merged, and that's fine.

As the whole application logic + data sources currently reside in the Elm codebase, and users have expressed a need for an HTTP api exposing an equivalent service to what the [Web application](https://mtes-mct.github.io/wikicarbone/#/simulator) currently provides, I'm experimenting creating one reusing that existing code. (Translation: No, I don't really want to rewrite the whole computation logic in js or Python)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,6 @@
 /public/app.js
 /public/main.css
 /public/main.css.map
+/server-app.js
 /xls
+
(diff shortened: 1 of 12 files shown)


n1k0:
I'm gonna land this but NOT leverage the server in production. Some of the Elm refactorings in the patch are very useful, and the server stuff is very lightweight and has no impact on the current webapp.

PR #27, opened 2021-11-16
Compute impact from output mass instead of input mass

magopian:
[User story](https://www.notion.so/8d0c859b065541388c910de2ac854913?v=21369a0f79db4cdcab3ca379ef893ba4&p=e7763b8506cc45fab7c6682d74bd445e)


src/Data/Simulator.elm
```diff
@@ -151,32 +157,29 @@ computeMaterialAndSpinningCo2Score ({ inputs } as simulator) =
...
 computeWeavingKnittingCo2Score : Simulator -> Simulator
-computeWeavingKnittingCo2Score ({ inputs, lifeCycle } as simulator) =
+computeWeavingKnittingCo2Score ({ inputs } as simulator) =
     simulator
         |> updateLifeCycleStep Step.WeavingKnitting
             (\({ country } as step) ->
                 let
                     { kwh, co2 } =
-                        -- NOTE: knitted elec is computed against previous step mass,
-                        -- weaved elec is computed against current step mass
```
magopian, line 176: Special attention here, as I believe the comment was wrong (computed against FOLLOWING step)?

src/Data/LifeCycle.elm
```diff
@@ -60,7 +60,7 @@ getStep label =
 
 getStepMass : Step.Label -> LifeCycle -> Mass
 getStepMass label =
-    getStep label >> Maybe.map .mass >> Maybe.withDefault (Mass.kilograms 0)
+    getStep label >> Maybe.map .inputMass >> Maybe.withDefault (Mass.kilograms 0)
```
n1k0, line 63: Maybe we shouldn't imply getting step mass is necessary the input one? We could probably replace this getter with something more general, eg.:

```elm
-- Data.LifeCycle.elm
getStepProp : Step.Label -> (Step -> a) -> LifeCycle -> Maybe a
getStepProp label prop =
    getStep label >> Maybe.map prop
```

So we could use `lifeCycle |> LifeCycle.getStepProp .inputMass` or `lifeCycle |> LifeCycle.getStepProp .outputMass` explicitely. What do you think?

src/Data/Simulator.elm
```diff
@@ -251,7 +277,13 @@ computeMaterialStepWaste ({ inputs, lifeCycle } as simulator) =
     in
     simulator
         |> updateLifeCycleStep Step.MaterialAndSpinning
-            (\step -> { step | mass = mass, waste = waste })
+            (\step ->
+                { step
+                    | inputMass = mass
+                    , outputMass = Quantity.difference mass waste
+                    , waste = waste
+                }
```
n1k0, line 285: I wonder if we couldn't have some `Step.updateWaste` fn which would be in charge of taking a waste, an input mass and perform the computations and assignments at the Step level? Eg

```elm
simulator
  |> updateLifeCycleStep Step.WeavingKnitting (Step.updateWaste waste mass)
```

n1k0 (commented):
The patch looks good, with a few remarks/suggestions 👍  

We now have much lower scores, so all Simulator tests break, which is expected as we now prevent waste impact to be accounted for twice.

This is especially spectacular for Cashmere which now gives this:

```
↓ Data.SimulatorTest
↓ Data.Simulator
↓ compute
↓ default dyeing weighting
✗ should compute co2 score for manteauCircuitEurope

    516.1201240213679
    ╷
    │ Expect.within Absolute 0.01
    ╵
    2068.4385835709104
```

Let's approve this PR once @PascalDagras and @paulboosz have played around with this patch deployed so we're sure it's consistent with our general expectations.

I'm gonna update the tests to reflect these new scores, so the branch can be deployed to staging.

src/Data/Simulator.elm
n1k0: Addressed in https://github.com/MTES-MCT/wikicarbone/pull/27/commits/acec9efac9dedcfe9932f965ed8f234c40a77532

src/Data/LifeCycle.elm
n1k0: Addressed in https://github.com/MTES-MCT/wikicarbone/pull/27/commits/277585c70ba1f5ae9d17fc11886cc5603eff44ab

src/Data/Step.elm
```diff
@@ -275,7 +294,7 @@ encode v =
         [ ( "label", Encode.string (labelToString v.label) )
         , ( "country", Country.encode v.country )
         , ( "editable", Encode.bool v.editable )
-        , ( "mass", Encode.float (Mass.inKilograms v.mass) )
+        , ( "inputMass", Encode.float (Mass.inKilograms v.inputMass) )
```
n1k0, line 297: I think we should encode `outputMass` as well.

n1k0: Addressed in 6b548eb

n1k0 (approved):

n1k0:
Validated by @paulboosz https://mattermost.incubateur.net/fabnum-mte/pl/eeombjfkm78uukddnbzsr6zpth

PR #35, opened 2021-11-24
Expose step transport processes to both UI and API.

n1k0:
[User Story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=e78a94318c4743c4a84a532a692506a6)

Note: the production API server doesn't reflect this patch, though you can inspect the generated `data/dump-presets.json` which should be available online [here](https://wikicarbone.beta.gouv.fr/branches/attach-step-transport-process-info/data/dump-presets.json) once the PR is built.

### DIFF
--- a/src/Data/LifeCycle.elm
+++ b/src/Data/LifeCycle.elm
@@ -69,28 +69,17 @@ fromQuery db =
 
 init : Inputs -> LifeCycle
 init inputs =
-    initSteps inputs |> update inputs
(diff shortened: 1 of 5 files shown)


paulboosz:
Perfect ! Thanks to this I spotted and fixed a bug in transport in the Gsheet

PR #36, opened 2021-11-26
Fix making step energy & co2 impact computations.

n1k0:
[User Story](https://www.notion.so/Bug-Ne-pas-multiplier-par-la-masse-pour-l-impact-de-l-tape-de-confection-506a72a2075e4b548dac887f18fc14a6)

### Notes

- Once built, this branch will be testable at  https://wikicarbone.beta.gouv.fr/branches/fix-making-step-impact/
- API can't be deployed to test this patch yet

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -29,6 +29,9 @@ jobs:
     - name: Install dependencies
       run: npm ci
 
+    - name: Run static analysis
+      run: npx elm-analyse
+
     - name: Run tests
       run: npm test
 
--- a/src/Data/Formula.elm
+++ b/src/Data/Formula.elm
@@ -128,27 +128,20 @@ dyeingCo2 ( dyeingLowProcess, dyeingHighProcess ) highDyeingWeighting heatCC ele
 
 
 makingCo2 :
-    { makingCC : Co2e
-    , makingElec : Energy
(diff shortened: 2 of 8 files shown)


n1k0:
I've added a section in [the Examples page](https://wikicarbone.beta.gouv.fr/branches/fix-making-step-impact/#/examples) to render the current test suite results :

image

@paulboosz tell me what you think

paulboosz:
Confection step impact is coherent with the fix in the Gsheet :)

But there's something strange in the test results of the examples page : 

When i click on China T shirt example, it says no "teinture majoré" and air transport 
image

But when i click on "Charger" the loaded example has both

image

n1k0:
@paulboosz yeah that's because the tests don't highlight default values; countries have specific default dyeing weigthing values, which are used even if the users don't provide any custom value themselves.

n1k0:
@paulboosz added a note about that:

image

paulboosz:
Yes I understand.
Could we put "Default" instead of "Non" because I find it still a bit misleading

n1k0:
> Could we put "Default" instead of "Non" because I find it still a bit misleading

Great idea, here we go 
image

n1k0:
@magopian yep it's been confirmed that this change only impacts the Making step.

PR #38, opened 2021-11-30
[WIP] Testing the async API

magopian:
We want to make sure that responses that don't take the same time to execute still are returned to the proper request

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,10 +5,13 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "wikicarbone",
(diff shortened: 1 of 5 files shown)


magopian:
Ok, we finally did a simple reference comparison of the handler before and after going through an elm port to make sure it's still exactly the same, and it is:

Capture d’écran 2021-12-01 à 14 48 45

This diff results in the message `equal? true` in the console, which confirms that the handler is the same, and that there's no issues with async requests that way.

Case closed.

PR #42, opened 2021-12-13
Implement all 15 target impacts.

paulboosz:


n1k0:
Thanks @paulboosz, I'm gonna take over the patch to update the code in order to handle the newly introduced trigrams.

n1k0 (commented):
Thanks @paulboosz, I'm gonna take over the patch to update the code in order to handle the newly introduced trigrams.

n1k0 (commented):
@paulboosz Just a note, I see we now have 15 impacts listed, while the PR title mentions we should have 16. Is that expected?

paulboosz:
@n1k0 indeed there's an error in the PR name, should be "add-15-impacts"

n1k0:
We should probably take some time to ensure everything is working file after this patch: https://wikicarbone.beta.gouv.fr/branches/add_16_impacts/#/simulator

magopian (approved):
I glazed over the "search and replace", and had a quick look at the code related to http errors, all looks good to me ;)

n1k0:
I landed this patch, we'll file another PR to implement remaining work and ease reviews.

PR #44, opened 2021-12-20
Multi impacts API endpoints

n1k0:


n1k0:
We now have a basic API server test suite 🎉 
image

PR #49, opened 2021-12-25
Compute and render global PEF score.

n1k0:
[User Story](https://www.notion.so/Calcul-score-PEF-8204b883243246e1a047b9da4bb9f7c4)


src/Data/Unit.elm
```diff
@@ -58,6 +58,13 @@ impactToFloat (Quantity value) =
     value
 
 
+impactPefScore : Impact -> Ratio -> Impact -> Impact
+impactPefScore normalization weighting =
+    Quantity.divideBy (impactToFloat normalization)
+        >> Quantity.multiplyBy (ratioToFloat weighting)
+        >> Quantity.multiplyBy 1000
```
magopian, line 65: Maybe add a comment here explaining why this multiplication?

magopian (approved):
Looks good to me. It does feel a bit weird though, to have an "impact" that's actually just the sum of all the other impacts (with weighting). It feels like it shouldn't be an "impact" but another way to display the impacts? Or maybe just an additionnal result?

Anyway, it's getting more complex and weird by the day :/ Hope it'll still be usable and ergonomic enough for our targetted users.

src/Data/Unit.elm
n1k0: Addressed by 7db578e9990af05cf5cda42a36cf06be041506a0

PR #54, opened 2022-01-03
Use the material's default country even when using the API

magopian:
In the UI, the first step's country is constrained by the material used: it's always going to be the material's default country.
This also needs to be the case when going through the API, as changing the country would lead to incorrect results (the data for the weaving and knitting is only provided for a single country for a given material).

### DIFF
--- a/public/data/openapi.yaml
+++ b/public/data/openapi.yaml
@@ -185,7 +185,7 @@ components:
     countriesParam:
       name: countries
       in: query
-      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit
+      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit (/!\ ATTENTION, pour le moment, le premier pays n'est pas pris en compte car c'est le pays par défaut de la matière première qui est utilisé pour le calcul)
       required: true
       schema:
         type: string
--- a/public/data/openapi.yaml
+++ b/public/data/openapi.yaml
@@ -185,7 +185,7 @@ components:
     countriesParam:
       name: countries
       in: query
-      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit (/!\ ATTENTION, pour le moment, le premier pays n'est pas pris en compte car c'est le pays par défaut de la matière première qui est utilisé pour le calcul)
+      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit (/!\ ATTENTION, pour le moment, le premier pays n'est pas pris en compte car c'est le [pays par défaut](https://fabrique-numerique.gitbook.io/wikicarbone/methodologie/filature#liste-complete) de la matière première qui est utilisé pour le calcul)
       required: true
       schema:
         type: string
--- a/src/Data/Inputs.elm
+++ b/src/Data/Inputs.elm
@@ -50,11 +50,24 @@ type alias CustomCountryMixes =
 
 fromQuery : Db -> Query -> Result String Inputs
 fromQuery db query =
+    let
+        material =
+            db.materials |> Material.findByUuid query.material
+    in
     Ok Inputs
         |> RE.andMap (Ok query.mass)
-        |> RE.andMap (db.materials |> Material.findByUuid query.material)
+        |> RE.andMap material
         |> RE.andMap (db.products |> Product.findById query.product)
-        |> RE.andMap (db.countries |> Country.findByCodes query.countries)
+        |> RE.andMap
+            (material
+                |> Result.andThen
+                    (\{ defaultCountry } ->
+                        query.countries
+                            -- Update the list of countries: the first country (from the material step) is constrained to be the material's default country
+                            |> updateCountryList 0 defaultCountry
+                            |> (\updatedCountries -> Country.findByCodes updatedCountries db.countries)
+                    )
+            )
         |> RE.andMap (Ok query.dyeingWeighting)
         |> RE.andMap (Ok query.airTransportRatio)
         |> RE.andMap (Ok query.recycledRatio)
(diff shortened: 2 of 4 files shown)


public/data/openapi.yaml
```diff
@@ -185,7 +185,7 @@ components:
     countriesParam:
       name: countries
       in: query
-      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit
+      description: Liste des codes pays pour chacune des 5 étapes du cycle de vie du produit (/!\ ATTENTION, pour le moment, le premier pays n'est pas pris en compte car c'est le pays par défaut de la matière première qui est utilisé pour le calcul)
```
n1k0, line 188: I wonder if rapidoc supports Markdown here, so we could add a link to relevant docs?

src/Data/Inputs.elm
```diff
@@ -50,11 +50,24 @@ type alias CustomCountryMixes =
 
 fromQuery : Db -> Query -> Result String Inputs
 fromQuery db query =
+    let
+        material =
+            db.materials |> Material.findByUuid query.material
+    in
```
n1k0, line 56: Why has this been moved in a let bock?

src/Server.elm
```diff
@@ -103,7 +103,7 @@ handleRequest db request =
                 , ( "documentation", Encode.string apiDocUrl )
 
                 -- FIXME: the openapi document should be served by some /openapi API endpoint
-                , ( "openapi", Encode.string "https://wikicarbone.beta.gouv.fr/data/openapi.yaml" )
+                , ( "openapi", Encode.string "/data/openapi.yaml" )
```
n1k0, line 106: Note: Eventually we'll expose the openapi document at the root of the API, so this is super temporary anyway.

src/Data/Inputs.elm
```diff
@@ -50,11 +50,24 @@ type alias CustomCountryMixes =
...
+        |> RE.andMap
+            (material
+                |> Result.andThen
+                    (\{ defaultCountry } ->
+                        query.countries
+                            -- Update the list of countries: the first country (from the material step) is constrained to be the material's default country
+                            |> updateCountryList 0 defaultCountry
+                            |> (\updatedCountries -> Country.findByCodes updatedCountries db.countries)
+                    )
+            )
```
n1k0, line 70: Two things:
- I wonder if this shouldn't/couldn't be extracted to a dedicated function?
- We need to get rid of the duplicate operation performed in Page.Simulator.update when material is changed from the the Web UI: https://github.com/MTES-MCT/wikicarbone/blob/master/src/Page/Simulator.elm#L301.

src/Data/Inputs.elm
```diff
@@ -50,11 +50,24 @@ type alias CustomCountryMixes =
...
         |> RE.andMap (db.products |> Product.findById query.product)
-        |> RE.andMap (db.countries |> Country.findByCodes query.countries)
+        |> RE.andMap
+            (material
+                |> Result.andThen
+                    (\{ defaultCountry } ->
+                        query.countries
+                            -- Update the list of countries: the first country (from the material step) is constrained to be the material's default country
+                            |> updateCountryList 0 defaultCountry
+                            |> (\updatedCountries -> Country.findByCodes updatedCountries db.countries)
```
n1k0, line 68: We may want to use [List.Extra.setAt](https://package.elm-lang.org/packages/elm-community/list-extra/latest/List-Extra#setAt) instead.

src/Data/Inputs.elm
```diff
@@ -104,10 +117,18 @@ setCustomCountryMix index value ({ customCountryMixes } as query) =
     }
 
 
+updateCountryList : Int -> Country.Code -> List Country.Code -> List Country.Code
+updateCountryList index code countryList =
+    countryList
+        |> Array.fromList
+        |> Array.set index code
+        |> Array.toList
```
n1k0, line 125: See previous remark about [List.Extra.setAt](https://package.elm-lang.org/packages/elm-community/list-extra/latest/List-Extra#setAt).

n1k0 (commented):
Looks good, as this is an important behavior, we should probably add a test.

src/Data/Inputs.elm
magopian: We need it line 59 **and** 62

src/Data/Inputs.elm
magopian: Sure thing, I merely reproduced the code that was present before, but indeed that would be a nice shortcut 👍

src/Server.elm
magopian: 👍 
In any case i'd like to keep that change here as it allows seeing the changes when developping locally (otherwise it keeps the production one in the openapi sandbox).

public/data/openapi.yaml
magopian: I added a link to https://fabrique-numerique.gitbook.io/wikicarbone/methodologie/filature#liste-complete

src/Data/Inputs.elm
magopian: Yes I extracted the function.
Regarding the duplicate operation, what exactly are you talking about?

The code in `Simulator.elm` is in charge of resetting the recycled ratio from what I can see

n1k0: Ahh yes my bad, scratch that

src/Data/Inputs.elm
```diff
@@ -11,6 +11,7 @@ import Data.Unit as Unit
 import Json.Decode as Decode exposing (Decoder)
 import Json.Decode.Pipeline as Pipe
 import Json.Encode as Encode
+import List.Extra
```
n1k0, line 14: I'd rather we alias it to `LE` as it's done everywhere else in the codebase

src/Data/Inputs.elm
```diff
@@ -117,12 +109,21 @@ setCustomCountryMix index value ({ customCountryMixes } as query) =
...
 
 
+updatedCountryList : Result String Material -> List Country -> List Country.Code -> Result String (List Country)
+updatedCountryList material countriesDB countries =
+    material
+        |> Result.andThen
+            (\{ defaultCountry } ->
+                countries
+                    -- Update the list of countries: the first country (from the material step) is constrained to be the material's default country
+                    |> updateCountryList 0 defaultCountry
```
n1k0, line 119: How about calling `LE.setAt 0 defaultCountry` directly here? I don't see much added value of the `updateCountryList` alias here…

magopian: It's also called [here](https://github.com/MTES-MCT/wikicarbone/pull/54/files#diff-43b5051efc2b0cb128ab779135733e99a5cc1ddd6a100f03f3896be52c9c7896R132), would you rather I change it to be a direct all in both places?

n1k0: Yes totally 👍

n1k0 (approved):

PR #62, opened 2022-01-17
Upgrade Parcel to v2.2.1, removed Elm hot-reload hacks.

n1k0:
https://github.com/parcel-bundler/parcel/pull/7547 :tada:


n1k0:
I played enough with this patch to claim it's robust enough so we can safely land it.

objarni:
Observations from tested of parcel 2.2.1 with small Elm (https://github.com/objarni/KataDB/ ) with following setup:

1. npm install
2. npm start

Observations (using Sublim Text control+S to save and FireFox browser):
1. Modifying title in index.html hotreloads
2. Modifying src/Main.elm "Filter by tag" hotreloads
3. Modifying src/KataDb.elm "Most recently used" DOES NOT hotreload (but hitting Ctrl+R in FireFox does refresh the page, so rebuilding of the Elm app has happened "behind the scenes")

Hypothesis: hotreloading works for html files and the "first" Elm file (Main.elm), but not dependent .elm files like KataDb.

PR #65, opened 2022-01-19
Add End of life stage

paulboosz:


n1k0:
> I'm proposing to always use this format : "0.00000e-0". I'm able to ensure that all numbers are formatted this way in the future. But for now this create a large noisy commit.

I'm very uncomfortable with this suggestion if we can't find an universal formatting tool everyone could use to ensure formatting JSON data the very same way. I have no tooling available to format JSON the way you suggest, is there some installable library to achieve just that?

n1k0:
@magopian once again, don't be too much scared with the diff stat counts: it's mostly because of updates made to the processes JSON file. I hope we'll eventually figure out a way to avoid such large JSON diffs in the future.

PR #73, opened 2022-02-01
Cache Elm dependencies.

n1k0:
This is an attempt at setting up an Elm dependencies cache strategy in CI & [CD](https://doc.scalingo.com/languages/nodejs/start#custom-cache-folder) contexts.


n1k0:
Note: this will work when [this Github incident](https://elmlang.slack.com/archives/C0CJ3SBBM/p1643710667221799) is resolved and Elm dependency package zipballs are reverted to match their expected checksum.

PR #79, opened 2022-02-07
Update portugal/spain processes

paulboosz:
[User Story](https://www.notion.so/Modification-proc-d-vapeur-pour-Espagne-et-Portugal-72220ed78f8240ae851cc1ee82ec77e0)

Checking Portugal I realized I had already implemented this user story in the gsheet so here it is for the UI


paulboosz:
Spain and Portugal checks for heat are synchronized with Gsheet

PR #104, opened 2022-03-17
Add a "source" url to the impacts

magopian:
Cf https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=da516c13667d4d098ba5d5581ca9ebae

### DIFF
--- a/src/Data/Impact.elm
+++ b/src/Data/Impact.elm
@@ -36,6 +36,7 @@ import Url.Parser as Parser exposing (Parser)
 
 type alias Definition =
     { trigram : Trigram
+    , source : String
     , label : String
     , description : String
     , unit : String
@@ -64,6 +65,7 @@ type alias PefData =
 default : Definition
 default =
     { trigram = defaultTrigram
+    , source = ""
     , label = "Changement climatique"
     , description = "Changement climatique"
     , unit = "kgCO₂e"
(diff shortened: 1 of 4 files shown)


magopian:
This is what it looks like for now

In the simulator:
Capture d’écran 2022-03-17 à 14 49 02

In the explorator:
Capture d’écran 2022-03-17 à 14 48 45

src/Views/Impact.elm
```diff
@@ -45,6 +48,14 @@ impactQuality quality =
     let
         maybeInfo =
             case quality of
+                Impact.NotFinished ->
+                    Just
+                        { cls = "btn-danger"
+                        , icon = Icon.exclamation
```
magopian, line 54: Maybe we'd like another icon there? Not sure how to add it to the icon pack?

src/Data/Impact.elm
```diff
@@ -64,6 +66,7 @@ type alias PefData =
 default : Definition
 default =
     { trigram = defaultTrigram
+    , source = ""
```
n1k0, line 69: We may want to set default source data matching what's in the JSON here

n1k0:
Also I suggest we add a `source` object in the JSON, having a `name` and an `url` properties:

```json
  "source": { "name": "Kering", "url": "https://…" }
```

And same in the Elm side, with a `type alias Source = { name : String, url : String }` attached to the `Definition` record.

n1k0 (approved):
Perfect, ship it 🚀

PR #106, opened 2022-03-23
add new countries

paulboosz:
[User Story](https://www.notion.so/Pays-Sans-attendre-une-int-gration-syst-matique-de-tous-les-pays-ajout-des-principaux-pays-voqu--d73057a3af4c476b9c2d94df0ff65ae8)

@n1k0 I changed impossible route values (eg "road" distance between US and FR for example) from `0` to `null` which make it fails


n1k0:
@paulboosz did you update specific distances from previous country data? tests fails with different computed distances and more annoyingly the app doesn't feature distances in the default simulation (CN->IN)

n1k0:
@paulboosz India has disappeared from the updated transports dataset

n1k0 (changes requested):
India is now missing and should be readded

n1k0 (approved):

PR #110, opened 2022-03-29
Fix broken bookmark+saved simulation links.

n1k0:


n1k0:
@magopian what started as a tiny refactor for modularizing saved simulation kinks view stuff eventually highlighted a bug we currently have in production, preventing saved links and bookmarks to work at all. So the fix is actually in commit https://github.com/MTES-MCT/wikicarbone/pull/110/commits/845cb410e45fc6ace55997e8bc5289248a3bc16a but I thought a while PR with the view modularization plus the fix was nice to have and made sense as a whole. Thanks.

n1k0:
Self-reviewed the patch a dozen times now, I'm landing this.

PR #123, opened 2022-05-03
Contribution/développement sur gitpod

magopian:
A configuration file to develop wikicarbone in gitpod.io (very useful to collaborate and pairprog).
This means that if you open https://gitpod.io/#https://github.com/MTES-MCT/wikicarbone/tree/gitpod-config it will create a workspace automatically configured, on the `gitpod-config` branch (this PR). It also offers to open the code in your locally installed vscode if you prefer (which is very convenient to keep your keyboard shortcuts and local extensions).

If you open it in your local vscode it also adds port forwarding, and the auto-restart of the frontend works as expected (albeit maybe a bit slower, as it goes over ssh ?).

We could give it a try, what do you think @n1k0 ?


magopian:
Just in case you missed it, you can open it in your local vscode app installed on your computer, which makes it very very comfortable imho. I might start using this regularly actually ;)

magopian:
@n1k0 j'ai rajouté un petit paragraphe dans le README, je ne sais pas ce que tu en penses ?

n1k0 (approved):
Nickel 👍

PR #125, opened 2022-05-10
Allow customizing country at Distribution and Use stages.

n1k0:
[User Story](https://www.notion.so/Permettre-la-s-lection-du-pays-d-utilisation-525666f542534a0c95cce5dca18fd024)


n1k0:
No reason to keep this opened as the need for this has been discarded a long while ago now.

PR #135, opened 2022-06-14
[WIP] add the three missing ecotoxicity impacts

magopian:


public/data/impacts.json
```diff
@@ -17,6 +17,10 @@
       "color": "#ff1493",
       "normalization": 5.55695e1,
       "weighting": 0.062
+    },
+    "scope": {
+      "textile": true,
+      "food": true
     }
```
n1k0, line 24: Wondering if we {c|sh}ouldn't go with `"scope": ["textile", "food"]` instead?

src/Views/Impact.elm
```diff
@@ -127,16 +127,20 @@ impactSelector { impacts, selectedImpact, switchImpact } =
                 , value <| Impact.toString trigram
                 ]
                 [ text label ]
+
+        impactsReady =
+            impacts
+                |> List.filter (.scope >> .textile)
```
n1k0, line 133: How about `textileImpacts`?

n1k0 (approved):
LGTM 👍

n1k0:
@magopian I quickly added support for these new impacts to the functional test suite.

magopian:
ah, yes, very good feedbacks, thanks a lot. I gave it a try, what do you think?

src/Data/Impact.elm
```diff
@@ -71,6 +73,15 @@ type alias PefData =
     }
 
 
+type alias Scope =
+    List Target
```
n1k0, line 77: I think you can directly have a `Scope` type:

```elm
type Scope = Food | Textile
```

And a `scopes` property (note the plural form) which would be a `List Scope`.

magopian:
How about that?

PR #146, opened 2022-07-15
Migrate decoders & encoders to use elm-codec.

n1k0:
⚠️ Experimental, do not review/merge.

Most of the few bugs we get is with data encoded to JSON using Elm. This is a know challenge with [regular JSON decoders and encoders](https://package.elm-lang.org/packages/elm/json/latest/) in Elm:

- They often need to match each other perfectly, and synchronizing them must be done manually;
- The compiler is useless checking that encoders aren't misspelling or missing a key.

Enters [elm-codec](https://package.elm-lang.org/packages/miniBill/elm-codec/latest) allowing to define one codec in charge of decoding & encoding data in a single bidirectional definition.


n1k0:
This would ideally need https://github.com/miniBill/elm-codec/pull/15 to be accepted so we could use object field codecs better suited to what was done with previous decoders here.

n1k0:
Tired waiting and don't want to fork and maintain this lib. Closing.

PR #150, opened 2022-08-22
Add a loading spinner to Food PoC page.

n1k0:
We can't track JSON db download progress because elm-http [doesn't allow tracking with task based requests](https://package.elm-lang.org/packages/elm/http/latest/Http#task), so [neither does remotedata-http](https://github.com/ohanhi/remotedata-http/issues/7).

So we're left with just a rolling spinner materializing the (long) loading stage, which is still better than nothing.


magopian (approved):
impressive work ;)

PR #151, opened 2022-08-22
[food] decode duplicated processes in products.json

magopian:
WIP: the impacts percentages aren't properly computed


magopian:
Pair reviewed with @n1k0 👍

PR #156, opened 2022-09-05
Display step details (cont)

magopian:


n1k0:
There's an issue with final product mass, I'm investigating this.

2022-09-12 at 09 19

PR #158, opened 2022-09-14
[food] Use grams as the unit to use for ingredient mass inputs.

n1k0:
I'm finding entering grams much more natural than kilograms for ingredients mass. Thoughts?

Edit: validated with @paulboosz over Mattermost that we can safely round ingredient weight values provided in kg by agribalyse to grams.

image


magopian (approved):
lgtm 👍 
Do we want to switch to INTs instead of FLOATs?

PR #160, opened 2022-09-15
Use elm-watch

magopian:
Replace parcel with [elm-watch](https://discourse.elm-lang.org/t/introducing-elm-watch-elm-make-in-watch-mode-fast-and-reliable/8653)

TODO:
  
  - [ ] update Scalingo env vars: eg `MATOMO_TOKEN` -> `VITE_MATOMO_TOKEN`


magopian:
So, this is now working correctly, BUT:
- no easy live-reloading on sass file change (we could hack something together using something like [this](https://github.com/evanw/esbuild/issues/802#issuecomment-819578182) but it's hairy)
- no easy way to use env vars (we need to use esbuild's [define](https://esbuild.github.io/api/#define) i guess)

In the current context, I'd say that elm-watch works well, but esbuild is less than ideal (at least for now, for us).

magopian:
Ok, this should now be up for review @n1k0 , would you like to take it for a spin?
It seems to me that everything works as expected, bar one "detail": the scalingo build times out for some reason... The server SHOULD be running on the proper port and host, from my local tests, but still it fails. Do you have an idea what could cause that?

lib/index.js
```diff
@@ -37,7 +37,7 @@ function setupTracker(matomoUrl, token, spec) {
 
   return {
     track(status, req) {
-      if (process.env.NODE_ENV !== "production") {
+      if (process.NODE_ENV !== "production") {
```
lydell, line 40: Are you sure this change is correct? I’ve never seen `process.NODE_ENV` before, while `process.env.NODE_ENV` is common.

postprocess.mjs
```diff
@@ -0,0 +1,100 @@
...
+ */
+export default async function postprocess({ code, targetName, compilationMode }) {
+  switch (compilationMode) {
+    case "standard":
+    case "debug":
+      return code;
+
+    case "optimize":
+      return minify(code, {
+        minimal: !["Html", "Sandbox", "Element", "Document", "Worker"].includes(targetName),
```
lydell, line 16: This needs to be adjusted for the target names in your elm-watch.json. The thing here is that `minimal: true` gives smaller output but takes much longer. For a customer facing app that can be worth it, but for an internal app size might not matter as much and it’s more worth to have a speedy setup.

build/main.js
```diff
@@ -0,0 +1,1889 @@
+// elm-watch hot {"version":"1.0.2","webSocketPort":59209}
```
lydell, line 1: This file isn’t supposed to be committed, is it?

lydell (commented):
Really cool to see elm-watch used together with Vite!

lib/index.js
magopian: yeah, well, oops 😬 my bad, thanks a lot for spotting it!

build/main.js
magopian: well spotted once again, my bad, i removed it 👍

.gitignore
```diff
@@ -7,12 +7,12 @@
 /public/icomoon/*.txt
 /public/icomoon/*.zip
 /public/icomoon/demo-files
-/public/icomoon/demo-files
```
n1k0, line 10: Why removing this line? We don't want to version icomoon HTML demo files…

package.json
```diff
@@ -6,36 +6,32 @@
...
     "server:build": "elm make src/Server.elm --optimize --output=server-app.js",
     "server:dev": "npm run server:build && nodemon server.js --config nodemon.json",
-    "server:debug": "elm make src/Server.elm --output=server-app.js && nodemon server.js --config nodemon.json",
+    "server:debug": "elm-watch make server && nodemon server.js --config nodemon.json",
     "server:start": "npm run server:build && node server.js",
     "server:test": "npm run server:build && jest",
-    "start:parcel": "parcel serve index.html --no-cache",
-    "start:dev": "npm run build:init && concurrently -n \"watch,parcel,server\" -c \"green,cyan\" \"npm run start:parcel\" \"npm run server:dev\"",
-    "start": "PARCEL_ELM_NO_DEBUG=1 npm run start:dev",
+    "start": "run-pty run-pty.json",
```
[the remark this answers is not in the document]
n1k0: Also, running this command doesn't tell me at what URL the app is served; we should also document this in README

README.md
```diff
@@ -70,8 +70,8 @@ Chaque _Pull Request_ effectuée sur le dépôt est également automatiquement d
 
 Certaines variables d'environnement peuvent ou doivent être configurées via l'interface de [configuration Scalingo](https://dashboard.scalingo.com/apps/osc-fr1/wikicarbone/environment) :
 
-- `SENTRY_DSN`: le DSN [Sentry](https://sentry.io) à utiliser pour les rapports d'erreur.
-- `MATOMO_TOKEN`: le token [Matomo](https://stats.data.gouv.fr/) permettant le suivi d'audience de l'API.
+- `VITE_SENTRY_DSN`: le DSN [Sentry](https://sentry.io) à utiliser pour les rapports d'erreur.
+- `VITE_MATOMO_TOKEN`: le token [Matomo](https://stats.data.gouv.fr/) permettant le suivi d'audience de l'API.
```
n1k0, line 74: Nit: It's annoying these must be prefixed; no way to avoid it?

n1k0 (commented):
Several remarks at this point:

- Port used for dev server is 5173; this should be updated in README (or couldn't we juyst set it up to use 1234 as before?)
- `elm-watch` seems young and promising, but also prone to changes in a near future; wouldn't we want it to settle a little before leveraging it in production?
- The API page is broken
image
- matter of personal taste, but I really appreciated the all-inclusive, configuration-less approach of Parcel; in this patch, I find the configuration files are a little hard to understand

lydell:
> young and promising, but also prone to changes in a near future

Worked on it for 1.5 years, used it in production for 5 months. No changes are planned, only enhancements.

n1k0:
This branch has diverged much while we'r epretty much happy with Parcel atm, so I'm gonna close this one.

PR #161, opened 2022-09-19
[food] Render transport distances by mode between steps.

n1k0:
2022-09-19 at 16 56

Naive yet efficient approach. Also I think it would be good to start moving computations made in View land to Data land (I'm uncomfortable having a `formatAmount` fn along this new `getStepTransport` one, though using only one would be rather complicated here).


magopian (approved):
Very good work, 👍 
🚢

PR #162, opened 2022-09-23
Fix final weight calculation (an original recipe's final weight is always 1kg)

magopian:
An original recipe's final weight is always 1kg (eaten), so the impact "per kg" and "total" on an original recipe should always be the same.

When displaying an original recipe, don't display the "total impact" as it's redundant.

https://user-images.githubusercontent.com/167767/191928563-067427af-da2c-4fa8-b238-3f6fddeabd23.mov

Also display the final weight after the "at consumer" step

Capture d’écran 2022-09-23 à 11 24 16

### DIFF
--- a/src/Data/Food/Product.elm
+++ b/src/Data/Food/Product.elm
@@ -17,6 +17,7 @@ module Data.Food.Product exposing
     , emptyProducts
     , findProductByName
     , formatItem
+    , getAmountRatio
     , getMainItemComment
     , getStepImpact
     , getStepTransports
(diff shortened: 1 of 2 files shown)


magopian:
The diff should be rather straightforward to review @n1k0, what do you think?

@paulboosz this is what you meant, right? Are the computations correct now?

n1k0:
Slightly improved rendering of final step weight in https://github.com/MTES-MCT/ecobalyse/pull/162/commits/017ac8d8895795b710452801cff105fe8f3cc261

image

n1k0 (approved):
LGTM, great job 👍

paulboosz:
Looks good to me too !

magopian:
Awesome, thanks for the reviews, and thanks for the modifications @n1k0 it looks good 🤩

PR #163, opened 2022-09-28
Add a new TEA-wired page for building recipes.

n1k0:


magopian (approved):
Excellent! I like the two buttons that link from one page to the other, nice touch 👍

PR #169, opened 2022-10-21
Decode the new "kind" key in processes.json

magopian:


src/Data/Food/Process.elm
```diff
@@ -136,7 +144,7 @@ decodeProcess definitions =
         |> Pipe.required "impacts" (Impact.decodeImpacts definitions)
         |> Pipe.required "unit" decodeStringUnit
         |> Pipe.required "simapro_id" (Decode.map codeFromString Decode.string)
-        |> Pipe.required "category" decodeCategory
+        |> Pipe.required "kind" decodeCategory
```
n1k0, line 147: It's a bit strange to decode a `kind` key with a `decodeCategory` decoder while we still have a `category` key we don't consume anymore. This looks super confusing for a maintainer looking at the JSON file and decoders.

Do we ever plan to use the old `category` values? I suggest we simply get rid of the old `category` field values and replace them with our new values. If we ever discover that we want agribalyse category on the client, we can always reintroduce an `agribalyse_category` field.

n1k0: I realize that this is basically what existed before https://github.com/MTES-MCT/ecobalyse-data/commit/dce08a138aa271ee3995d6d07ca42f9ed8852b42, but yeah, what I'm basically suggesting is to revert the changes in the JSON export introduced since this commit.

n1k0:
Superseded by #171

PR #175, opened 2022-11-14
Update the ingredient selector to use the new limited ingredients from ingredients.json

magopian:
[User story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=7307e444cbce4545b700b55677a9049e&pm=s)


magopian:
@n1k0 if you want to have a look? Disclaimer: not very happy about the current state of the code, because i'm not sure how much to optimize it (and factor it), as I guess we might have more labels in the future, and maybe also use such a "checkbox for labels" for other steps like packaging? Something like "recycled cardboard" or the like?

src/Data/Food/Ingredient.elm
```diff
@@ -0,0 +1,59 @@
...
+
+decodeIngredients : List Process -> Decoder (List Ingredient)
+decodeIngredients processes =
+    let
+        processesDict =
+            processes
+                |> List.map (\process -> ( Process.codeToString process.code, process ))
+                |> Dict.fromList
+    in
+    Decode.list (decodeIngredient processesDict)
```
n1k0, line 33: Nit: I know you don't like it much but this can be written as

```elm
decodeIngredients : List Process -> Decoder (List Ingredient)
decodeIngredients =
    List.map (\process -> ( Process.codeToString process.code, process ))
        >> Dict.fromList
        >> decodeIngredient
        >> Decode.list
```

Up to you

n1k0 (commented):
Some things remaining to do :

- [x] I think it's good idea to remove the `tunaPizza` sample query in `Data/Food/Inputs.elm` to rather use a new `carrotCake` sample featuring our existing ingredients.

n1k0:
Sorry, realized I've submitted the pending review inadvertently… I'm a bit puzzled atm with the large chunks of copied/pasted code (`ingredientSelectorView`, `addIngredientFormView`); Ideally we'd want to be able to abstract these two think to either deal with ingredients or "other things", them being regular processes or other specific data structures referring to them (packaging, etc)

n1k0:
Also, I think the API should be updated https://ecobalyse-pr175.osc-fr1.scalingo.io/api/food/ingredients

src/Page/Food/Builder.elm
```diff
@@ -299,6 +326,105 @@ addProcessFormView { category, defaultMass, excluded, foodDb, kind, noOp, select
         ]
 
 
+type alias AddIngredientConfig msg =
+    { defaultMass : Mass
+    , excluded : List Process.Code
+    , foodDb : FoodDb.Db
+    , isOrganic : Bool
```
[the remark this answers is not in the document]
magopian: Yes we'll have more labels, so i think we need to switch to a union type 👍

src/Page/Food/Builder.elm
```diff
@@ -100,9 +104,21 @@ update : Session -> Msg -> Model -> ( Model, Session, Cmd Msg )
...
+                processCode =
+                    if model.isIngredientOrganic then
+                        session.foodDb.ingredients
+                            |> List.filter (.conventional >> .code >> (==) code)
+                            |> List.head
+                            |> Maybe.map (.labels >> .organic >> Maybe.map .code >> Maybe.withDefault code)
+                            |> Maybe.withDefault code
+
+                    else
+                        code
```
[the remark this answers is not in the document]
magopian: This `if` will be replaced by a `case` as soon as I rework the format of the `ingredients.json` file to have all the different labels at the same level, instead of treating the "conventional/default"  differently.

src/Page/Food/Builder.elm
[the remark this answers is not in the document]
magopian: After discussing this with Pascal, we don't care about the other labels (variants) for now.

n1k0 (commented):
Great work, a few nits and remarks to be addressed

src/Data/Food/Ingredient.elm
magopian: I'll rewrite it as a single pipeline (removing the `let in`), and keep pipelines instead of chevrons ;)

src/Data/Food/BuilderQuery.elm
```diff
@@ -0,0 +1,147 @@
...
+
+
+deleteIngredient : IngredientQuery -> Query -> Query
+deleteIngredient ingredientQuery query =
+    { query
+        | ingredients =
+            query.ingredients
+                |> List.filter ((/=) ingredientQuery)
+    }
+        |> updateTransformMass
```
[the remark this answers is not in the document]
magopian: I like that the intent is clearly expressed the way it is:
- build the new query
- then update the transform mass
If you feel it's still better the way you put it, I won't fight you ;)

src/Request/Food/Db.elm
```diff
@@ -1,20 +1,35 @@
-module Request.Food.Db exposing (loadDb)
+module Request.Food.Db exposing (loadBuilderDb, loadExplorerDb)
```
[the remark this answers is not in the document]
magopian: This needed quite some work, but I believe it was a good thing to do 👍

src/Data/Food/Explorer/Recipe.elm
```diff
@@ -304,60 +304,51 @@ computeProcessImpacts item =
...
-
-encodeTransform : TransformQuery -> Encode.Value
-encodeTransform p =
-    Encode.object
-        [ ( "code", p.code |> Process.codeToString |> Encode.string )
-        , ( "mass", Encode.float (Mass.inKilograms p.mass) )
-        ]
+-- FIXME: remove this code once we're sure we don't want an API for the recipe explorer
+-- encodeQuery : Query -> Encode.Value
+-- encodeQuery q =
```
n1k0, line 309: I'd remove it right now. We can always crawl the git history if we ever want this code back.

src/Server/Query.elm
```diff
@@ -7,8 +7,10 @@ module Server.Query exposing
 
 import Data.Country as Country exposing (Country)
 import Data.Env as Env
-import Data.Food.Db as FoodDb
-import Data.Food.ExplorerRecipe as Recipe
+import Data.Food.Builder.Db as BuilderDb
+import Data.Food.Builder.Query as BuilderQuery
+import Data.Food.Ingredient as Ingredient
```
n1k0, line 12: You should `import Data.Food.Ingredient as Ingredient exposing (Ingredient)` to avoid repeating `Ingredient.Ingredient` in the code

src/Server/Query.elm
```diff
@@ -71,16 +72,34 @@ ingredientListParser key ingredients =
...
+                |> RE.andMap (Result.map .name ingredient)
+                |> RE.andMap (validateMass mass)
+                |> RE.andMap (Ok BuilderQuery.Default)
+
+        [ id, mass, variant ] ->
+            let
+                ingredient =
+                    id
+                        |> IngredientID.fromString
+                        |> Ingredient.findByID ingredients
```
n1k0, line 96: Small note as a nit: it's really idiomatic to have the data source at the last position in the args list, eg here I'd really really expect we have 

```elm
ingredient =
    ingredients
        |> Ingredient.findByID id
```

Maybe there's a good reason doing it the other way?

src/Static/Db.elm-template
```diff
@@ -1,11 +1,12 @@
...
 
-import Data.Food.Db as FoodDb
+import Data.Food.Explorer.Db as ExplorerDb
+import Data.Food.Builder.Db as BuilderDb
 import Data.Textile.Db as TextileDb
 
 
 type alias Db =
-    { foodDb : FoodDb.Db, textileDb : TextileDb.Db }
+    { builderDb : BuilderDb.Db, explorerDb : ExplorerDb.Db,  textileDb : TextileDb.Db }
```
n1k0, line 9: Hmm that means this file contains 3 times the impact definitions ; I wonder how we could have it parsed/loaded and defined only once

src/Static/Db.elm-template
```diff
@@ -30,17 +31,23 @@ foodIngredientsJson : String
 foodIngredientsJson =
     """%foodIngredientsJson%"""
 
+explorerDb : Result String ExplorerDb.Db
+explorerDb =
+    textileDb
+        |> Result.andThen
+            (\textileDbParsed ->
+            ExplorerDb.buildFromJson textileDbParsed.impacts foodProcessesJson foodProductsJson
```
n1k0, line 39: Nit: indentation here

tests/Server/RouteTest.elm
```diff
@@ -36,21 +36,22 @@ foodEndpoints db =
...
-          , "ingredients[]=568c715f977f32948813855d5efd95ba;149"
-          , "ingredients[]=65e2a1f81e8525d74bc3d4d5bd559114;100"
-          , "ingredients[]=a343353e431d7dddc7bb25cbc41e179a;168"
-          , "ingredients[]=3af9739fc89492167dd0d273daac957a;425"
-          , "transform=aded2490573207ec7ad5a3813978f6a4;1140"
+          -- Here goes our "famous" carrot cake…
+          , "ingredients[]=egg;120"
+          , "ingredients[]=wheat;140"
+          , "ingredients[]=milk;60"
+          , "ingredients[]=carrot;225"
```
n1k0, line 48: So much clearer, nicer DX 👍

n1k0 (commented):
Great work! A bit of fine tuning here and there and we're good to go 👍

src/Data/Food/Builder/Db.elm
```diff
@@ -0,0 +1,50 @@
...
+        |> Decode.decodeString (Process.decodeList impacts)
+        |> Result.andThen
+            (\processes ->
+                Decode.decodeString (Ingredient.decodeIngredients processes) ingredientsJson
+                    |> Result.map
+                        (\ingredients ->
+                            Db impacts processes ingredients
+                        )
+            )
+        |> Result.mapError Decode.errorToString
```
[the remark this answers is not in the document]
magopian: ah, yes, much better thanks 🙏

src/Static/Db.elm-template
magopian: We have the impacts in each "db" for two reason I believe: 
- in case the other databases aren't loaded (eg if at some point we'd rather not load the textile DB on the recipe builder page)
- to have access to it without passing around the local db and the "textile db" (or impacts db)

It would probably make sense to refactor that at a later point, indeed, but we might want to think a bit about how to architecture it, and i'd rather do it in a future PR.

n1k0 (approved):
Great work! Ship it after QA is happy 🚀

PR #192, opened 2022-12-20
Fix bvi normalization coefficient.

paulboosz:
The previous bvi normalization coef was the global normalization coefficient.
The right coefficient is the global coefficient divided by the world population, hence this fix


n1k0 (approved):
Much more significant indeed :) 👍

PR #196, opened 2023-01-03
Improve the builder forms

magopian:


styles.scss
```diff
@@ -606,6 +606,48 @@ blockquote {
...
+  grid-template-columns: minmax(75px, 116px) 1fr 160px auto 120px 31px;
+  grid-template-areas: "massInputWrapper ingredientSelector countrySelector bioCheckbox impactDisplay ingredientDelete";
+  padding: $list-group-item-padding-y $list-group-item-padding-x;
+  @media screen and (max-width: $viewport-md) {
+    grid-template-columns: minmax(75px, 116px) 1fr 89px 31px;
+    grid-template-areas:
+      "massInputWrapper ingredientSelector impactDisplay impactDisplay"
+      "countrySelector countrySelector bioCheckbox ingredientDelete";
+  }
+}
```
n1k0, line 645: I'm really not sure about all this... From my tests it gives weird results at smaller viewports, which would be the whole purpose of using such a complex grid layout. In the screenshot blow, we can't even read the ingredient names. I'm pretty sure all this would be better achieved using good old bootstrap responsive cols. Also, I'm not sure the CSS grid layout here would be much more maintainable and debuggable in the future.

image

src/Page/Food/Builder.elm
```diff
@@ -295,88 +302,83 @@ findExistingBookmarkName { builderDb, store } query =
...
+                excluded
+        , span [ class "text-end ImpactDisplay" ]
+            [ impact ]
+        , button
+            [ type_ "button"
+            , class "btn btn-sm btn-outline-primary IngredientDelete"
+            , title <| "Supprimer "
+            , onClick deleteEvent
+            ]
+            [ Icon.trash ]
```
n1k0, line 368: For some reason I couldn't identify browsing the patch, ingredient delete buttons don't work anymore

src/Data/Food/Builder/Recipe.elm
```diff
@@ -83,19 +83,30 @@ type alias Transform =
...
+        |> List.filter (\process -> not (List.member process.code usedProcesses))
+
+
+availableTransforms : Maybe Process.Code -> List Process -> List Process
+availableTransforms maybeCode processes =
+    case maybeCode of
+        Just code ->
+            processes
+                |> Process.listByCategory Process.Transform
+                |> List.filter (\process -> process.code /= code)
```
n1k0, line 104: Nit: could be

```elm
|> List.filter (.code >> (/=) code)
```

src/Page/Food/Builder.elm
```diff
@@ -152,9 +139,50 @@ update ({ queries } as session) msg model =
...
+                        |> Recipe.availableTransforms (Maybe.map .code query.transform)
+                        |> List.sortBy Process.getDisplayName
+                        |> List.head
+                        |> Maybe.map
+                            (\process ->
+                                Recipe.processQueryFromProcess process
+                                    |> (\processQuery ->
+                                            { processQuery | mass = defaultMass }
+                                       )
+                            )
```
n1k0, line 176: Nit: could be 

```elm
(Recipe.processQueryFromProcess
    >> (\processQuery -> { processQuery | mass = defaultMass })
)
```

src/Page/Food/Builder.elm
```diff
@@ -295,88 +302,83 @@ findExistingBookmarkName { builderDb, store } query =
...
-        , onSubmit
-            (case selectedProcess of
-                Just selected ->
-                    submit selected
-
-                Nothing ->
-                    noOp
-            )
+addProcessFormView { isDisabled, event, kind } =
+    li [ class "list-group-item" ]
```
n1k0, line 313: You need to apply margin and padding for visual consistency:

```elm
li [ class "list-group-item px-3 py-2" ]
```

src/Page/Food/Builder.elm
```diff
@@ -295,88 +302,83 @@ findExistingBookmarkName { builderDb, store } query =
...
+        [ button
+            [ class "btn btn-outline-primary"
+            , class "d-flex justify-content-center align-items-center"
+            , class " gap-1 w-100"
```
n1k0, line 317: Nit: leading space is unwanted

src/Page/Food/Builder.elm
```diff
@@ -390,22 +392,23 @@ updateIngredientFormView { excluded, db, ingredient } =
...
-                            NoOp
-            , disabled = False
-            }
-        )
-        (db.ingredients
+    li [ class "IngredientFormWrapper" ]
+        [ span [ class "MassInputWrapper" ]
+            [ MassInput.view
+                { mass =
+                    ingredient.mass
```
n1k0, line 399: Nit: this should be on the same line

n1k0 (commented):
The patch looks generally good, but atm ingredient deletion is broken. Also, I'm personnaly not a fan of the CSS grid based layout for the form, and would rather use the classic bootstrap responsive grid system myself instead. The thing being I don't feel good maintaining the CSS grid stuff on the long run... But I won't fight hard on this, if you feel it's the best way to go, let's go with it at least for a few weeks/mlonths and see how it goes.

src/Page/Food/Builder.elm
n1k0: Bah, it was a weird Parcel cache issue on my temporary linux machine, solved by itself after clearing it.

styles.scss
magopian: What screen width is that? I tested with "bootstrap sm", and thought it was the smallest we were interested in "supporting"?

n1k0: It was 375px wide (iPhone SE, a *very* common phone). RGAA [states](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#10.11) a min of 320px...

n1k0: But don't spend too much time on this, we can live with the grid stuff as it is in the patch just fine, we can always refactor/update this later

PR #199, opened 2023-01-10
Display the impact per kg of product

magopian:
cf [notion](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=13b41d53b0354d1ea5cab648107f7ad2&pm=s)

### DIFF
--- a/src/Page/Food/Builder.elm
+++ b/src/Page/Food/Builder.elm
@@ -783,15 +783,15 @@ sidebarView session db model recipe results =
                             |> String.fromFloat
                 in
                 [ div [ class "d-flex flex-column m-auto gap-1 px-2 text-center text-nowrap" ]
-                    [ h2 [ class "h5 m-0" ] [ text <| "Impact pour " ++ totalWeightStr ++ "kg de produit" ]
+                    [ h2 [ class "h5 m-0" ] [ text "Impact par kg de produit" ]
                     , div [ class "display-4 lh-1" ]
                         [ results.total
-                            |> Format.formatFoodSelectedImpact model.impact
+                            |> Format.formatFoodSelectedImpactPerKg model.impact totalWeight
                         ]
-                    , h3 [ class "h6 m-0 mt-2" ] [ text "Impact par kg de produit" ]
+                    , h3 [ class "h6 m-0 mt-2" ] [ text <| "Impact pour " ++ totalWeightStr ++ "kg de produit" ]
                     , div [ class "display-5 lh-1" ]
                         [ results.total
-                            |> Format.formatFoodSelectedImpactPerKg model.impact totalWeight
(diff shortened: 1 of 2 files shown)


src/Page/Food/Builder.elm
```diff
@@ -773,14 +773,29 @@ sidebarView session db model results =
...
+                    , h3 [ class "h6 m-0 mt-2" ] [ text <| "Impact pour " ++ totalWeightStr ++ "kg de produit" ]
+                    , div [ class "display-5 lh-1" ]
                         [ results.total
                             |> Format.formatFoodSelectedImpact model.impact
                         ]
-                    , small [ class "d-flex align-items-center gap-1" ]
+                    , small []
                         [ Icon.warning
-                        , text "Attention, ce résultat est partiel"
+                        , text " Attention, ce résultat est partiel"
```
n1k0, line 798: This should now read "Attention, ces résultats sont partiels"

src/Views/Format.elm
```diff
@@ -54,6 +55,14 @@ formatFoodSelectedImpact { trigram, unit, decimals } =
         >> formatRichFloat decimals unit
 
 
+formatFoodSelectedImpactPerKg : Impact.Definition -> Mass -> Impacts -> Html msg
+formatFoodSelectedImpactPerKg { trigram, unit, decimals } mass =
+    Impact.getImpact trigram
+        >> Unit.impactToFloat
+        >> (\impact -> impact / Mass.inKilograms mass)
+        >> formatRichFloat decimals unit
```
n1k0, line 63: So this is just and only an Html view thing? Don't we ever expect to have to expose this unitin API results? 

Shouldn't we rather have a dedicated function in the `Unit` module?

magopian: There's at the moment nothing in `Unit` related to `Impacts` or `Impact.Definition` or `Html`, so I'm not sure what you're suggesting? This helper is the pendant of `formatFoodSelectedImpact` defined just above.

n1k0: https://github.com/MTES-MCT/ecobalyse/blob/master/src/Data/Unit.elm#L393-L440 ?

magopian: This isn't the same `Impact.Definition` we're talking about, is it? I'm confused

n1k0: I think we should use `formatFoodSelectedImpact` and do the "per kg" conversion before using it. That with or without introducing some helper in `Unit`, like we did for `impactAggregateScore` for example.

n1k0: bah scratch it

n1k0 (approved):
LGTM ship it

PR #202, opened 2023-01-17
Display a score out of 100

magopian:
cf [notion](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=c0d9599591084dbeba62132aa8c49202&pm=s)

### DIFF
--- a/src/Data/Impact.elm
+++ b/src/Data/Impact.elm
@@ -13,6 +13,8 @@ module Data.Impact exposing
     , encodeImpacts
     , filterImpacts
     , getAggregatedScoreData
+    , getAggregatedScoreLetter
+    , getAggregatedScoreOutOf100
     , getDefinition
     , getImpact
     , grabImpactFloat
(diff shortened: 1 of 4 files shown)


src/Views/Format.elm
```diff
@@ -63,6 +64,48 @@ formatFoodSelectedImpactPerKg { trigram, unit, decimals } totalMass =
         >> formatRichFloat decimals (unit ++ "/kg")
 
 
+formatFoodSelectedImpactScore : Impact.Definition -> Mass -> Impacts -> Html msg
+formatFoodSelectedImpactScore { trigram } totalMass impacts =
```
n1k0, line 68: Let's suppose for a minute we want to expose this score to the API; would we be happy with an HTML only function? Even if that's not a requirement atm, wouldn't it be nice to have the computation extracted to a more agnostic function retyurning an `Int`, with maybe a few unit tests added?

src/Views/Format.elm
```diff
@@ -63,6 +64,48 @@ formatFoodSelectedImpactPerKg { trigram, unit, decimals } totalMass =
...
+                "B"
+
+            else if score >= 40 then
+                "C"
+
+            else if score >= 20 then
+                "D"
+
+            else
+                "E"
```
n1k0, line 99: I couldn't find a requirement for rendering a letter, is that an informal one?

n1k0 (commented):
Nice :+1: just a few comments

src/Views/Format.elm
magopian: yep, a discussion with Lisa about adding this as an experiment and base for future discussions

src/Views/Format.elm
```diff
@@ -63,6 +64,53 @@ formatFoodSelectedImpactPerKg { trigram, unit, decimals } totalMass =
...
+            else if score >= 60 then
+                "B"
+
+            else if score >= 40 then
+                "C"
+
+            else if score >= 20 then
+                "D"
+
+            else
```
n1k0: We don't need any of this now it's been moved to Impact.elm

src/Data/Impact.elm
```diff
@@ -403,6 +405,42 @@ getAggregatedScoreData defs getter =
...
+            logBase e
+    in
+    impacts
+        |> perKg totalMass
+        |> getImpact trigram
+        |> Unit.impactToFloat
+        |> (\value ->
+                -- See the documentation at https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire/impacts-consideres/score-100
+                (ln 2077 - ln value) / ln 2 * 20
+           )
```
n1k0, line 421: I'd really like to see at least a soingle unit test added, just for the sake of non-regression control over time

n1k0 (approved):
Perfect, ship it :rocket:

PR #203, opened 2023-01-17
Add and use a default origin for ingredients by default.

n1k0:
[User Story](https://www.notion.so/Introduire-un-pays-par-d-faut-pour-l-origine-des-ingr-dients-6ccba38b0be8440f9f08d16d10514b24)

### TODO

- [x] Ingredients default origin data (@paulboosz)
- [x] Implementation


n1k0:
While we still need the correct `default_origin` data for ingredients (for now I've set everything to `EuropeAndMaghreb`, cc @paulboosz), I think this patch can already be reviewed @magopian (@ccomb you're obviously welcome to give it a look)

public/data/food/ingredients.json
```diff
@@ -3,6 +3,7 @@
     "id": "egg",
     "name": "Oeuf",
     "default": "83752804d1bf53b08b817b1bf9f83d1b",
+    "default_origin": "EuropeAndMaghreb",
```
magopian, line 6: The proper way to edit this file is to edit [ingredients_base.json](https://github.com/MTES-MCT/ecobalyse-data/blob/master/src/ecobalyse_data/food/export_agb/ingredients_base.json) in ecobalyse-data, and have the script `export_builder.py` generate the `ingredients.json` file, and copy it over to ecobalyse.

magopian (commented):
We need to discuss that further i think (with @ccomb and @paulboosz ), at the moment the proper way to edit the following files should be to edit them in ecobalyse-data: 
- ingredients.json (actually edit ingredients_base.json and then generate the ingredients.json from it)
- impacts.json
- builder_processes.json (which becomes `food/processes.json` in ecobalyse)

So maybe before landing this PR have another one on ecobalyse-data with the changes in ingredients_base.json and ingredients.json? I can do it if you want, won't take long.

And great work for the code 👍

public/data/food/ingredients.json
n1k0: Fixed

PR #209, opened 2023-01-26
Exclude packaging mass from impacts/kg of product computations.

n1k0:
[User Story](https://www.notion.so/Ne-pas-prendre-en-compte-l-emballage-dans-la-masse-permettant-de-calculer-le-score-kg-4dad80072fa34aaf8285351033144ed5)


n1k0:
Pascal validated the numbers. As I need this patch for my next one, self-reviewing, approving and landing this now.

PR #210, opened 2023-01-26
Implement food category scales based scoring.

n1k0:
[User story](https://www.notion.so/RESULTATS-Introduire-des-chelles-intra-cat-gorie-pour-la-note-A-B-C-D-E-c3b7e176d44d4be985612ccf2dce3266)

### DIFF
--- a/src/Data/Impact.elm
+++ b/src/Data/Impact.elm
@@ -452,7 +452,7 @@ foodCategories =
         ]
 
 
-getAggregatedCategoryScoreOutOf100 : Definition -> String -> Impacts -> Int
+getAggregatedCategoryScoreOutOf100 : Definition -> String -> Impacts -> Result String Int
 getAggregatedCategoryScoreOutOf100 { trigram } foodCategory impactsPerKg =
     case Dict.get foodCategory foodCategories of
         Just { bounds } ->
@@ -472,9 +472,10 @@ getAggregatedCategoryScoreOutOf100 { trigram } foodCategory impactsPerKg =
                    )
                 |> floor
                 |> clamp 0 100
+                |> Ok
 
         Nothing ->
-            0
+            Err <| "Invalid: " ++ foodCategory
 
 
 getAggregatedScoreLetter : Int -> String
--- a/src/Data/Impact.elm
+++ b/src/Data/Impact.elm
@@ -12,6 +12,8 @@ module Data.Impact exposing
     , encodeAggregatedScoreChartEntry
     , encodeImpacts
     , filterImpacts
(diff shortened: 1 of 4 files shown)


src/Data/Impact.elm
```diff
@@ -418,10 +421,44 @@ getAggregatedScoreOutOf100 { trigram } impactsPerKg =
...
+getAggregatedCategoryScoreOutOf100 : Definition -> String -> Impacts -> Int
+getAggregatedCategoryScoreOutOf100 { trigram } foodCategory impactsPerKg =
+    case Dict.get foodCategory foodCategories of
+        Just { bounds } ->
+            let
+                ln =
+                    logBase e
+
+                ( min, max ) =
+                    bounds
```
magopian, line 446: Should we use the same names as in the docs, so `( impact100, impact0) = bounds` in this case? I had a hard time wrapping my head around the formula in the docs, and it didn't seem to fit with the formula I read below (took me some thinking to understand that `min` was in fact `impact100`)

src/Data/Impact.elm
```diff
@@ -418,10 +421,44 @@ getAggregatedScoreOutOf100 { trigram } impactsPerKg =
...
+                |> Unit.impactToFloat
+                |> (\value ->
+                        -- See docs at https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire/impacts-consideres/score-100#projet-declinaisons-du-score-100
+                        (ln (toFloat max) - ln value) / ln (toFloat max / toFloat min) * 20 * 5
+                   )
+                |> floor
+                |> clamp 0 100
+
+        Nothing ->
+            -99
```
magopian, line 459: This makes me kinda uncomfortable... I understand it's an "impossible value"... would we want to use `0` which is still in the range of "acceptable scores", or instead return a maybe or a result?

Not sure about this, just a feeling, a "smell"

magopian (approved):
Excellent work, thanks! A minor nit and an open question, you get to decide what to do with them ;)

src/Data/Impact.elm
n1k0: https://github.com/MTES-MCT/ecobalyse/pull/210/commits/b624b0c543a31e6bf8b9870387fb2b22d013ae90

src/Data/Impact.elm
n1k0: image

Fixed in https://github.com/MTES-MCT/ecobalyse/pull/210/commits/d975ddbab2add9dc2d117ac03c3ecccea0db959e

PR #212, opened 2023-01-27
Display ingredient detailed transports

magopian:
[user story: detailed transport per ingredient](https://www.notion.so/TRANSPORTS-Attacher-les-transports-chaque-ingr-dient-bb3413dd2db843cea1d938e3f0c307d0)
[user story: no more "recipe" step](https://www.notion.so/ETAPES-Suppression-de-Recette-ingr-dients-transformation-transports-78b992afaa3a4025a1d2be0d4528cc11)


styles.scss
```diff
@@ -549,7 +549,9 @@ blockquote {
   display: grid;
   grid-gap: 10px;
```
n1k0, line 550: In my tests in the devtools, `grid-gap: 7px 10px` gives slightly better results with the added transport info height.

n1k0 (approved):
Great job :+1:

PR #219, opened 2023-02-02
Add new BVI

paulboosz:


magopian (approved):
Looks great! I've updated the PR with the new export, and will now land it as I believe you're off today ;)

PR #225, opened 2023-02-17
Fix makingWaste openAPI maximum value

cerve:


paulboosz:
Merci pour cette contribution @cerve !

Est ce que tu aurais des sources/justifications sur cette augmentation de la valeur du taux de perte en confection maximal à 40% ?

cerve:
image

I think the documentation has an error in that at the top the maximum value that can be entered is 0.25 but below it is 0.4.
The API implements checking that the value is not greater than 0.4 so I thought the 0.25 in the documentation is wrong

paulboosz:
Oh indeed thanks for spotting this error !

n1k0:
Thanks for contributing!

PR #220, opened 2023-02-03
Enable or disable the plane transport

magopian:
[user story](https://www.notion.so/TRANSPORTS-Ajouter-du-transport-en-avion-par-exemple-en-ajoutant-la-mangue-5d57dee47db04f42b26776a1fbc77922)
[user story (bis)](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=2786364aac184ee88a14b480045ebe61&pm=s)

Following #218

TODO:
- [x] add an argument in the API to specify whether or not the transport is by plane
- [x] only allow plane selection for type4 ingredients (coming from out of Maghred and Europe by plane)
- [ ] modifications UI/UX avec Laurent
- [x] fix tests
- [x] fix failing (commented out) test in [server.spec.js](https://github.com/MTES-MCT/ecobalyse/pull/220/files#diff-1177cee604b4d39be39ced46f1b1e6e2c9af668d863d4072994ce6dc07080807R333-R340)

PR #227, opened 2023-02-17
Allow disabling the transport by plane even from default origin

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=2786364aac184ee88a14b480045ebe61&pm=s)

Follows on #220 by allowing the disabling of the transport by plane for category 4 ingredients (default origin by plane) even from their default origin.

### DIFF
--- a/src/Server/Query.elm
+++ b/src/Server/Query.elm
@@ -117,11 +117,7 @@ ingredientParser { countries, ingredients } string =
                 |> RE.andMap (Result.map .name ingredient)
                 |> RE.andMap (validateMass mass)
                 |> RE.andMap (variantParser variant)
-                |> RE.andMap
-                    (countries
-                        |> validateCountry countryCode Scope.Food
-                        |> Result.map Just
-                    )
+                |> RE.andMap (foodCountryParser countries countryCode)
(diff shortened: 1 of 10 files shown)


ccomb (approved):
Since a (Maybe Bool) is a set with 3 states, it might be made more explicit by defining a 3-state sum type for the byPlane with 3 constructors ?

magopian:
I've pushed a commit to use a union type, which made a significant change, would you mind reviewing again @ccomb and see if that's what you had in mind?

src/Server/Query.elm
```diff
@@ -230,17 +233,17 @@ validateBool str =
...
         "byPlane" ->
-            Ok (Just True)
+            Ok Ingredient.ByPlane
 
         "noPlane" ->
-            Ok (Just False)
+            Ok Ingredient.NoPlane
 
         _ ->
             Err "La valeur ne peut être que parmis les choix suivants: 'default', 'byPlane', 'noPlane'."
```
ccomb, line 249: parmis→parmi

ccomb (approved):
:+1: It's easier to understand with this explicit type !

PR #228, opened 2023-02-20
Add the transport to/from the logistics center for ingredients out of France

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=c5ea6ee3baa149198e25acbf6926c08d&pm=s)


src/Server/Query.elm
```diff
@@ -172,6 +164,17 @@ variantParser variant =
             Err <| "Format de variant invalide : " ++ variant
 
 
+foodCountryParser : List Country -> String -> Result String (Maybe Country.Code)
+foodCountryParser countries countryStr =
+    if countryStr == "default" then
```
ccomb, line 169: "default" and Nothing seem to play the same role. Isn't it possible to merge them ?

magopian: Yeah, `default` is a way to say that we don't want to define this value in the API, eg `milk;123;default;FR` to specify 123g of milk from France, with the `default` variant (which is no variant, so we take the non-organic process).

I'll land this PR as is, and in a future one, might get rid of all the `default` in favor of `;;`, eg `milk;123;;FR`.

PR #229, opened 2023-02-22
Remove `default` from api

magopian:
After some discussion on mattermost, we decided that it made more sense to just leave an value empty in the API instead of explicitly using `default`.


ccomb (approved):
:+1:

PR #233, opened 2023-03-08
Include raw to cooked ratio computations at the consumer step

n1k0:
Followup to #231


magopian (approved):
Nice! A couple nits in case you want to address them

PR #234, opened 2023-03-10
Add cold transport

magopian:
[user story](https://www.notion.so/TRANSPORTS-Diff-rencier-le-transport-depuis-pays-d-origine-frigorifique-non-9f310834a1a54cbfbfb8436b509e5877)

TODO:
- [x] proper icon (icomoon, check with @n1k0)
- [x] check results (with gsheet, check with @paulboosz)


data/food/export_agb/ingredients_base.json
```diff
@@ -6,6 +6,7 @@
     "default_origin": "EuropeAndMaghreb",
     "raw_to_cooked_ratio": 0.974,
     "density": 1,
+    "transport_cooling": "none",
```
n1k0, line 9: No cooling at all for eggs? We sure of that? I'm especially surprised as eggs seem to be the only food from our list which is never cooled when transported, which sounds counter-intuitive.

src/Data/Food/Builder/Recipe.elm
```diff
@@ -188,13 +188,22 @@ compute db =
                         Result.map (Retail.computeImpacts db volume distribution)
                             (Process.loadWellKnown db.processes)
 
+                    distributionTransportNeedsCooling =
+                        ingredients
+                            |> List.map (.ingredient >> .transportCooling)
+                            |> List.any ((/=) Ingredient.NoCooling)
```
n1k0, line 194: Nit: this could be written as 

```elm
ingredients
    |> List.any (.ingredient >> .transportCooling >> (/=) Ingredient.NoCooling)
```

src/Data/Transport.elm
```diff
@@ -33,7 +38,9 @@ type alias Distances =
 
 type alias Transport =
     { road : Length
+    , roadCooled : Length
     , sea : Length
+    , seaCooled : Length
     , air : Length
     , impacts : Impacts
     }
```
n1k0, line 46: Is it possible/wanted to have mixed cooled and non-cooled transports? If yes, scratch the rest of this comment, but if not, how about having a simple boolean doing this distinction? eg. 

```elm
type alias Transport =
    { cooled : Bool
    , road : Length
    , sea : Length
    , air : Length
    , impacts : Impacts
    }
```

n1k0 (approved):
Love the icons combo! A few nit and a question about the cooling transport distances data structure

data/food/export_agb/ingredients_base.json
magopian: Eggs don't need cooling when stored, and I guess this means they don't either need it when transported? See the last table in [the docs](https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire/transport#procedes-de-transport).

src/Data/Transport.elm
magopian: Yes it's indeed possible to mix cooled and non-cooled transports (eg: Mango is first transported with a non-cooled truck to the logistics center, and then cooled all the way to the transformation site in France)

PR #236, opened 2023-03-13
Replace picking input parameter with yarnSize.

n1k0:
[User Story](https://www.notion.so/Tissage-remplacer-Duitage-par-Titrage-ebaacda3fb8f4b29a6700b5440608622)


magopian (approved):
I played with https://ecobalyse-pr236.osc-fr1.scalingo.io/#/textile/simulator a bit, do we really need those sliders to be visible when they have no effect (when the textile is knitted), and not just when they are woven?

Other than that, as far as I can tell with my limited knowledge of the field, this looks good, with a couple of very minor nits.

n1k0:
> do we really need those sliders to be visible when they have no effect (when the textile is knitted), and not just when they are woven?

Went with disabling them, we may want to simply hide them in the future but we'll have to discuss this further when Alban is back. Landing this as is for now.

n1k0:
Inconsistencies have been reported, the feature has been rolled back on master. Investigation ongoing in #241.

PR #237, opened 2023-03-13
Manual migration agb3.1.1

paulboosz:
Script to manually update the impacts to agb3.1.1 with a manual export from simapro (agb311.csv)

Also remove the Apple and Durum Wheat ingredients ([user story](https://www.notion.so/Pomme-Correction-du-rendement-bio-ou-suppression-0837d3adca08498ea478c4ffe34699ac)).

Also change the ingredients `id`s to replace spaces with `-` in `ingredients_base.json` (simpler for the API).


magopian (approved):
Impressive work!

Just a few nits that we might not want/need to address if this is indeed a one-time-throw-off piece of code, you decide (I can also take over and do it, most of them are really easy and quick to address).

Great work, looking forward to use those split impacts in the export script!

Issue #226, opened 2023-02-17
[API] Error in the calculation of the sum of materials


cerve:
Due to a javascript issue, in some cases, the API materials sum check does not work.

Using 

`materials[]=coton;0.7&materials[]=acrylique-r;0.2&materials[]=jute;0.1` 

the API returns this error 

`{"errors":{"materials":"La somme des parts de matières doit être égale à 1 (ici : 0.9999999999999999)"}}`

In this case, we have 0.7 + 0.2 + 0.1 and it should not give error.

magopian:
Thanks for the feedback! We've fixed this in #242 ;)

Heads up: we're also looking to change all the "percentages" inputs in the API from float (between 0 and 1) to integer (between 0 and 100). There will be an entry in the changelog when that happens (soon hopefully!)

PR #242, opened 2023-03-16
Use Percentages

magopian:
This is a first draft of using "proper" percentages (stored internally as integers) to fix issues like #226.

What do you think @n1k0 , should I continue?


magopian:
Ok, I believe this is now up for review.

Please @n1k0 could you double check that I didn't mistook "ratios as percentages" (from 0 to 100) and "ratios as floats" (from 0 to 1) in the changes? I tried to be very cautious, and appart from a very minor rounding change in the tests, I didn't spot any deviations in the computation.

I don't think there are other ratios that could be changed to `Split`s, I've searched for `Unit.Ratio` on the codebase, and the only remaining uses are:
- `AggregatedScoreData.weighting`
- `Ingredient.rawToCookedRatio`
- `Preparation.elec`
- `Preparation.heat`

magopian:
@paulboosz if you have automated checks with the gsheet, and would love that you give this a second look too 🙏

src/Data/Food/Builder/Recipe.elm
```diff
@@ -357,13 +358,11 @@ computeIngredientTransport db { ingredient, country, mass, planeTransport } =
         planeRatio =
             -- Special case: if the default origin of an ingredient is "by plane"
             -- and we selected a transport by plane, then we take an air transport ratio of 1
-            Unit.Ratio
-                (if planeTransport == Ingredient.ByPlane then
-                    1
+            if planeTransport == Ingredient.ByPlane then
+                Split.full
```
n1k0, line 362: Same, `Spit 100` is more effective to me, matter of taste maybe.

src/Data/Food/Builder/Recipe.elm
```diff
@@ -357,13 +358,11 @@ computeIngredientTransport db { ingredient, country, mass, planeTransport } =
...
 
-                 else
-                    0
-                )
+            else
+                Split.zero
```
n1k0, line 365: Same

src/Data/Split.elm
```diff
@@ -0,0 +1,159 @@
...
+
+
+add : Split -> Split -> Result String Split
+add (Split a) (Split b) =
+    let
+        total =
+            a + b
+    in
+    if total < 0 || total > 100 then
+        Err "L'addition de plusieurs part doit être entre comprise entre 0 et 100%"
```
n1k0, line 122: s/L'addition/La somme
s/part/parts

src/Data/Split.elm
```diff
@@ -0,0 +1,159 @@
...
+
+
+toFloatString : Split -> String
+toFloatString =
+    asFloat >> String.fromFloat
+
+
+toPercentString : Split -> String
+toPercentString (Split int) =
+    String.fromInt int
```
n1k0, line 112: I'm surprise this doesn't contain the unit? (`%`)

src/Data/Textile/Inputs.elm
```diff
@@ -319,15 +320,15 @@ weavingOptionsToString _ _ =
...
     , airTransportRatio
         |> Maybe.andThen
             (\ratio ->
-                if Unit.ratioToFloat ratio == 0 then
+                if Split.asPercent ratio == 0 then
                     Nothing
 
                 else
-                    Just (Format.ratioToPercentString ratio ++ " de transport aérien")
+                    Just (Split.toPercentString ratio ++ " de transport aérien")
```
n1k0, line 331: Nit: maybe rename `ratio` to `split` in such circumstances to ease readability a bit

n1k0 (approved):
Looks good. A bunch of remarks you may want to address or not.

src/Data/Split.elm
```diff
@@ -0,0 +1,159 @@
...
+    if int < 0 || int > 100 then
+        Err ("Une part (en pourcentage) doit être comprise entre 0 et 100 inclus (ici: " ++ String.fromInt int ++ ")")
+
+    else
+        Ok (Split int)
+
+
+asFloat : Split -> Float
+asFloat (Split int) =
+    toFloat int / 100
```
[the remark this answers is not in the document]
magopian: yep, I pondered about that too, i'll do it

src/Data/Env.elm
```diff
@@ -37,14 +37,14 @@ githubUrl =
...
 minMakingWasteRatio =
-    Unit.ratio 0
+    Split.zero
 
 
-maxMakingWasteRatio : Unit.Ratio
+maxMakingWasteRatio : Split
 maxMakingWasteRatio =
-    Unit.ratio 0.4
+    Split.fourty
```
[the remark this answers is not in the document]
magopian: For now i'd like to keep the type as Opaque, so `Split.percent 40` would return a Result. All those helpers are for numbers we use in the code, and at least for now there aren't many, so I added them.

n1k0:
Forgot to answer your question; indeed `Preparation.elec` and `Preparation.heat` are percentages and should complete each others.

src/Data/Split.elm
```diff
@@ -0,0 +1,159 @@
...
+    asFloat >> String.fromFloat
+
+
+toPercentString : Split -> String
+toPercentString (Split int) =
+    String.fromInt int
+
+
+add : Split -> Split -> Result String Split
+add (Split a) (Split b) =
```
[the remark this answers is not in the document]
magopian: Good idea. When changing that, I realized it's not used anywhere... so actually i'll just drop this code entirely 😬

src/Data/Split.elm
magopian: The `%` sign is added in `Format.splitAsPercentage`

src/Data/Split.elm
```diff
@@ -0,0 +1,159 @@
...
+    in
+    if total < 0 || total > 100 then
+        Err "L'addition de plusieurs part doit être entre comprise entre 0 et 100%"
+
+    else
+        Ok (Split total)
+
+
+complement : Split -> Split
+complement (Split int) =
```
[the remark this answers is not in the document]
n1k0: I was thinking being able to do `complement [ Split.tenth, Split.twenty ] == Split 70` but now I'm not sure.

src/Data/Textile/Inputs.elm
magopian: Not sure about that: the record field name (that corresponds to the API, and the json data) is still `airTransportRatio`.

src/Data/Split.elm
magopian: Yep, it's sometimes useful to have a full representation with the % sign and as HTML, sometimes we just need the number "as an int between 0 and 100" without the sign (like in `ShareField` which uses this as the input value)

src/Data/Textile/Inputs.elm
n1k0: I won't die on this hill, please land this patch 😄

PR #244, opened 2023-03-17
update bvi norm

paulboosz:
[User Story](https://www.notion.so/MAJ-Facteur-de-normalisation-BVI-51c189433e9d44ce9880a04923cc9252?pvs=4)
Following a recalculation from P.Koch there is a new normalization factor for bvi


magopian (changes requested):
If you are happy with the changes in the numbers we can see in [the tests](https://github.com/MTES-MCT/ecobalyse/actions/runs/4448892158/jobs/7812318144?pr=244) then great! I can update the tests so it passes, and land if you want!

PR #248, opened 2023-03-21
Modify the comparator modal

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=90c1a9d44ff845c090b6f491065ba2ef&p=18b996c9758f45779b13777eb6d9ac38&pm=s)

- [x] reorder the impacts
- [x] change the colors
- [x] reaffect the impacts to the sub-scores so they're only affected once
- [x] add an option to view the sub-scores instead of the impacts (aggregated the impacts)

### DIFF
--- a/src/Views/Comparator.elm
+++ b/src/Views/Comparator.elm
@@ -229,7 +229,7 @@ foodComparatorView { session } { comparisonUnit, switchComparisonUnit, groupByPr
                     , onCheck updateGroupByProtectionAreas
                     ]
                     []
-                , text "Grouper les impacts par sous-scores"
+                , text "Grouper les impacts"
                 ]
     in
     div []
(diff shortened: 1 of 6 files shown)


src/Views/Comparator.elm
```diff
@@ -195,16 +247,49 @@ foodComparatorView { session } { comparisonUnit, switchComparisonUnit } =
...
+                                    (Tuple.mapSecond
+                                        (Impact.toProtectionAreas builderDb.impacts
+                                            >> (\protectionAreas ->
+                                                    [ { name = "Climat", color = "#7f7f7f", value = Unit.impactToFloat protectionAreas.climate }
+                                                    , { name = "Biodiversité", color = "#00b050", value = Unit.impactToFloat protectionAreas.biodiversity }
+                                                    , { name = "Santé environnementale", color = "#ffc000", value = Unit.impactToFloat protectionAreas.health }
+                                                    , { name = "Ressource", color = "#0070c0", value = Unit.impactToFloat protectionAreas.resources }
+                                                    ]
+                                                        |> List.reverse
+                                               )
```
n1k0, line 281: Niot: how about

```elm
    >> (\{ climate, biodiversity, health, resources } ->
            List.reverse
                [ { name = "Climat", color = "#7f7f7f", value = Unit.impactToFloat climate }
                , { name = "Biodiversité", color = "#00b050", value = Unit.impactToFloat biodiversity }
                , { name = "Santé environnementale", color = "#ffc000", value = Unit.impactToFloat health }
                , { name = "Ressource", color = "#0070c0", value = Unit.impactToFloat resources }
                ]
        )
```

n1k0: Also, why the need of reversing the list?

n1k0 (approved):
Looks good from a code standpoint, dunno from a methodology one.

src/Views/Comparator.elm
magopian: the list needs to be reverted because if i understand correctly the stacked data are presented with the first data point to the rightmost, and we'd like it to the leftmost. It's done the same for the "non-grouped" version

PR #250, opened 2023-03-23
Display bonuses impact in the comparator

magopian:


n1k0 (approved):
I just tested the feature and it's awesome. Couldn't spot any mistake in the diff, but tbh I'm exhausted so this is to be taken with a grain of salt. Maybe @ccomb wants to give the patch a look?

PR #257, opened 2023-03-29
Update the home page

magopian:
[user story](https://www.notion.so/Pr-senter-cobalyse-sur-le-site-f7bdbf887b7844619e0f8a3e21b6ae94)


public/img/Illustration_Score.png
n1k0, line 1: This image is 1.55MB. You can optimize png and jpeg files using programs like optipng or jpegoptim, both installable via homebrew on mac. We may want to explore the idea of leveraging Parcel to optimize these, though they have to be linked from html or css files, which isn't convenient here. In any case, we don't want to serve nor deal with 1.55MB png source files.

n1k0 (approved):
I'll be honnest, I don't like this new homepage, both in design and contents, especially the modal window to select either food or textile while we could simply have two buttons like… we have today. Anyway, I came to understand that UX/UI isn't part of my job anymore so scratch my feelings, I'll just validate the code, which is good and does the job. Just ensure optimizing image assets before landing.

n1k0:
A few things I've noticed since I first reviewed

This white stripe should be removed:
image

This app name isn't aligned with the logo anymore,n most probably since the font update:
image

Bottom card corners are mixing square and rounded corners, they should not:
image

Bottoms cards should be bottom-aligned for consistency — I know Laurent didn't do it that way but trust me on this one, that'd be much better
image

All these are nits though, up to you if you want to address them or not.

n1k0:
Not sur if P1/required but the iPad viewport size isn't rendering the three column layout nicely
image

PR #259, opened 2023-03-30
Disable animal welfare slider for ingredients not from animal origin.

n1k0:
[User Story](https://www.notion.so/Am-liorer-la-mod-lisation-Elm-des-des-bonuses-87adff9cf09e4b24bfa9ae98e46cb9da)

This also fixes `ecs` impacts where we didn't apply the default 10% animal-welfare bonus to ingredients from animal origin produced in a conventional fashion.


n1k0:
Self reviewed and landed.

PR #262, opened 2023-04-05
Fix mobile menu

magopian:
Some more DSFR fixes

### DIFF
--- a/src/Views/Page.elm
+++ b/src/Views/Page.elm
@@ -181,18 +181,18 @@ pageHeader config =
                 [ span [ class "fs-3" ] [ Icon.ham ] ]
             ]
         , Container.centered []
-            [ div [ class "pt-4 pb-2 ps-3" ]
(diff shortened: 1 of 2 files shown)


n1k0:
Isn't the fix I already pushed to master this morning isn't enough?

n1k0:
I remember we decided with @ccomb a few weeks ago that hotfixes could be pushed to master directly, but I understand watching the repo is painful. So we'll switch back to a "no direct push to master" policy, and I'll configure the repo to reject direct pushes to master as of now.

n1k0:
Hmm I don't think we should stick this much to the dsfr styles, as now the header takes a lot of vertical space, while the app already have to deal with a lot of information density, especially in the simulators… If you feel uncomfortable not strictly sticking to the dsfr guidelines, look at what [aides-territoires](https://aides-territoires.beta.gouv.fr/) does for instance: we can safely expect being "allowed" to do the same, in respect to end users comfort of use of the service.

image

magopian:
as you wish.

PR #266, opened 2023-04-11
Expose bonuses as API parameters.

n1k0:
[User Story](https://www.notion.so/Ajouter-les-param-tres-de-bonus-l-API-alimentaire-84fb88128c98449cbfdb97a3a77de569)

### DIFF
--- a/src/Data/Food/Builder/Recipe.elm
+++ b/src/Data/Food/Builder/Recipe.elm
@@ -11,7 +11,6 @@ module Data.Food.Builder.Recipe exposing
     , computeIngredientTransport
     , computeProcessImpacts
     , deletePackaging
-    , encode
     , encodeResults
     , fromQuery
     , getMassAtPackaging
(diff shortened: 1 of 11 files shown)


src/Page/Api.elm
```diff
@@ -48,11 +48,20 @@ getApiServerUrl { clientUrl } =
 
 changelog : List News
 changelog =
-    [ { date = "20 mars 2023"
+    [ { date = "12 avril 2023"
```
magopian, line 51: Why replace the previous changelog entry instead of adding a new one?

src/Data/Food/Builder/Recipe.elm
```diff
@@ -531,35 +530,6 @@ deletePackaging code query =
...
-encode q =
-    Encode.object
-        [ ( "ingredients", Encode.list encodeIngredient q.ingredients )
-        , ( "transform", q.transform |> Maybe.map encodeProcess |> Maybe.withDefault Encode.null )
-        , ( "packaging", Encode.list encodeProcess q.packaging )
-        , ( "preparation", Encode.list Preparation.encodeId q.preparation )
-        , ( "distribution", q.distribution |> Maybe.map Retail.encode |> Maybe.withDefault Encode.null )
-        ]
-
-
```
magopian, line 562: Removal of unused code? Nice one, I wonder if it would be possible to catch those early on? Or is it code that was used up til now?

src/Server/Query.elm
```diff
@@ -149,13 +149,79 @@ ingredientParser { countries, ingredients } string =
...
+                    (ingredient
+                        |> Result.andThen
+                            (\ingredientResult ->
+                                validateByPlaneValue byPlane ingredientResult
+                                    |> Result.andThen
+                                        (\maybeByPlane ->
+                                            Ingredient.byPlaneAllowed maybeByPlane ingredientResult
+                                        )
+                            )
+                    )
```
magopian, line 173: Maybe this could be extracted to a helper like `byPlaneParser` to be reused between the last two case branches?

tests/server.spec.js
```diff
@@ -347,6 +347,30 @@ describe("API", () => {
...
+      });
+
+      it("should validate an ingredient bonus applicability", async () => {
+        expectFieldErrorMessage(
+          await makeRequest("/api/food/recipe", ["ingredients[]=carrot;123;;BR;;100:100:100"]),
+          "ingredients",
+          /Carotte ne permet pas l'application d'un bonus sur les conditions d'élevage/,
+        );
+      });
+
```
magopian, line 373: It always feels as if the validation tests are duplicated between the RouteTest.elm and the server.spec.js, is that a good thing?

magopian (approved):
Excellent work, with a few nits 👍

src/Page/Api.elm
n1k0: Wow I was confused with the textile PR I guess. Alzheimer is at the corner…

PR #267, opened 2023-04-13
Optimize page weight and load performances.

n1k0:
Only load the rapi-doc script when users actually ask for the API page. That saves ~600KB on the all pages but the API one.

Before:

```
$ npm run build
(…)
dist/index.b4252d3a.js                 ⚠️  1.44 MB    3.80s
```

After:

```
$ npm run build
(…)
dist/index.05b5422a.js                 825.01 KB    2.44s
```

### DIFF
--- a/src/Page/Api.elm
+++ b/src/Page/Api.elm
@@ -33,7 +33,13 @@ type alias News =
 
 init : Session -> ( Model, Session, Cmd Msg )
 init session =
-    ( (), session, Ports.scrollTo { x = 0, y = 0 } )
+    ( ()
+    , session
+    , Cmd.batch
+        [ Ports.loadRapidoc "/vendor/rapidoc-9.3.4.min.js"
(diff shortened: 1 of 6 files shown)


src/Page/Api.elm
```diff
@@ -33,7 +33,7 @@ type alias News =
 
 init : Session -> ( Model, Session, Cmd Msg )
 init session =
-    ( (), session, Ports.scrollTo { x = 0, y = 0 } )
+    ( (), session, Ports.loadRapidoc "/js/rapidoc-9.3.4.min.js" )
```
magopian, line 36: We don't want/need the scrollTo anymore?

magopian (approved):
Looks good to me 👍

src/Page/Api.elm
n1k0: Damn, you're fast. Fixed and rebased.

PR #273, opened 2023-04-24
Use multiple knitting processes

magopian:
[user story](https://www.notion.so/Tricotage-mod-liser-les-diff-rents-proc-d-s-f9d5941b1f7e495a8d1b9bdbb66c9e47)

and 

[user story](https://www.notion.so/Confection-adaptation-selon-tricotage-seamless-et-fully-fashioned-f662b7f8b52d453fafc0bd16a122a010)


magopian:
@n1k0 can you have a look at this PR please?

I'm not sure it makes sense to have the knitting processes added to the list of `wellKnown`, it does simplify some things, and I don't think we'll expand on those to add new ones in the future (and even if we do, not sure it'll be a problem really)... what do you think?

src/Data/Textile/KnittingProcess.elm
```diff
@@ -0,0 +1,91 @@
+module Data.Textile.KnittingProcess exposing
```
n1k0, line 1: How about naming this module just `Knitting`?

src/Data/Textile/KnittingProcess.elm
```diff
@@ -0,0 +1,91 @@
...
+    , toLabel
+    , toString
+    )
+
+import Json.Decode as Decode exposing (Decoder)
+import Json.Decode.Extra as DE
+import Json.Encode as Encode
+
+
+type KnittingProcess
```
n1k0, line 15: This could be just `Knitting` or maybe `Kind` (but I like the main type being named the same as the module)

src/Data/Textile/KnittingProcess.elm
```diff
@@ -0,0 +1,91 @@
...
+            Ok FullyFashioned
+
+        "knitting-seamless" ->
+            Ok Seamless
+
+        "knitting-circular" ->
+            Ok Circular
+
+        "knitting-straight" ->
+            Ok Straight
```
n1k0, line 50: I'm not sure the `knitting-` prefix adds a lot of value; what would be the case where it's important or useful?

src/Data/Textile/Simulator.elm
```diff
@@ -424,43 +425,56 @@ computeSpinningImpacts db ({ inputs } as simulator) =
...
+        |> Process.loadWellKnown
+        |> Result.map
+            (\wellKnown ->
+                simulator
+                    |> updateLifeCycleStep Label.Fabric
+                        (\({ country, knittingProcess } as step) ->
+                            let
+                                productDefaultKnittingProcess =
+                                    knittingProcess
+                                        |> Maybe.withDefault KnittingProcess.Mix
```
n1k0, line 445: We could also define a default knitting process uuid for each product in products.json, and resolve them automatically. This would allow to specify different knitting techniques for different products, which seems very interesting and powerful to me, and very simple to implement. Thoughts?

n1k0 (commented):
A few questions, and I'd like to see the patch rebased on top of latest master with the new DSFR-like styles.

src/Data/Textile/Simulator.elm
magopian: at the moment there's already a default knitting process for each knitted product, in the `fabric` key. It's not really used here, and I was about to just get rid of the `fabric` process id and only keep `knitting` or `weaving`, as Alban confirmed that we won't need a "different default value for different products", and it's already always the same process for weaved products...

So I guess I went halfway: we don't need nor want different default values (at least for now?), but I didn't remove the `processUuid` from `fabric` either... so, what do you think, which way should I go?

magopian:
Ok I've address your review, I'll now start working towards the second user story regarding "making waste" 👍

magopian:
I believe this is now done and ready for a last review (maybe?) before landing

src/Page/Textile/Simulator.elm
```diff
@@ -313,7 +313,25 @@ update ({ db, queries, navKey } as session) msg model =
...
+                        | knittingProcess = Just knittingProcess
+                        , makingWaste =
+                            case knittingProcess of
+                                Knitting.FullyFashioned ->
+                                    Split.fromFloat 0.02
+                                        |> Result.toMaybe
+
+                                Knitting.Seamless ->
+                                    Split.fromFloat 0
+                                        |> Result.toMaybe
```
n1k0, line 327: It feels a lot this should be a function in the `Knitting` module

n1k0 (approved):
LGTM 👍

PR #275, opened 2023-04-26
Allow requesting the Food API using a JSON body.

n1k0:
image


openapi.yaml
```diff
@@ -294,6 +294,52 @@ paths:
...
+                value:
+                  ingredients:
+                  - id: egg
+                    mass: 0.12
+                  - id: wheat
+                    mass: 0.14
+                  - id: milk
+                    mass: 0.06
+                  - id: carrot
+                    mass: 0.225
```
magopian, line 319: Could we have some example ingredients with more info, like country of origin, by plane, variant, bonuses...?

magopian (approved):
Huge work, well done!

PR #281, opened 2023-05-12
Add a processes explorer

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04&p=5db16a15bec74db2b087b41232460b99&pm=s)


src/Data/Textile/Process.elm
```diff
@@ -35,6 +35,9 @@ type alias Process =
     { name : String
     , info : String
     , unit : String
+    , source : String
+    , correctif : String
+    , step_usage : String
```
n1k0, line 40: Nit: I know `elec_pppm` is in snake_case but it's currently one of the few exceptions to the general rule of using cameCase for record property names. FTR, I don't have a strong preference between these two styles but I'd really love consistency across the codebase.

n1k0 (approved):
Looks awesome, great work 👍

src/Page/Explore.elm
```diff
@@ -91,6 +119,26 @@ update session msg model =
...
+            , session
+            , (case model.dataset of
+                Dataset.Countries _ ->
+                    Route.Explore scope (Dataset.Countries Nothing)
+
+                _ ->
+                    Route.Explore scope (Dataset.Impacts Nothing)
+              )
+                |> Route.toString
+                |> Nav.pushUrl session.navKey
```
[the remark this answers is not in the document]
magopian: This test is the one that was [lower in the code](https://github.com/MTES-MCT/ecobalyse/pull/281/files#diff-bd0fa215f7e8cd01a3409dd77957a4e56349f4ca0a9aa64db42bc504231a9390L121) and was moved up here.

From what I understand, and that I added now as a comment:
```
                -- When changing scopes, if we were on a tab that is common between both scopes, don't "reset" the selected tab.
                -- Only the "impacts" and "countries" tabs are common at the moment, and the "impacts" tab is the one by default,
                -- so in effect this check makes sure that if we selected the "countries" tab and we change the scope, the
                -- selected tab isn't changed back automatically to the "impacts" tab.
```

src/Page/Explore/Table.elm
```diff
@@ -1,55 +1,79 @@
...
+    -> Html msg
 viewDetails scope createTable item =
     TableView.responsiveDefault [ class "view-details" ]
         [ createTable { detailed = True, scope = scope }
             |> List.map
                 (\{ label, toCell } ->
                     tr []
                         [ th [] [ text label ]
-                        , td [] [ toCell item ]
+                        , td [] [ item |> toCell ]
```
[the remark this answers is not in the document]
magopian: Because i changed the code a few times, and the final result doesn't need this anymore 😬 good catch!

src/Data/Textile/Process.elm
magopian: Sure, renaming now 👍

styles.scss
```diff
@@ -642,7 +642,6 @@ q {
           white-space: nowrap;
           overflow: hidden;
           text-overflow: ellipsis;
-          max-width: 320px;
```
n1k0, line 645: This was to prevent exaggeratedly wide columns like this:

image

magopian: yup I guess it was, but here's what [Laurent asked](https://mattermost.incubateur.net/fabnum-mte/pl/kszcyzrzu3nixj6h4kcu8zpjzr) : `Possible de ne pas tronquer la dernière colonne ?`

n1k0: Maybe we should show him this capture and ask him what to do?

n1k0: Maybe what we want is a parameter so we can specify the kind of column we want? narrow/wide, and maybe alignment (center, left, right) as I see we're not always consistent especially with numbers

image

My take would be to always right-align numbers.

n1k0: Sorry if this is outside of scope, we can probably file another card and I can work on this on my spare time

src/Page/Explore/Impacts.elm
```diff
@@ -14,94 +14,114 @@ import Views.Impact as ImpactView
...
+                        def.pefData
+                            |> Maybe.map (.normalization >> Unit.impactToFloat >> Format.formatRichFloat 2 def.unit)
+                            |> Maybe.withDefault (text "N/A")
+               }
+            :: { label = "Pondération (PEF)"
+               , toValue = .pefData >> Maybe.map (.weighting >> Unit.ratioToFloat >> Format.formatFloat 2) >> Maybe.withDefault "N/A"
+               , toCell = .pefData >> Maybe.map (.weighting >> Format.ratio) >> Maybe.withDefault (text "N/A")
+               }
+            :: (if scope == Scope.Food then
+                    -- No "scope d'impacts" for textile
```
n1k0, line 56: My previous comment disappeared, so here it is again: if the intent of this refactor is to handle conditional list entries, we could use `List.filterMap` here. Otherwise, I feel like I'm missing the use case of chaining many `::` operator calls instead of using a plain old regular list…

magopian: Yeah the `::` is the solution i went with as an elm-review rule wasn't happy about using `++` between lists which felt a bit nicer to me, but whatever. I could indeed use `Maybe`s  and `List.filterMap`, I guess i'll leave it as is for now.

PR #282, opened 2023-05-12
Use the selected knitting process' waste

magopian:
[user story](https://www.notion.so/Tricotage-adapter-taux-de-perte-8c6226030a734dfc90b3e17216b29b0b) : it was always the default product's knitting process' waste that was used, instead of the selected one's.


n1k0 (commented):
Looks good, though I'm surprised no API parameter and tests have been added?

n1k0:
Hmm sorry I was confused, it's not a new simulation parameter. Though I'd still have expected some tests to break and/or new ones to be added.

magopian:
I added tests, what do you think @n1k0 , is that enough?

n1k0:
Yeah sorry for the latency, added tests are good.

PR #286, opened 2023-05-22
Tooltips on comparative charts

magopian:
[user story](https://www.notion.so/Info-bulles-graphe-R-sultat-000634b5c7de4d3c902d1f1557b124b0)

### DIFF
--- a/src/Page/Food/Builder.elm
+++ b/src/Page/Food/Builder.elm
@@ -9,6 +9,7 @@ module Page.Food.Builder exposing
 
 import Browser.Events
 import Browser.Navigation as Navigation
+import Chart.Item as CI
 import Data.Bookmark as Bookmark exposing (Bookmark)
 import Data.Country as Country
 import Data.Dataset as Dataset
@@ -57,6 +58,7 @@ import Views.Impact as ImpactView
 import Views.Link as Link
 import Views.Modal as ModalView
 import Views.Spinner as Spinner
+import Views.Textile.ComparativeChart as ComparativeChart
 import Views.Transport as TransportView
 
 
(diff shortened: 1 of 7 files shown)


src/Page/Food/Builder.elm
```diff
@@ -91,6 +94,7 @@ type Msg
     | DeletePreparation Preparation.Id
     | LoadQuery Query
     | NoOp
+    | OnHover (List (CI.Many ComparativeChart.Entry CI.Any))
```
n1k0, line 97: Maybe name this msg slightly more explicitely eg. `OnChartHover` or `OnComparativeChartHover`?

src/Page/Food/Builder.elm
```diff
@@ -131,6 +135,7 @@ init ({ db, builderDb, queries } as session) trigram maybeQuery =
               , comparisonUnit = ComparatorView.PerKgOfProduct
               , displayChoice = ComparatorView.IndividualImpacts
               , modal = NoModal
+              , hovering = []
```
n1k0, line 138: Same, maybe `chartHovering` or `comparativeChartHovering`?

n1k0 (commented):
Looks nice, is it possible to decrease the font size and round the numbers though?

n1k0:
Also just wondering if there's any available API in elm-chart to have hovering applied on individual bar segments rather than on the whole segment itself?

src/Page/Food/Builder.elm
magopian: Renamed to `chartHovering`

n1k0 (approved):
The patch does the job, good work. We may want to investigate switching this chart to use Highcharts, but in another PR, later, if needed.

n1k0:
Hmm just tested the review app, we have this:
image

Is that expected?

magopian:
Yep, it's what we settled down to in the notion card ;)

PR #288, opened 2023-05-25
Reintroduce sub-scores, remove lettered scoring related code.

n1k0:


magopian:
oooooops my bad, I clicked on the button in the wrong window 😱

PR #290, opened 2023-05-30
add beef feedlot ingredient

paulboosz:


n1k0 (approved):
I'm so happy I heavily insisted so we add this ingredient variant without heavily refactoring the whole app :) Works like a charm.

n1k0 (approved):

PR #291, opened 2023-05-31
Spread food impact information over three tabs in the sidebar.

n1k0:
This is the first part of [this card](https://www.notion.so/Harmoniser-la-pr-sentation-du-d-tail-des-postes-entre-textile-et-alimentaire-c7530e3369e64dabaed7df8026bea733) (remaining work will be to add support for ecoscore to Textile — which should happen separately — and use the blocks provided by this patch to render impacts information tabs the same way we do for food here)


n1k0:
I need this patch to use the CardTabs stuff for a new feature, landing now.

PR #292, opened 2023-06-05
Allow exporting API requests from Web, expose Web url from API results.

n1k0:
[User Story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04)


magopian (approved):
That's awesome! I like the weburl which gives "two way" sharing between the UI and the API, well done 💯 

🚢

PR #293, opened 2023-06-06
Refactor aggregated scores

magopian:
The idea is to simplify the bonuses computation for food... and it's currently complicated because the aggregated scores must be computed before. And those are scattered around the food and textile codebases.

So isolate the `updateAggregatedScores` calls, and make sure there's only one (other than in tests) being called, right when we decode the impacts from the DB.

Then the rest of the computation for those aggregated scores happens just like the other impacts: they get multiplied by the quantity to get the total impact.

This has the benefit of
- not computing the scores so often for no real reason
- making sure we don't inadvertantly compute the sum of impacts, or bonuses, on an incomplete impact (if the aggregated score is not up to date)
- simplify the code

From what I could check, the only differences in the tests are rounding changes because the computation is done in a different order.

Previously:
- decode each **non aggregated** impact
- multiply each ** non aggregated** impact by the quantity chosen
- compute the aggregated impacts using the weighting and normalization of each of those computed impacts

Now:
- decode each **non aggregated** impact
- compute the aggregated impacts using the weighting and normalization of each of those impacts
- multiply each impact (included non aggregated) by the quantity chosen

Possible future work
- [ ] maybe limit the `Impact.updateAggregatedScores` calls so they are never called outside of `Impacts`... or replace all of them with `Impact.computeAggregatedScores` which is needed in some tests
- [ ] prevent the creation of `Impact.Impacts` from outside the module (make it an opaque type) so there's no way to inadvertantly create one with incorrect aggregated scores
    - [ ] this would need getting rid of `Impact.updateImpact` outside of the module (maybe have `insertImpact` that fails if the impact is already present, and that automatically computes the aggregatedScore)
    - [ ] and having proper constructors instead of both `noImpacts` and `impactsFromDefinitions` which seem somewhat redundant


src/Data/Food/Builder/Recipe.elm
```diff
@@ -165,7 +159,7 @@ compute db =
 
                     transformImpacts =
                         transform
-                            |> Maybe.map (computeProcessImpacts db.impacts >> List.singleton >> updateImpacts)
+                            |> Maybe.map (computeProcessImpacts >> List.singleton >> Impact.sumImpacts db.impacts)
```
n1k0, line 162: So we're creating a list with a single item which we sum. I'm pretty confident we could skip the singleton + sum stuff here.

src/Data/Food/Builder/Recipe.elm
```diff
@@ -203,39 +197,30 @@ compute db =
...
-                        updateImpacts
-                            [ ingredientsTotalImpacts
-                            , transformImpacts
-                            , ingredientsTransport.impacts
-                            ]
+                        [ ingredientsTotalImpacts
+                        , transformImpacts
+                        , ingredientsTransport.impacts
+                        ]
+                            |> Impact.sumImpacts db.impacts
```
n1k0, line 204: Nit: maybe keep the previous format?

```elm
Impact.sumImpacts db.impacts
    [ ingredientsTotalImpacts
    , transformImpacts
    , ingredientsTransport.impacts
    ]
```

n1k0 (approved):
Genius.

src/Data/Impact.elm
```diff
@@ -423,7 +436,7 @@ updateAggregatedScores : List Definition -> Impacts -> Impacts
 updateAggregatedScores definitions impacts =
     let
         aggregateScore getter trigram =
-            updateImpact trigram
+            AnyDict.insert trigram
                 (computeAggregatedScore getter definitions impacts)
```
n1k0, line 440: Nit: Maybe this form would be slightly clearer?

```elm
    aggregateScore getter trigram =
        impacts
            |> computeAggregatedScore getter definitions
            |> AnyDict.insert trigram
```

n1k0 (approved):

PR #296, opened 2023-06-08
More refactorings

magopian:
- a small rename to make sure we don't update an `Impact.Impacts` dict without later updating the aggregated scores
- load the well known processes at DB load (app load), instead of having to deal with results later on throughout the code

### DIFF
--- a/src/Data/Textile/Db.elm
+++ b/src/Data/Textile/Db.elm
@@ -19,6 +19,7 @@ type alias Db =
     , materials : List Material
     , products : List Product
     , transports : Distances
+    , wellKnown : Process.WellKnown
     }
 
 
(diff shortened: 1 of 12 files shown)


src/Data/Textile/Db.elm
```diff
@@ -53,5 +42,18 @@ decode =
...
+                                            result =
+                                                Process.loadWellKnown processes
+                                        in
+                                        case result of
+                                            Ok wellKnown ->
+                                                Decode.succeed (partiallyLoaded wellKnown)
+
+                                            Err err ->
+                                                Decode.fail err
+                                    )
```
n1k0, line 57: The same way you've used `RemoteData.fromResult`, you could use `Json.Decode.Extra.fromResult`:

```elm
import Json.Decode.Extra as DE

    |> Decode.andThen
        (\partiallyLoaded ->
            processes
                |> Process.loadWellKnown
                |> Result.map partiallyLoaded
                |> DE.fromResult
        )
```

n1k0 (approved):
Great job 👍

src/Data/Textile/Db.elm
magopian: ah yes true I wanted to do that, and forgot, thanks for the spot!

PR #297, opened 2023-06-08
Fixed WTU impacts and a few activities.

ccomb:
Fixed WTU impacts and a few activities


data/food/export_agb/export_builder.py
```diff
@@ -21,6 +20,7 @@
 BIOSPHERE = DBNAME + " biosphere"
 PROCESSES2EXPORT = "builder_processes_to_export.csv"
 INGREDIENTS_BASE = "ingredients_base.json"
+IMPACTS = "../../../public/data/impacts.json"  # TODO move the impact definition somewhere else and remove base impact
```
magopian, line 23: FYI in https://github.com/MTES-MCT/ecobalyse/pull/298 we're getting rid of this file altogether ;)

data/food/export_agb/export_builder.py
```diff
@@ -109,13 +128,22 @@ def compute_new_processes(activities, ingredients_base):
             for d in list(csv.DictReader(f, dialect="unix"))
         ]
 
-    # Parse the ingredients_base.json, which may contain complex ingredients to add/compute
+    # Parse the ingredients.json, which may contain complex ingredients to add/compute
```
magopian, line 131: Are we not loading `ingredients_base.json` anymore? If not, then maybe rename the global variable on the next line?

data/food/export_agb/export_builder.py
```diff
@@ -212,18 +244,32 @@ def compute_new_processes(activities, ingredients_base):
...
     with open(INGREDIENTS, "w") as outfile:
-        json.dump(ingredients_base, outfile, indent=2, ensure_ascii=False)
-        outfile.write("\n")  # Add a newline at the end of the file, as many editors do.
+        json.dump(ingredients, outfile, indent=2, ensure_ascii=False)
 
     # Add the new processes we computed for the complex ingredients
     export = [v["export_data"] for v in activities.values()] + new_processes
 
+    # compute the corrected impacts
+    print("Computing corrected impacts from weihgted subimpacts")
```
magopian, line 258: typo: `weighted`

data/food/export_agb/export_builder.py
```diff
@@ -212,18 +244,32 @@ def compute_new_processes(activities, ingredients_base):
...
+                if sub_impact_name in process["impacts"]:
+                    sub_impact = process["impacts"].get(sub_impact_name, 1)
+                    corrected_impact += sub_impact * correction_item["weighting"]
+                    del process["impacts"][sub_impact_name]
+            process["impacts"][impact_to_correct] = corrected_impact
+
     print(f"Export de {len(export)} procédés vers {BUILDER}")
     with open(BUILDER, "w") as outfile:
         json.dump(export, outfile, indent=2, ensure_ascii=False)
-        outfile.write("\n")  # Add a newline at the end of the file, as many editors do.
```
magopian, line 229: I believe @n1k0 asked me to do this, so when you open the exported file in an editor that adds a newline at the end of the file, it doesn't create a diff, so maybe it's better to keep it?

magopian (approved):
This looks good, impressive work!
A few questions regarding the loading of ingredients_base.json, other than that, looks good to me!

🚢

data/food/export_agb/export_builder.py
```diff
@@ -125,38 +153,42 @@ def compute_new_processes(activities, ingredients_base):
...
+        results = db.search(process["name"])
+        assert (
+            len(results) >= 1
+        ), f"In {PROCESSES2EXPORT}:{i}, searching this \"name\" field doesn't give a result: {process['name']}"
         activity = results[0]
-        activities[p["name"]] = {
+        activities[activity["Process identifier"]] = {
             "activity": activity,
-            "export_data": p,
+            "export_data": process,
```
n1k0, line 172: Now we're using "search strings" as process identifier proxies, it may be worth renaming that specific column and/or add a specific comment about this.

data/food/export_agb/export_builder.py
n1k0: Yes, please (thanks) 🙏

data/food/export_agb/ingredients_base.json
```diff
@@ -58,7 +58,7 @@
         "ratio": 0.0,
         "beyondLCA": {
           "agro-diversity": 50,
-          "agro-ecology": 0.5
+          "agro-ecology": 50
```
n1k0, line 61: I'm surprised with this change, are we sure this works with the way the Elm client currently decodes this data? 🤔

data/import_agribalyse.py
```diff
@@ -154,7 +154,13 @@ def import_agribalyse(
     agribalyse = bw2io.importers.simapro_csv.SimaProCSVImporter(
         datapath, dbname, normalize_biosphere=True
     )
-    agribalyse.strategies = [s for s in agribalyse.strategies if "water" not in repr(s)]
+    agribalyse.strategies = [
+        s
+        for s in agribalyse.strategies
+        if "water" not in repr(s)
+        and "normalize_simapro_biosphere_names" not in repr(s)
+        and "normalize_biosphere_names" not in repr(s)
```
n1k0, line 162: 💡 How about putting these in some `EXCLUDED_STRATEGIES` var…

data/import_method.py
```diff
@@ -30,14 +30,13 @@ def import_method(datapath=METHODPATH, project=PROJECT, biosphere=BIOSPHERE):
...
-    #        fix_localized_water_flows,
-    #    ]
-    #    + ef.strategies[8:]
-    # )
+    ef.strategies = [
+        s
+        for s in ef.strategies
+        if "water" not in repr(s)
+        and "normalize_simapro_biosphere_names" not in repr(s)
+        and "normalize_biosphere_names" not in repr(s)
```
n1k0, line 38: 💡… so we could reuse `EXCLUDED_STRATEGIES` here while maintaining it in a single place

n1k0 (commented):
Looks good! I'm only having questions about identifiers but other than that looks absolutely fantastic, great work 🎉

Ideally, we'd also want to introduce some activities impacts variation/regression test suite as mentioned this morning, though that would probably wants to be first landed on master, then this patch rebased over it, which may sounds like a little hassle. Up to you!

data/food/export_agb/export_builder.py
ccomb: ingredients_base.json is still parsed and the source of the ingredients.
This comment was changed during a too quick variable renaming...

data/food/export_agb/export_builder.py
ccomb: I suppose you mean the `name` column of the csv file. I renamed to `search`

data/food/export_agb/ingredients_base.json
ccomb: I was not sure but I blindly copied what I saw in the `plug-brightway-without-wtu` branch. 

I see that  in elm:
```
type alias Bonuses =
    { agroDiversity : Split  
    , agroEcology : Split
    , animalWelfare : Split    
    }
```
and 
```
type Split
    = Split Int
```
and `decodeBonuses` uses `decodePercent` which uses `Decode.int` and `fromPercent` and `fromPercent` expects 0 < n < 100

data/food/export_agb/ingredients_base.json
```diff
@@ -461,39 +460,18 @@
...
-        }
-      }
-    }
+    "visible": true
   },
   {
     "id": "eggplant",
     "name": "Aubergine",
-    "default": "Zucchini, conventional, national average, at farm gate {FR} U",
+    "default": "In-season eggplant consumption mix",
```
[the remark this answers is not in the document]
ccomb: Actually the reference information is the search string. Then during export, we retrive and expose the agribalyse identifiers so that people can rely on it to find the used processes.

Thinking about it, I'm not even sure we can define an invariable identifier, because even if the name/unit/location of a given process doesn't change between agribalyse and agribalyse N+1, the upstream processes can be different, which is enough to have a different identifier.  I guess we won't be able unique identifiers on a scope larger than a single database.

data/import_method.py
ccomb: done good idea

data/food/export_agb/ingredients_base.json
n1k0: Yes, I can see that, so the question is why was were these small floats before, and why was it working?

paulboosz (approved):

PR #299, opened 2023-06-16
Store the food builder db in the session as a WebData (no more Db.empty)

magopian:
More refactoring: 

- [x] save the food wellknown processes in the db as soon as it's loaded (same as what we did for the textile: #296)
- ~~[ ] split Textile.Db.(impacts|countries|transports) into Common.Db to be reused by Textile and Food databases~~ this is not possible because the countries are actually "Textile countries" with textile processes for electricity and stuff... so we would also need to load the textile processes, which makes less sense... or split the "textile part" of the countries data in another db... we'll see that in another PR ;)
- [x] load the food builder db in main so we don't have to deal with WebData in the food builder and explorer pages

### DIFF
--- a/src/Data/Food/Builder/Recipe.elm
+++ b/src/Data/Food/Builder/Recipe.elm
@@ -170,10 +170,9 @@ compute db =
                                         volume =
                                             getTransformedIngredientsVolume recipe
                                     in
-                                    Result.map (Retail.computeImpacts db volume distrib)
-                                        (Process.loadWellKnown db.processes)
+                                    Retail.computeImpacts db volume distrib db.wellKnown
                                 )
-                            |> Maybe.withDefault (Ok Impact.noImpacts)
+                            |> Maybe.withDefault Impact.noImpacts
 
                     distributionTransportNeedsCooling =
                         ingredients
@@ -192,9 +191,7 @@ compute db =
                                         )
                                     |> Maybe.withDefault (Transport.default Impact.noImpacts)
                         in
-                        db.processes
-                            |> Process.loadWellKnown
-                            |> Result.map (Transport.computeImpacts db.impacts mass transport)
+                        Transport.computeImpacts db.impacts mass transport db.wellKnown
 
                     recipeImpacts =
                         Impact.sumImpacts db.impacts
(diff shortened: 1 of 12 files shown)


src/Data/Food/Builder/Recipe.elm
```diff
@@ -192,9 +191,7 @@ compute db =
                                         )
                                     |> Maybe.withDefault (Transport.default Impact.noImpacts)
                         in
-                        db.processes
-                            |> Process.loadWellKnown
-                            |> Result.map (Transport.computeImpacts db.impacts mass transport)
+                        Transport.computeImpacts db.impacts mass transport db.wellKnown
```
n1k0, line 194: This function could just accept a `db` argument and extract `impacts` and `wellKnown` from it.

tests/Data/Food/PreparationTest.elm
```diff
@@ -23,8 +23,7 @@ suite =
                   , applyRawToCookedRatio = False
                   }
                     |> Preparation.apply builderDb (Mass.kilograms 1)
-                    |> Result.map (Impact.getImpact (Impact.trg "cch") >> Unit.impactToFloat)
-                    |> Result.withDefault 0
+                    |> (Impact.getImpact (Impact.trg "cch") >> Unit.impactToFloat)
                     |> Expect.within (Expect.Absolute 0.001) 0.08
```
n1k0, line 27: Nit: maybe just me but I'd find the traditional chaining more legible:

```elm
    |> Preparation.apply builderDb (Mass.kilograms 1)
    |> Impact.getImpact (Impact.trg "cch")
    |> Unit.impactToFloat
    |> Expect.within (Expect.Absolute 0.001) 0.08
```

n1k0 (approved):
Makes sense, great job 👍

src/Data/Food/Builder/Recipe.elm
magopian: Good point indeed, but for the signature i'll need to use extensible records notation `{ impacts : List Definition, wellKnown : Process.WellKnown }` as I can't import the Db in this file, or it'll create an import loop between `src/Data/Transport.elm` and `src/Data/Food/Db.elm`.

tests/Data/Food/PreparationTest.elm
magopian: Indeed, makes more sense, thanks for pointing it out!

PR #301, opened 2023-06-20
Fix hot module reloading by keeping the NavKey in the main model

magopian:
Parcel's hot module reloading needs the NavKey to always be in the same place. With the new way to load the session (sometimes a partial session, without the db, sometimes a full session with the db loaded), the nav key "changes" place.

The easy and naive fix is to duplicate the NavKey in the model.


n1k0 (approved):
Ship it 👍

PR #302, opened 2023-06-23
Substract inedible mass from ingredients', update impacts.

paulboosz:
[User Story](https://www.notion.so/ETAPES-Prendre-en-compte-la-part-non-comestible-des-ingr-dients-75a0d88be96241a6b70da7c64bdaf5b3)


src/Page/Food/Builder.elm
```diff
@@ -1046,28 +1044,38 @@ transportToTransformationView selectedImpact recipe results =
         ]
 
 
-transportToPackagingView : Recipe -> Html Msg
-transportToPackagingView recipe =
+transportToPackagingView : Recipe -> Recipe.Results -> Html Msg
+transportToPackagingView recipe results =
```
magopian, line 1048: I believe the `Recipe` is "kinda" also in the `Results`, right? Maybe we don't need to pass both the recipe and the results here? Not sure about the data architecture...

magopian (approved):
One very small nit, other than that, outstanding 💯

src/Page/Food/Builder.elm
n1k0: Hmm no, here with need `Recipe` to access the transform, `results.recipe` is just resulting impacts for just the recipe step, we don't know about the transform. So I'm gonna keep them both.

PR #305, opened 2023-06-28
Process impacts as a record

magopian:
This is an experiment, to decode process impacts as a record instead of an AnyDict

### DIFF
--- a/src/Data/Impact/Definition.elm
+++ b/src/Data/Impact/Definition.elm
@@ -1,8 +1,8 @@
 module Data.Impact.Definition exposing
     ( AggregatedScoreData
+    , Base
     , Definition
     , Definitions
-    , DefinitionsBase
     , Quality(..)
     , Source
     , Trigram(..)
@@ -93,7 +93,7 @@ type alias Definition =
     }
 
 
-type alias DefinitionsBase a =
+type alias Base a =
     {- We use a type variable here because this type is used for both
        * impact definitions (Definition.Definition)
        * processes impacts (Data.Impacts)
@@ -126,14 +126,14 @@ type alias DefinitionsBase a =
 
 
 type alias Definitions =
-    DefinitionsBase Definition
+    Base Definition
 
 
 
 ---- Helpers
 
 
-init : a -> DefinitionsBase a
+init : a -> Base a
 init a =
     { acd = a
     , bvi = a
@@ -162,7 +162,7 @@ init a =
     }
 
 
-update : Trigram -> (a -> a) -> DefinitionsBase a -> DefinitionsBase a
+update : Trigram -> (a -> a) -> Base a -> Base a
 update trigram updateFunc definitions =
     case trigram of
         Acd ->
@@ -262,7 +262,7 @@ trigrams =
     ]
 
 
-get : Trigram -> DefinitionsBase a -> a
+get : Trigram -> Base a -> a
 get trigram definitions =
     case trigram of
         Acd ->
@@ -333,7 +333,7 @@ get trigram definitions =
             definitions.pef
 
 
-map : (Trigram -> a -> b) -> DefinitionsBase a -> DefinitionsBase b
+map : (Trigram -> a -> b) -> Base a -> Base b
 map func definitions =
     { acd = func Acd definitions.acd
     , bvi = func Bvi definitions.bvi
@@ -362,7 +362,7 @@ map func definitions =
     }
 
 
-filter : (Trigram -> Bool) -> (a -> a) -> DefinitionsBase a -> DefinitionsBase a
+filter : (Trigram -> Bool) -> (a -> a) -> Base a -> Base a
 filter func zero base =
     -- Use the "zero-ing" function to "filter out" the fields that don't return True
     trigrams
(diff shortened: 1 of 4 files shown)


public/data/food/processes/builder.json
```diff
@@ -31,7 +31,9 @@
       "bvi": 0.2271,
       "etf-c": 1.29211471562,
       "htc-c": 1.577471183e-10,
-      "htn-c": 2.46769729382e-09
+      "htn-c": 2.46769729382e-09,
+      "pef": 0.0,
+      "ecs": 0.0
```
n1k0, line 36: Hmm why do we suddenly need these aggregated impacts as I think we decided they should always be computed in Elm. Have I missed something?

public/data/processes.json
```diff
@@ -21,7 +25,9 @@
       "pma": 3.32057e-07,
       "swe": 0.00235429,
       "tre": 0.0256924,
-      "wtu": 0.0
+      "wtu": 0.0,
+      "pef": 0.0,
+      "ecs": 0.0
```
n1k0, line 30: Same remark.

src/Data/Impact/Definition.elm
```diff
@@ -1,10 +1,11 @@
-module Data.Impact.Definition exposing (AggregatedScoreData, Definition, Definitions, Quality(..), Source, Trigram(..), decode, forScope, get, isAggregate, toString, toTrigram, trigrams)
+module Data.Impact.Definition exposing (AggregatedScoreData, Definition, Definitions, DefinitionsBase, Quality(..), Source, Trigram(..), decode, decodeBase, encodeBase, filter, foldl, forScope, get, init, isAggregate, map, toString, toTrigram, trigrams, update)
```
n1k0, line 1: Would you mind spreading these over multiple lines as we did for other modules? Thanks!

src/Data/Impact/Definition.elm
```diff
@@ -69,36 +70,137 @@ type alias Definition =
...
-    , ldu : Definition
-    , mru : Definition
-    , ozd : Definition
-    , pco : Definition
-    , pef : Definition
-    , pma : Definition
-    , swe : Definition
-    , tre : Definition
-    , wtu : Definition
+type alias DefinitionsBase a =
```
n1k0, line 73: It's probably worth adding a comment explaining that we need accepting either definitions or impacts.

n1k0 (approved):
Looks great 👍

src/Data/Impact/Definition.elm
```diff
@@ -126,7 +228,7 @@ trigrams =
     ]
 
 
-get : Definitions -> Trigram -> Definition
+get : DefinitionsBase a -> Trigram -> a
 get definitions trigram =
     case trigram of
         Acd ->
```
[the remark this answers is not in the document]
magopian: Good idea, thanks 🙏

public/data/food/processes/builder.json
magopian: Yes, so, about this: at the moment, the PEF scores are computed at the "exporting ciqual products" step in `export_ciqual.py` **for the food explorer**. But in any case, they are computed at the decoding step (so we actually don't use the decoded aggregated scores that are present for the food explorer).

I'm going to also modify the `export_ciqual.py` script to avoid compute the PEF score, if that's ok with @ccomb and @paulboosz ?

n1k0 (approved):
A few remarks ans questions. Fantastic work!

tests/TestUtils.elm
```diff
@@ -24,3 +28,32 @@ suiteWithDb name suite =
...
+        , Impact.getImpact Definition.Mru >> Unit.impactToFloat >> impacts.mru
+        , Impact.getImpact Definition.Ozd >> Unit.impactToFloat >> impacts.ozd
+        , Impact.getImpact Definition.Pco >> Unit.impactToFloat >> impacts.pco
+        , Impact.getImpact Definition.Pef >> Unit.impactToFloat >> impacts.pef
+        , Impact.getImpact Definition.Pma >> Unit.impactToFloat >> impacts.pma
+        , Impact.getImpact Definition.Swe >> Unit.impactToFloat >> impacts.swe
+        , Impact.getImpact Definition.Tre >> Unit.impactToFloat >> impacts.tre
+        , Impact.getImpact Definition.Wtu >> Unit.impactToFloat >> impacts.wtu
+        ]
+        subject
```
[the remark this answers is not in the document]
magopian: `Definition.map` will not return a list, so it's not totally useful here, but there's another way to write it, which is more concise, but less readable:

```
Definition.trigrams
        |> List.map
            (\trigram ->
                Impact.getImpact trigram >> Unit.impactToFloat >> Definition.get trigram impacts
            )
        |> (\expectations ->
                Expect.all expectations subject
           )
```

I can go with this if you think it's worth it 👍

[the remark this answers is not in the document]
n1k0: Yes I find it better!

src/Data/Impact/Definition.elm
```diff
@@ -69,42 +93,151 @@ type alias Definition =
...
-    , ldu : Definition
-    , mru : Definition
-    , ozd : Definition
-    , pco : Definition
-    , pef : Definition
-    , pma : Definition
-    , swe : Definition
-    , tre : Definition
-    , wtu : Definition
+type alias DefinitionsBase a =
```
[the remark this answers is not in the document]
magopian: Let's go with `Base`, you're right

paulboosz (approved):
`food/processes/explorer.json` and `food/processes.json` are generated from `export_ciqual.py` and I dont use these files with any other scripts/gsheets so there's no risk of breaking something on my part

PR #308, opened 2023-07-06
Create an ecoscore for textile

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04&p=80bbb462095d4a7b880072906cbbb4ec&pm=s)


src/Data/Bookmark.elm
```diff
@@ -122,24 +108,14 @@ findByTextileQuery textileQuery =
...
-
 sort : List Bookmark -> List Bookmark
 sort =
     List.sortBy (.created >> Time.posixToMillis) >> List.reverse
 
 
 toId : Bookmark -> String
 toId bookmark =
-    Scope.toString (scope bookmark) ++ ":" ++ bookmark.name
+    bookmark.name
```
n1k0, line 118: The idea here was to ensure always having unique ids by prefixing them with the scope, whatever the bookmark name. Imagine using `test` as a food bookmark then `test` as a textile one to understand the issue. I think we should keep addressing it, though I'm gonna leave that decision up to you.

src/Data/Food/Builder/Recipe.elm
```diff
@@ -472,11 +472,11 @@ deletePackaging code query =
...
 
 
-encodeResults : Definitions -> Results -> Encode.Value
-encodeResults definitions results =
+encodeResults : Results -> Encode.Value
+encodeResults results =
     let
         encodeImpacts =
-            Impact.encodeImpacts definitions Scope.Food
+            Impact.encodeImpacts
```
n1k0, line 479: We should now use `Impact.encodeImpacts` directly

src/Data/Impact.elm
```diff
@@ -231,11 +218,9 @@ encodeBonusesImpacts bonuses =
...
 
-encodeImpacts : Definitions -> Scope -> Impacts -> Encode.Value
-encodeImpacts definitions scope (Impacts impacts) =
+encodeImpacts : Impacts -> Encode.Value
+encodeImpacts (Impacts impacts) =
     Definition.encodeBase
-        definitions
-        scope
         Unit.encodeImpact
         impacts
```
n1k0, line 225: Do we really need this proxy, or should we now rather use `Definition.encodeBase` directly? 🤔

src/Data/Textile/LifeCycle.elm
```diff
@@ -153,6 +152,6 @@ updateSteps labels update_ lifeCycle =
     labels |> List.foldl (\label -> updateStep label update_) lifeCycle
 
 
-encode : Definitions -> LifeCycle -> Encode.Value
-encode definitions =
-    Encode.array (Step.encode definitions)
+encode : LifeCycle -> Encode.Value
+encode =
+    Encode.array Step.encode
```
n1k0, line 157: Do we really need this proxy, or should we now rather use `Encode.array Step.encode` directly? 🤔

n1k0 (approved):
Great one!

n1k0:
Testing the Web UI, the link to Textile in the main header menu is 404: https://ecobalyse-pr308.osc-fr1.scalingo.io/#/textile/simulator/ecs/per-item/simple

magopian:
Ah! That was a good catch about the 404s, I didn't think of testing the UI again after replacing all the "default trigrams" for textile and food, but hey, there was a [tricky bug](https://github.com/MTES-MCT/ecobalyse/pull/308/files#diff-11a6b4ecd50e091e5efa54b3ece6ae3afca372d5cfa57d25d69d6a34dfd9532cL178-R178)... So now I'm wondering if we should reintroduce some kind of `defaultTrigram` or `defaultImpact` that would be used in place of `Definition.Ecs` everywhere to prevent this kind of bug?
But then, the bug would only be prevented as long as we remember to use `defaultImpact` instead of `Definition.Ecs`... which would take us back to square one.

What do you think?

n1k0:
@magopian yeah I was hesitant about suggesting to reintroduce `defaultImpact`… I still like `defaultImpact` as being the one single place where the default impact is defined, so none are defined in a route. Maybe what we'd ideally like is a way to have a default route like `Route.TextileSimulatorHome` — and the same for the food one — so we don't have to mess around with default impact at all. Your call.

magopian:
Yeah, very good point for the `FoodBuilderHome` and `TextileSimulatorHome` routes, I've added them, and reintroduced an `Impact.default` that's used wherever `Definition.Ecs` was used as a default trigram. Hope I haven't missed any, would you care to have a last look before I merge (and go on holidays?).

n1k0 (approved):
Looks great! 🛳️

PR #316, opened 2023-07-13
Activities editor

ccomb:
- New activity/ingredient editor
- bleublanccoeur should not be the main category


n1k0 (commented):
I'll take the time to properly review this Monday, but meanwhile notice that FWIW this branch is conflicting with master on `activities.json`.

ccomb:
Vu, merci c'est le fameux feedlot bio qui est encore là :-)

data/notebooks/ingredients.py
```diff
@@ -50,22 +50,24 @@ def reverse(d):
...
-    "variants.organic.beyondLCA.animal-welfare": "Animal Welfare",
-    "variants.bleu_blanc_coeur": "Selected BleuBlancCoeur Process",
+    "bvi": "Bio-diversité",
+    "explain": "Détails",
+    "subingredient_default": "Sous-ingrédient conventionnel",
+    "subingredient_organic": "Sous-ingrédient bio",
+    "ratio": "Ratio Complexe/Simple",
+    "complements.agro-diversity": "Agro Diversity",
+    "complements.agro-ecology": "Agro Ecology",
+    "complements.animal-welfare": "Animal Welfare",
```
n1k0, line 70: I created a Banana ingredient, left untouched the animal-welfare slider, and got `0` in the resulting json:

image

Several observations:

- I'd like we either skip defining this key for non-animal (and non-dairy) products, or set it to `null`, so we can detect if a rangeslider is actionable by the user on the frontend. With a value of `0`, the frontend has no clue if the slider can be adjusted by end users or not. 
Note that I don't want to rely on some `if category /= AnimalProduct and category /= DairyProduct then allowSlider` check, because if we add another category impacting this behavior, the code will have to be updated in this specific test, which is easy to forget. Relying on the json data structure directly to know if the values should be updatable or not is the way to go.
- We have both a `category` and a `categories` keys. On the frontend, we only have `categories`. Why is that, and shouldn't we keep things coherent across the two environments? I understand that the  final `ingredients.json` file will be generated but I feel like the sooner we have proper data generated from the editor, the better, from a single point of maintenance perspective; thoughts?
Alternatively, maybe `category` wants to actually be `type`? Food for thoughts for sure.

n1k0 (approved):
Looks good with a few remarks, caveats and questions which I'll leave you to decide addressing them or not.

data/notebooks/ingredients.py
```diff
@@ -131,18 +135,34 @@ def read_ingredients():
...
+        ("misc"),
+        ("spice_condiment_additive"),
+        ("vegetable_fresh"),
+        ("vegetable_processed"),
+        ("organic"),
+        ("bleublanccoeur"),
+    ],
+    style=style,
+    allow_duplicates=False,
+)
```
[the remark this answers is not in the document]
ccomb: J'arrive pas à trouver comment élargir la cellule.

data/notebooks/ingredients.py
ccomb: Note this is the source file for ingredients, not the generated file. The `if category /= AnimalProduct ...` is actually present in the `export_builder.py` and the resulting builder.json file does not have the `animal-welfare` for animal_products.  However it looks like I forget to do the same for dairy_products. So I'm adding it.

ccomb: The category is the process category, while the categories are the ingredient categories. There are already two kinds oi categories in the frontend. 

One is in `Data/Food/Ingredient/Category.elm` and the other `Data/Food/Process.elm`

PR #319, opened 2023-07-18
Improve rendering of ingredient bonuses & transports.

n1k0:
[User Story](https://www.notion.so/Ajuster-la-pr-sentation-de-la-contribution-des-bonus-aux-impacts-ingr-dients-17cb8ae9487445c3b90ded98e6aac561)

image


n1k0:
I'm landing this because why not.

PR #320, opened 2023-07-18
Fix complex ingredients and beef-feedlot

paulboosz:
Issue : the ids for complex ingredients in `ingredients.json` were the ids of the "conventional" variant.

- added a `complex` tag to ingredients to explicit if it's complex or not. (Arguable)
- changed where the id is computed to associate the right id in `ingredients.json`
- added the missing elements to organic butter and feedlot-beef


data/food/export_agb/export_builder.py
```diff
@@ -53,6 +53,13 @@ def search(name):
     assert len(results) >= 1, f"'{name}' was not found in Brightway"
     return results[0]
 
+def find_id(activity):
+    # if this is a complex ingredient, the id is the one constructed by ecobalyse
+    if "ratio" in activity.keys():
```
n1k0, line 58: So if we have a ratio, it's the sign of a complex ingredient. So why do we also need a `complex` boolean? Isn't it redundant?

data/food/export_agb/export_builder.py
```diff
@@ -70,7 +77,8 @@ def search(name):
             "id": activity["id"],
             "name": activity["name"],
             "categories": [c for c in activity["categories"] if c != "ingredient"],
-            "default": search(activity["search"])["Process identifier"],
+            "complex": True if "ratio" in activity.keys() else False,
```
n1k0, line 80: Also if that's really what you want you can simply write `"complex": "ratio" in activity.keys()`. 

The thing I'm struggling understanding the most is where this JSON property is used? Hint?

n1k0 (approved):

n1k0 (changes requested):
@paulboosz any idea why we have diffing impacts in the failing tests? 

@ccomb could you please take a moment to carefully review the changes in `builder.json`? Impacts and bonus ratios have been updated for a few ingredients, I just want to ensure it's not a mistake.

data/food/export_agb/export_builder.py
n1k0: I've just read in the description

> added a complex tag to ingredients to explicit if it's complex or not. (Arguable)

So maybe what we want is a proper JSON data structure to declare the type of ingredient? If so — and that'd be a larger refactoring — we could go with something like this for simple ingredients:

```json
{
  "id": "onion",
  "name": "Oignon",
  …
  "construct": {
    "type": "simple"
  }
}
```

and for complex ones:

```json
{
  "id": "butter-organic",
  "name": "Beurre bio",
  …
  "construct": {
    "type": "complex",
    "default": "milk",
    "organic": "milk-organic",
    "ratio": 6.58
  }
}
```

Thoughts @cccomb @magopian?

ccomb: Another idea is to add an ingredient category called "complex ingredient".
Otherwise I've not yet understood why we need a "complex" attribute. It introduces the possibility to have an inconsistency by defining complex=True on an ingredient without ratio or the opposite

n1k0: Yeah the `complex` attribute should definitely go away.

As for the `ingredient-complex` new category, it smells a little complicated vs. a simple `construct` data structure explicitly expressing the different required configuration elements; decoding will be bullet-proof and easy to map to proper Elm types at least, if that's ever a concern.

paulboosz: agreed, I remove the `complex` attribute

paulboosz:
> @paulboosz any idea why we have diffing impacts in the failing tests?

Because I corrected the `animal-welfare` complements which were erroneously at 0 for milk and milk-organic. I put them to 10 so it impacts the carrot cake

PR #321, opened 2023-07-20
Introduce food impacts deviation checks in the test suite.

n1k0:


n1k0:
I need to pack up for holidays, so I'm gonna land this patch while nobody approved it. I'm a guy like that, yeah.

PR #333, opened 2023-08-16
Disallow setting air transport from France at the Making step.

n1k0:
[Bug report](https://www.notion.so/Distance-transport-1f160ec79d084fb98eaa92453411e51b)

This patch disables the air transport ratio rangeslider when the Making step takes place in France (and is then shipped to France by default). There's no real world situations where clothes are transported by plane from France to France.


magopian (approved):
Straight to the point, compact, perfect ;)

PR #336, opened 2023-08-21
Display the yarnSize field on all products, not only for weaved

magopian:
- [x] customize yarn size for knitted AND weaved clothes: [user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04&p=790775f0a49a4932b33ee28b42e42586&pm=s)
- [x] update the electricity impact computation for the spinning step: [user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04&p=78b584d1571f4b7694d7e0b03d4ae959&pm=s)
- [x] update the waste computation for the spinning step: [user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=37bf8aeb782747c3bf744ecaeafefd04&p=563c7616758c4f6e8ac181f1cbac4b16&pm=s)


magopian:
This is sadly a pretty big one, as the three user stories where kinda interlinked and depending on each other... sorry about that.

src/Data/Textile/Material.elm
```diff
@@ -2,22 +2,35 @@ module Data.Textile.Material exposing
...
+    , getAvailableSpinningProcesses
+    , getDefaultSpinning
     , getRecyclingData
+    , getSpinningElec
     , groupAll
     , idToString
+    , spinningFromString
+    , spinningToLabel
+    , spinningToString
+    , wasteForSpinning
```
n1k0, line 21: Nit, just an observation: from a distance, it almost looks like `Spinning` could be a module.

src/Data/Textile/Material.elm
```diff
@@ -40,6 +52,149 @@ type Id
...
+decodeSpinning : Decoder Spinning
+decodeSpinning =
+    Decode.string
+        |> Decode.andThen (spinningFromString >> DE.fromResult)
+
+
+encodeSpinning : Spinning -> Encode.Value
+encodeSpinning spinning =
+    spinningToString spinning
+        |> Encode.string
```
n1k0, line 120: Nit: how about just

```elm
encodeSpinning : Spinning -> Encode.Value
encodeSpinning =
    spinningToString >> Encode.string
```

src/Data/Textile/Simulator.elm
```diff
@@ -437,9 +446,7 @@ computeFabricImpacts db ({ inputs, lifeCycle } as simulator) =
                                     , yarnSize =
                                         inputs.yarnSize
                                             |> Maybe.withDefault
-                                                (inputs.product.yarnSize
-                                                    |> Maybe.withDefault Unit.minYarnSize
-                                                )
+                                                inputs.product.yarnSize
```
n1k0, line 449: Nit: for legibility, this can fit on its own line `|> Maybe.withDefault inputs.product.yarnSize`

src/Data/Textile/Simulator.elm
```diff
@@ -508,15 +515,28 @@ computeSpinningStepWaste ({ inputs, lifeCycle } as simulator) =
...
+                                        outputMaterialMass =
+                                            -- The output mass is the input mass of the next step
+                                            inputMass
+                                                |> Quantity.multiplyBy (Split.toFloat share)
+
+                                        inputMaterialMass =
+                                            -- Formula : inputMass - inputMass * waste = outputMass
+                                            -- => inputMass * (1 - waste) = outputMass
+                                            -- => inputMass = outputMass / (1 - waste)
+                                            Split.divideBy (Mass.inKilograms outputMaterialMass) (Split.complement processWaste)
```
n1k0, line 536: I know it's highly unlikely to ever happen but I'm slightly nervous with this giving `Infinity` when `Split.complement processWaste` is `0`. Well, that would output `Infinity` on the UI and we could live with it, and it will never happen anyway as no waste will ever be 100% so I understand we don't bother here.

src/Server/Query.elm
```diff
@@ -509,6 +531,35 @@ parseSplit string =
...
+                else
+                    Err <| "Un procédé de filature/filage doit être choisi parmi (" ++ (availableSpinningProcesses |> List.map Material.spinningToString |> String.join "|") ++ ") (ici: " ++ spinningString ++ ")"
+            )
+        |> Result.mapError
+            (always <|
+                "Un procédé de filature/filage doit être choisi parmi ("
+                    ++ (availableSpinningProcesses |> List.map Material.spinningToString |> String.join "|")
+                    ++ ") (ici: "
+                    ++ spinningString
+                    ++ ")"
```
n1k0, line 559: 🤩

src/Views/Textile/Step.elm
```diff
@@ -55,6 +56,7 @@ type alias Config msg =
     , updateAirTransportRatio : Maybe Split -> msg
     , updateDyeingMedium : DyeingMedium -> msg
     , updateEnnoblingHeatSource : Maybe HeatSource -> msg
+    , updateMaterialSpinning : Material -> Material.Spinning -> msg
```
n1k0, line 59: Nit: instinctively, I'd do the signature the other way around: `updateMaterialSpinning : Material.Spinning -> Material -> msg` so you can write stuff like `material |> updateMaterialSpinning spinning`

src/Views/Textile/Step.elm
```diff
@@ -185,6 +187,59 @@ dyeingMediumField { inputs, updateDyeingMedium } =
...
+                                    availableSpinningProcesses
+                                        |> List.map
+                                            (\spinningProcess ->
+                                                option
+                                                    [ value <| Material.spinningToString spinningProcess
+                                                    , selected <| Just spinningProcess == spinning
+                                                    ]
+                                                    [ text <| Material.spinningToLabel spinningProcess
+                                                    ]
+                                            )
```
n1k0, line 227: I'm surprised we don't use the extracted value here, which could (should?) be the available processes. I'm trying to wrap my head around why it's done that way, though if unavoidable a comment would be welcome.

src/Views/Textile/Step.elm
```diff
@@ -185,6 +187,59 @@ dyeingMediumField { inputs, updateDyeingMedium } =
...
+                                [ text material.shortName ]
+                            , case availableSpinningProcesses of
+                                [ spinningProcess ] ->
+                                    input
+                                        [ type_ "text"
+                                        , class "form-control w-75"
+                                        , disabled True
+                                        , value <| Material.spinningToLabel spinningProcess
+                                        ]
+                                        []
```
n1k0, line 215: This is a bit weird looking from a UX standpoint as one may think that's a text input while no input is allowed. Please use some greyed text instead (`text-muted` class will do the job I guess)

n1k0 (commented):
Very good job, congrats! A few comment and remarks to possibly address and we're good to go 👍

openapi.yaml
```diff
@@ -534,16 +534,22 @@ components:
...
         > Par exemple, les paramètres à passer pour un t-shirt 30% coton, 30% coton recyclé et 40% acrylique sont :
         >
         > ```js
         > materials[]=coton;0.3&materials[]=coton-rdp;0.3&materials[]=acrylique;0.4
         > ```
+
+        Le procédé de filature est à choisir entre :
+        - ConventionalSpinning : pour les matières naturelles ou artificielles
+        - UnconventionalSpinning : pour les matières naturelles ou artificielles
+        - SyntheticSpinning : pour les matières synthétiques
```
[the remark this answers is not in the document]
magopian: Ah, yes, thanks, I missed all that, sorry

src/Data/Textile/Material.elm
magopian: Yeah, I thought about that, and didn't... but I'm doing it right now, you're right, it makes more sense 👍

src/Data/Textile/Simulator.elm
magopian: Yeah, I didn't really think of an alternative to be honest... Should I instead first validate that the waste is bigger than 0, or else return a default value? Which one? Or display an error message of some sorts? But it would make it a pain to deal with at this point in the code...

src/Server/Query.elm
```diff
@@ -482,10 +482,32 @@ materialListParser key materials =
...
+                                    parseSpinning material spinningString
+                                        |> Result.map (\spinning -> { materialQuery | spinning = Just spinning })
+                                )
+                    )
+
         [ id, share ] ->
             Ok Inputs.MaterialQuery
                 |> RE.andMap (parseMaterialId_ materials id)
                 |> RE.andMap (parseSplit share)
+                |> Result.map (\constructor -> constructor Nothing)
```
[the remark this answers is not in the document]
magopian: Is that readable enough?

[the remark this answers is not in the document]
n1k0: Yeah, it's fine and I like `constructor` but I confess I slightly preffered `partiallyApplied` which is super super explicit.

src/Views/Textile/Step.elm
magopian: We could, but in this specific case, we're using it the other way round: we already have the material, and we're updating the **spinning** for this material, see line 235. What do you think?

src/Server/Query.elm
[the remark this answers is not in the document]
magopian: Excellent idea, renaming 👍

src/Views/Textile/Step.elm
n1k0: Maybe rename as `updateSpinningMaterial` then? I agree that's a funny name 😅

Your call 🤷‍♂️

magopian: We're updating the spinning of the material, so we're updating the material's spinning, so I guess the current name makes more sense.

src/Views/Textile/Step.elm
magopian: Yeah, I have the extracted value already in the `let..in` block above, but you're right I could move it in the `case..of` and then use the extract values, maybe that's more readable 👍

src/Views/Textile/Step.elm
magopian: Capture d’écran 2023-08-29 à 16 23 40
 What do you think?

n1k0 (approved):
LGTM with the remaining comments addressed 🛳️

magopian:
I believe i've now addressed all the comments, thanks!

PR #342, opened 2023-09-04
Rename some knitting processes

magopian:
[user story](https://www.notion.so/Tricotage-modif-nomenclature-00c4927b01584c03bfc770a4a493006f)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -687,9 +687,9 @@ components:
       description: |
         Taux de perte en confection (incluant la découpe), entre `0` et `0.4`.
 
-        **⚠️ Sur les produits tricotés en "fully fashioned", ce paramètre n'a aucun impact (valeur fixe: 0.02).**
+        **⚠️ Sur les produits tricotés en "fully fashioned / seamless", ce paramètre n'a aucun impact (valeur fixe: 0.02).**
 
-        **⚠️ Sur les produits tricotés en "seamless", ce paramètre n'a aucun impact (valeur fixe: 0).**
+        **⚠️ Sur les produits tricotés en "integral / whole garment", ce paramètre n'a aucun impact (valeur fixe: 0).**
       required: false
       style: form
       schema:
@@ -707,9 +707,9 @@ components:
           - `low` : faible
           - `very-low` : très faible
 
-        **⚠️ Sur les produits tricotés en "fully fashioned", ce paramètre n'a aucun impact (valeur fixe: very-low).**
+        **⚠️ Sur les produits tricotés en "fully fashioned / seamless", ce paramètre n'a aucun impact (valeur fixe: very-low).**
 
-        **⚠️ Sur les produits tricotés en "seamless", ce paramètre n'a aucun impact (valeur fixe: vide, non applicable).**
+        **⚠️ Sur les produits tricotés en "integral / whole garment", ce paramètre n'a aucun impact (valeur fixe: vide, non applicable).**
       required: false
       style: form
       schema:
@@ -737,8 +737,8 @@ components:
       description: |
         Le processus utilisé pour tricoter le tissu, parmi la liste de choix suivants :
           - mix : Tricotage moyen (mix de métiers circulaire & rectiligne)
-          - fully-fashioned : Tricotage fully-fashioned
-          - seamless : Tricotage seamless
+          - fully-fashioned : Tricotage fully-fashioned / seamless
+          - integral : Tricotage intégral / whole garment
           - circular : Tricotage circulaire, inventaire désagrégé
           - straight : Tricotage rectiligne, inventaire désagrégé
 
@@ -747,7 +747,7 @@ components:
       style: form
       schema:
         type: string
-        enum: ["mix", "fully-fashioned", "seamless", "circular", "straight"]
+        enum: ["mix", "fully-fashioned", "integral", "circular", "straight"]
     surfaceMassParam:
       name: surfaceMass
       in: query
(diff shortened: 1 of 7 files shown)


openapi.yaml
```diff
@@ -737,8 +737,8 @@ components:
       description: |
         Le processus utilisé pour tricoter le tissu, parmi la liste de choix suivants :
           - mix : Tricotage moyen (mix de métiers circulaire & rectiligne)
-          - fully-fashioned : Tricotage fully-fashioned
-          - seamless : Tricotage seamless
+          - fully-fashioned : Tricotage fully-fashioned / seamless
+          - integral : Tricotage intégral / whole garment
           - circular : Tricotage circulaire, inventaire désagrégé
           - straight : Tricotage rectiligne, inventaire désagrégé
```
n1k0, line 743: Could you add backticks around these identifiers so they're formatted as `<code>` tags?

openapi.yaml
```diff
@@ -1221,7 +1221,7 @@ components:
           enum:
           - mix
           - fully-fashioned
-          - seamless
+          - integral
           - circular
           - straight
```
n1k0, line 1226: Note that this enum is the very same as the one above, and may be factored and linked out, so we have a single place to update it.

src/Data/Textile/Knitting.elm
```diff
@@ -94,10 +94,10 @@ toLabel knittingProcess =
             "Tricotage moyen (par défaut)"
 
         FullyFashioned ->
-            "Fully fashioned"
+            "Fully fashioned / Seamless"
 
-        Seamless ->
-            "Seamless"
+        Integral ->
+            "Intégral / Whole garment"
```
n1k0, line 100: Do we keep the English terminology because *Whole garment* isn't directly translatable to eg. `vêtement complet`?

src/Page/Api.elm
```diff
@@ -54,7 +54,21 @@ getApiServerUrl { clientUrl } =
...
+      , level = "major"
+      , domains = [ "Textile" ]
+      , md =
+            """Le choix de valeurs pour le paramétrage `knittingProcess` a changé.
+            Il faut maintenant choisir entre
+            - mix : Tricotage moyen (mix de métiers circulaire & rectiligne)
+            - fully-fashioned : Tricotage fully-fashioned / seamless
+            - integral : Tricotage intégral / whole garment
+            - circular : Tricotage circulaire, inventaire désagrégé
+            - straight : Tricotage rectiligne, inventaire désagrégé
```
n1k0, line 67: Backticks around identifiers would be nice

n1k0 (approved):
LGTM 🛳️

src/Data/Textile/Knitting.elm
magopian: I believe it's the technical term used, the same way as `Seamless` (sans coutures) was ?

src/Page/Api.elm
```diff
@@ -54,7 +54,21 @@ getApiServerUrl { clientUrl } =
...
 changelog =
-    [ { date = "29 août 2023"
+    [ { date = "4 septembre 2023"
+      , level = "major"
+      , domains = [ "Textile" ]
+      , md =
+            """Le choix de valeurs pour le paramétrage `knittingProcess` a changé.
+            Il faut maintenant choisir entre
+            - mix : Tricotage moyen (mix de métiers circulaire & rectiligne)
+            - fully-fashioned : Tricotage fully-fashioned / seamless
```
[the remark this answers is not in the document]
magopian: Forwarding the question to Alban, and landing this in the meantime, thanks for the feedback 👍

openapi.yaml
```diff
@@ -737,8 +737,8 @@ components:
       description: |
         Le processus utilisé pour tricoter le tissu, parmi la liste de choix suivants :
           - mix : Tricotage moyen (mix de métiers circulaire & rectiligne)
-          - fully-fashioned : Tricotage fully-fashioned
-          - seamless : Tricotage seamless
+          - fully-fashioned : Tricotage fully-fashioned / seamless
+          - integral : Tricotage intégral / whole garment
```
n1k0, line 741: Actually it seems that the knitting process is `seamless` in English, I can't find any reference to `integral`. I think we should keep having identifiers iun English and descriptions in French, but we should definitevely NOT mix English and French in identifiers or descriptions.

openapi.yaml
```diff
@@ -736,18 +736,28 @@ components:
...
+        mix:
+          value: "mix"
+        fully-fashioned:
+          value: "fully-fashioned"
+        integral:
+          value: "integral"
+        circular:
+          value: "circular"
+        straight:
+          value: "straight"
```
n1k0, line 760: Ah sorry if it wasn't clear, I was actually suggesting to link to a definition, eg

```yaml
        enum:
            "$ref": "#/components/schemas/KnittingProcessEnum"
```

so the same definition can be used for both the GET param and thr JSON one for POST.

openapi.yaml
magopian: After discussion with Alban, we're leaving this as is for now, and we'll revisit the naming in the future if we get feedback from users.

openapi.yaml
magopian: I removed the `enum` there, and simply kept it in `knittingProcess`, which does exactly that, right?

n1k0: I'll have to checkout the branch and check, give me a minute

n1k0: So, nope, examples are not acceptable values 
image

PR #343, opened 2023-09-04
Allow comparing steps impacts for food products.

n1k0:
[user story](https://www.notion.so/Ajouter-un-onglet-tapes-dans-la-comparaison-alimentaire-f16987b8c6c247bfacfc1baa810506af)

### DIFF
--- a/review/src/ReviewConfig.elm
+++ b/review/src/ReviewConfig.elm
@@ -29,8 +29,9 @@ config =
         |> Rule.ignoreErrorsForFiles [ "src/Views/Debug.elm" ]
       -- Common
     , NoExposingEverything.rule
-        |> Rule.ignoreErrorsForDirectories [ "tests/" ]
+        |> Rule.ignoreErrorsForFiles [ "src/Data/Color.elm" ]
         |> Rule.ignoreErrorsForFiles [ "src/Views/Icon.elm" ]
+        |> Rule.ignoreErrorsForDirectories [ "tests/" ]
     , NoImportingEverything.rule
         [ "Html"
         , "Html.Attributes"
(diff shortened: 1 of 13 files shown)


review/src/ReviewConfig.elm
```diff
@@ -29,8 +29,9 @@ config =
         |> Rule.ignoreErrorsForFiles [ "src/Views/Debug.elm" ]
       -- Common
     , NoExposingEverything.rule
-        |> Rule.ignoreErrorsForDirectories [ "tests/" ]
+        |> Rule.ignoreErrorsForFiles [ "src/Data/Color.elm" ]
```
magopian, line 32: Why is this exception needed? Maybe a comment here?

src/Data/Food/Recipe.elm
```diff
@@ -668,6 +669,23 @@ resetDistribution query =
...
+            Impact.getImpact trigram
+                >> Just
+    in
+    { materials = getImpact results.recipe.ingredientsTotal
+    , transform = getImpact results.recipe.transform
+    , packaging = getImpact results.packaging
+    , transports = getImpact results.transports.impacts
+    , distribution = getImpact results.distribution.total
+    , usage = getImpact results.preparation
+    , endOfLife = Nothing
```
magopian, line 685: I believe you added the `endOfLife` here for future comparison with the textile?

magopian (approved):
Nice work!
I like that the `src/Data/Color.elm` are now reusable 👍 

🚢

review/src/ReviewConfig.elm
n1k0: For exactly the same reason as `Icon.elm` is excluded: it's a catalog of items we want to fully expose, there's nothing "private" and above all we want to keep the ability of having "unused" colors in the module, because it's a colorset and it doesn't make sense to have it partially implemented.

src/Data/Food/Recipe.elm
n1k0: Exactly, note that `StepsImpacts` is just a renaming of your previous `Steps` in the view, I've just made it a generic and transverse type of impacts.

PR #344, opened 2023-09-05
Harmonizing textile and food simulators

magopian:
[user story](https://www.notion.so/Brainstorming-vue-d-ensemble-volutions-front-textile-9c72a35830d74077bb950c6e5e49a75c)


src/Views/BaseComponent.elm
```diff
@@ -0,0 +1,144 @@
...
+import Html.Events exposing (..)
+import Views.Format as Format
+import Views.Icon as Icon
+
+
+type alias BaseComponent component quantity =
+    { component : component
+    , quantity : quantity
+    , country : Maybe Country
+    }
```
magopian, line 19: Hopefully it's clear enough that here the `BaseComponent` represents either
- a `MaterialInput`
- a `RecipeIngredient`

In both cases there is a component (`Material` or `Ingredient`), and a quantity (`Mass` or `Split`) and a country.

src/Views/Component/SplitInput.elm
```diff
@@ -0,0 +1,46 @@
+module Views.Component.SplitInput exposing (view)
```
magopian, line 1: This is a bit like `MassInput`, but for `MaterialInput.share`

magopian (commented):
This PR groups a few changes that are interconnected:
- no more tabs for textile: no more "detailed view" nor "dataviz" tabs
- use a generic "BaseComponent" to display and update base components, their quantity and their country
- use a modal to select base components

src/Views/BaseComponent.elm
```diff
@@ -0,0 +1,145 @@
...
+    , impact : Impacts
+    , selectedImpact : Definition
+    , update : BaseComponent component quantity -> BaseComponent component quantity -> msg
+    , delete : component -> msg
+    , selectComponent : component -> Autocomplete component -> msg
+    , quantityView : { disabled : Bool, quantity : quantity, onChange : Maybe quantity -> msg } -> Html msg
+    , toString : component -> String
+    , disableCountry : Bool
+    , disableQuantity : Bool
+    }
```
n1k0, line 43: How about ordering these alphabetically?

src/Views/BaseComponent.elm
```diff
@@ -0,0 +1,145 @@
...
+
+        deleteEvent =
+            delete baseComponent.component
+
+        autocompleteState =
+            AutocompleteSelector.init
+                toString
+                (db.components
+                    |> List.filter (\component -> not (List.member component excluded))
+                )
```
n1k0, line 60: Nit: how about:

```elm
        autocompleteState =
            db.components
                |> List.filter (\component -> not (List.member component excluded))
                |> AutocompleteSelector.init toString
```

src/Views/Textile/Step.elm
```diff
@@ -27,15 +30,18 @@ import Html exposing (..)
...
 import Views.Icon as Icon
 import Views.RangeSlider as RangeSlider
 import Views.Transport as TransportView
 
 
-type alias Config msg =
+type alias Config msg modal =
     { db : TextileDb.Db
     , inputs : Inputs
     , daysOfWear : Duration
```
n1k0, line 47: Nit: I'd really like these properties to be ordered alphabetically

src/Views/Textile/Step.elm
```diff
@@ -343,7 +353,7 @@ printingFields { inputs, updatePrinting } =
         ]
 
 
-fadingField : Config msg -> Html msg
+fadingField : Config msg odal -> Html msg
```
n1k0, line 356: I think you meant `modal`

src/Views/Textile/Step.elm
```diff
@@ -594,80 +612,232 @@ stepHeader { current, inputs, toggleStep } =
...
+
+         else
+            (inputs.materials
+                |> List.map
+                    (\materialInput ->
+                        let
+                            materialQuery : Inputs.MaterialQuery
+                            materialQuery =
+                                { id = materialInput.material.id
+                                , share = materialInput.share
```
n1k0, line 708: Nit: This `let…in` statement is huge, could we extract part of it into external functions, mostly for legibility purpose?

src/Views/Textile/Step.elm
```diff
@@ -594,80 +612,232 @@ stepHeader { current, inputs, toggleStep } =
...
+                                    \_ newComponent ->
+                                        updateMaterial
+                                            materialQuery
+                                            { materialQuery
+                                                | id = newComponent.component.id
+                                                , share = newComponent.quantity
+                                            }
+                                , delete = deleteMaterial
+                                , selectComponent = \_ autocompleteState -> setModal (addMaterialModal (Just materialInput) autocompleteState)
+                                , quantityView = \{ disabled, quantity, onChange } -> SplitInput.view { disabled = disabled, share = quantity, onChange = onChange }
```
n1k0, line 760: Nit: would gain in legibility spread over multiple lines

styles.scss
```diff
@@ -845,6 +833,32 @@ q {
     }
   }
 }
+.AddComponentFormWrapper {
+  border-bottom: $list-group-border-width solid $list-group-border-color;
+  display: grid;
+  grid-gap: 5px 10px;
+  grid-template-columns: 100px 1fr 1fr auto 120px 31px;
+  grid-template-areas:
+    "sharesTotal addComponentButton addComponentButton    addComponentButton    addComponentButton     addComponentButton";
```
n1k0, line 842: Nit: the multiple spaces separators were there to align the blocks, which isn't applicable anymore

n1k0 (changes requested):
Nice work! The patch is huge and I'm pretty sure I missed a few things in my review. 

My main concern to address before landing is that we lost a nice feature which allowed to navigate autocomplete results using the `Tab` and arrow keys, we should add it back. This may just be matter of ensuring the autocomplete search input gains focus when the modal opens and we start typing, which is not the case anymore with this patch.

Otherwide and again, great work 👍

index.js
```diff
@@ -91,12 +91,6 @@ app.ports.scrollIntoView.subscribe((id) => {
   node.scrollIntoView({ behavior: "smooth" });
 });
 
-app.ports.selectInputText.subscribe((id) => {
-  try {
-    document.getElementById(id).select();
-  } catch (_) {}
-});
-
```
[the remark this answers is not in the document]
n1k0: Now I think about it, it was for automatically selecting the url to share when clicking on the text input. But then we added the button for copying the link to the clipboard, so it's not useful anymore. Good thing you removed it!

src/Views/BaseComponent.elm
magopian: I quite like "Element", I'll give it a go and see how it unfolds, thanks for the suggestions.

src/Views/Textile/Step.elm
```diff
@@ -594,80 +610,232 @@ stepHeader { current, inputs, toggleStep } =
...
+                                (setModal
+                                    (addMaterialModal Nothing
+                                        (AutocompleteSelector.init .shortName availableMaterials)
+                                    )
+                                )
+                            , disabled <| length >= Env.maxMaterials
+                            ]
+                            [ Icon.plus
+                            , text "Ajouter une matière"
+                            , span [ class "text-muted" ] [ text "jusqu'à 5 maximum" ]
```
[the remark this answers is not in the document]
magopian: I changed that to either display "Ajouter une matière"

Capture d’écran 2023-10-02 à 10 52 58

or:

Capture d’écran 2023-10-02 à 10 53 40

src/Page/Textile/Simulator/ViewMode.elm
```diff
@@ -1,49 +1,17 @@
...
 
-import Url.Parser as Parser exposing (Parser)
-
 
 type ViewMode
-    = Dataviz
-    | DetailedAll
-    | DetailedStep Int
+    = DetailedStep Int
     | Simple
```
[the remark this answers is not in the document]
magopian: Good point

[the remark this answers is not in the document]
magopian: We might also get rid of the "detailed/simple" difference to harmonize with the food simulator, at some point... to be discussed ;)

src/Views/AutocompleteSelector.elm
```diff
@@ -0,0 +1,89 @@
...
+                        { onSelect = onAutocompleteSelect
+                        , mapHtml = onAutocomplete
+                        }
+
+                renderChoice : (Int -> List (Attribute msg)) -> Maybe Int -> Int -> a -> Html msg
+                renderChoice events selectedIndex_ index element =
+                    let
+                        selected =
+                            Autocomplete.isSelected selectedIndex_ index
+                    in
```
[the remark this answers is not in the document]
magopian: Good catch, thanks

src/Views/Textile/Step.elm
magopian: Extracted most of it in a `createConfig` helper 👍

magopian:
@n1k0 I believe i've addressed your comments, let me know what you think?

src/Page/Food.elm
```diff
@@ -510,8 +543,54 @@ type alias UpdateIngredientConfig =
     }
 
 
+createConfig : Query.IngredientQuery -> UpdateIngredientConfig -> BaseElement.Config Ingredient Mass Msg
+createConfig ingredientQuery { excluded, db, recipeIngredient, impact, selectedImpact } =
```
n1k0, line 547: Nit: I realize this now lives at the root of `Page/Food.elm`, so maybe we should name it more explicitely eg. `createElementSelectorConfig`

src/Page/Textile/Simulator.elm
```diff
@@ -416,6 +451,74 @@ update ({ textileDb, queries, navKey } as session) msg model =
...
+        |> Maybe.withDefault
+            -- Add a new Material
+            (model
+                |> update session (SetModal NoModal)
+                |> selectMaterial autocompleteState
+            )
+
+
+selectMaterial : Autocomplete Material -> ( Model, Session, Cmd Msg ) -> ( Model, Session, Cmd Msg )
+selectMaterial autocompleteState ( model, session, commands ) =
```
n1k0, line 498: Putting this here but it's probably not in this specific code block that it should be done: when testing the UI, I realize when selecting an Element using the keyboard (arrow keys then pressing Enter), the modal closes and the focus is then positioned on the "add bookmark button" in the sidebar, which makes using the app using the keyboard difficult (you have to switch to using the mouse).

What would be great is to set the focus on the quantity input after an element is selected and the autocomplete modal closes, so we can immediately define it's 50% or 200g, then tab again to gain focus on the "add new element" btn, rinse and repeat. Full keyboard navigation.

I have a feeling it's easy to implement in the update function when an element is selected but I may be wrong; if it's trivial, let's impolement it in this PR, if it's not, let's create a card because I think this would add significant UX value to the app.

src/Route.elm
```diff
@@ -65,13 +64,32 @@ parser =
...
+          -- for a view mode between `simple` and `detailed`, but now it's only `simple`,
+          -- and we used to have a functional unit parameter
+          (Parser.s "textile"
+            </> Parser.s "simulator"
+            </> Impact.parseTrigram
+            -- This is the unused "functional unit" parameter
+            </> Parser.string
+            -- This is the unused "viewmode" parameter
+            </> Parser.string
+            </> TextileQuery.parseBase64Query
```
n1k0, line 84: Nit: If we want to keep this BC bit, maybe extract this in some `deprecatedTextileRouteParser` fn to easily identify and remove it when the time comes?

I'd go with removing this BC layer entirely, but I'll leave the final decision up to you.

src/Views/BaseElement.elm
```diff
@@ -0,0 +1,143 @@
...
+    , span [ class "text-end ImpactDisplay fs-7" ]
+        [ impact
+            |> Format.formatImpact selectedImpact
+        ]
+    , deleteItemButton (List.length excluded == 1) deleteEvent
+    ]
+
+
+deleteItemButton : Bool -> msg -> Html msg
+deleteItemButton disable event =
```
n1k0, line 119: Nit: how about typing this `Bool` explicitly as `{ disabled : Bool }`, so it can be called as `deleteItemButton { disabled = List.length excluded == 1 } deleteEvent`?

n1k0 (approved):
Amazing work! A few bits you want to address or not, in this patch or in a next PR.

src/Route.elm
magopian: Good point on extracting it. I believe it doesn't cost much to keep it for now, this way we won't brake bookmarks. Maybe we'll revisit (and remove) in the future?

magopian:
Ok I've addressed your comments... but the "manage the focus". I'm working on a solution to that, but sadly it seems to be a bit complex. I'll have a solution to test shortly, see if it works well enough, and warrants the added code complexity.

n1k0:
@magopian yeah this kind of stuff can get tricky, don't bother if it's too complicated, we can address that later

magopian:
@n1k0 what do you think about the latest commit? Is it working for you?

src/Views/BaseElement.elm
```diff
@@ -0,0 +1,146 @@
...
+        , Attr.disabled disabled
+        ]
+        [ Icon.trash ]
+
+
+selectorView : element -> (element -> String) -> msg -> Html msg
+selectorView selectedElement toString selectElement =
+    button
+        [ class "form-select ElementSelector text-start"
+        , id <| "selector-" ++ toString selectedElement
```
n1k0, line 136: Maybe use the id, I'm slightly uncomfortable with any possible character being used as DOM ids

image

n1k0:
> @n1k0 what do you think about the latest commit? Is it working for you?

@magopian it now focuses on the selector itself rather than the quanbtity input but that works for me 👍

magopian:
Updated

n1k0:
@magopian ⛴️

PR #345, opened 2023-09-06
Allow comparing food and textile product impacts.

n1k0:
[user story](https://www.notion.so/Permettre-la-comparaison-textile-alimentaire-eb12435e4374432ba27c2c08b6b53515)


n1k0:
self-reviewed, landing before my weekend

PR #346, opened 2023-09-11
Simplify & harmonize food & textile impact score summaries.

n1k0:
[User Story](https://www.notion.so/Harmonisation-du-cartouche-bleu-alimentaire-textile-6bd67d08b76e4bd8978f428e682f062b)


n1k0:
I could go further but would need to get rid of the Examples page. I asked on MM, but I think we should land this patch first and iterate from here.

magopian (approved):
Very nice cleanup 💯 
The examples page doesn't make much sense anymore as you said, maybe get rid of it in a future PR? Seems to be the consensus on mattermost.

Good job!
🚢

PR #349, opened 2023-09-19
Compute the Out of Europe EoL textile complement.

n1k0:
[User Story](https://www.notion.so/Correctif-Fin-de-vie-hors-Europe-eddeb140a97f4955be410f7dc033082d)


src/Data/Textile/Simulator.elm
```diff
@@ -48,6 +49,8 @@ encode v =
         [ ( "inputs", Inputs.encode v.inputs )
         , ( "lifeCycle", LifeCycle.encode v.lifeCycle )
         , ( "impacts", Impact.encode v.impacts )
+
+        -- TODO: encode complementsImpacts?
```
magopian, line 53: Do we need it encoded? Where would it be used? In the API results? In a debug view like we have in the food simulator?

magopian (changes requested):
I believe there's a display issue: the end of life impact displayed on the left isn't the same as the end of life impact displayed in the summary on the right: it's missing the impact of the "out of Europe end of life", which probably needs to be added to the `Data.Textile.Steps.updateFromInputs` method?

Capture d’écran 2023-09-21 à 09 50 59

Other than that, I believe this looks good to me, nice!

src/Data/Textile/Simulator.elm
n1k0: Hence the question mark… It may be useful as an API detailed response indeed, so I'll implement it

n1k0:
@magopian could you please give the patch another look? thanks!

magopian (changes requested):
Looks good, with one final caveat:
The eol impact displayed on "simple view" is still the old impact without the "out of europe" complement taken into account (compared to the "detailed view" which displays the correct impact).

Capture d’écran 2023-09-21 à 17 39 53
Capture d’écran 2023-09-21 à 17 39 58

PR #351, opened 2023-10-02
Remove complex ingredients

paulboosz:
[User story](https://www.notion.so/Cr-er-les-ingredients-complexe-dans-brightway-fdf5665c29da432d8ee57c55ffd8bd4f?pvs=4)
Remove complex ingredients (eg organic flour, organic sunflower oil,...) precalculation
Instead add complex ingredients as activities in brightway when we're importing agribalyse into brightway
And then compute their impacts as any other activities


ccomb (approved):
Very nice!!

paulboosz:
After looking at diff in impacts with master : 
- textile : no diff in impacts
- food : 
      - difference on processes : chicken-breast-organic,huilecolza-organic,sunfloweroil-organic
      - reason : PR fix pre-existing errors in the model of complex ingredients

n1k0 (approved):
Looks good from a general data structure and implementation strategy perspectives. My personal Python coding style is way too foreign in the sense that as I usually avoid any kind of in-place & through-args mutations at all cost, but I'm not sure I want to have a debate about this. I'll let others comment on the implementation details if they want to.

ccomb:
Is the `activities_to_create` file supposed to be eventually contributed through the ingredient editor ?

paulboosz:
> Is the `activities_to_create` file supposed to be eventually contributed through the ingredient editor ?

I think it's a bit too technical so I'd say no. But I don't really know the need for contribution that much

PR #353, opened 2023-10-05
Fix spinning docs link

magopian:
[user story](https://www.notion.so/Lien-corriger-_-Filature-5e93958341e14f52a713b5d3dc3c5bfd)


n1k0 (approved):
It seems like the current situation with the documentation is far from ideal, so this will do for a little while until the broader issue is addressed.

PR #355, opened 2023-10-05
Empty distribution step textile

magopian:
[user story](https://www.notion.so/Ajout-d-une-tape-vide-distribution-pour-le-textile-bec3dff02dc5443c93127a3a49aad9ad)

PR #358, opened 2023-10-09
Color coded steps

magopian:
[user story](https://www.notion.so/Regroupement-des-tapes-transformation-pour-le-textile-b0d409f149b041bbaba94828621db792)

The idea is to use color codes to visualize the different steps accross food and textile.


magopian:
Caution: this PR is set on top of #355

@n1k0 not sure yet if that solution answers to the problem described in the user story, i'll wait for feedback from @PascalDagras .

Also, not sure it's worth it to create a new "component" for the border style, or if we should instead use css classes. The pro of using this solution is that we can reuse the color codes defined in `Data.Color`.

n1k0:
@magopian being color blind myself, I'm sure you'll understand why I'm slightly against this visual identification strategy :)

magopian:
> @magopian being color blind myself, I'm sure you'll understand why I'm slightly against this visual identification strategy :)

100% agree with that, we don't want the only information to be color! The information is already there (see the naming that's been changed in #355 to spell `transformation` in front of the relevant step names in the textile simulator).

This is reusing the color codes we already have in the comparator to add some coherence between the comparator, the steps tab in the right, and the steps headers on the left.
I'm also thinking about adding a "scroll to" when clicking on the right on the step name, to scroll the left part.

src/Data/Textile/Step/Label.elm
```diff
@@ -38,6 +41,63 @@ all =
...
+
+
+toId : Label -> String
+toId label =
+    case label of
+        Material ->
+            "materials-step"
+
+        Spinning ->
+            -- We only want a single "transform-step" id
```
n1k0, line 79: Nit: what would be useful is to explain *why* we only want a single one

n1k0 (approved):
👍

magopian:
@n1k0 : thanks for the feedback, I believe i've addressed it. I've also reverted the API "step labels" to return the previous strings (without the `transformation -` prefix).

src/Data/Impact.elm
```diff
@@ -153,7 +153,18 @@ noStepsImpacts =
...
-stepsColors : { materials : String, transform : String, packaging : String, transports : String, distribution : String, usage : String, endOfLife : String }
+type alias StepsColors =
+    { materials : String
+    , transform : String
+    , packaging : String
+    , transports : String
+    , distribution : String
+    , usage : String
+    , endOfLife : String
+    }
```
n1k0, line 164: Now if you look at the `StepImpacts`, it feels like we could abstract the type to accept a type variable.

n1k0: Something like this:

```diff
diff --git a/src/Data/Impact.elm b/src/Data/Impact.elm
index 5dd8d9fa..2a3b5742 100644
--- a/src/Data/Impact.elm
+++ b/src/Data/Impact.elm
@@ -130,17 +130,21 @@ totalComplementsImpactAsChartEntry complementsImpacts =
 -- Steps impacts
 
 
-type alias StepsImpacts =
-    { materials : Maybe Unit.Impact
-    , transform : Maybe Unit.Impact
-    , packaging : Maybe Unit.Impact
-    , transports : Maybe Unit.Impact
-    , distribution : Maybe Unit.Impact
-    , usage : Maybe Unit.Impact
-    , endOfLife : Maybe Unit.Impact
+type alias Steps a =
+    { materials : a
+    , transform : a
+    , packaging : a
+    , transports : a
+    , distribution : a
+    , usage : a
+    , endOfLife : a
     }
 
 
+type alias StepsImpacts =
+    Steps (Maybe Unit.Impact)
+
+
 noStepsImpacts : StepsImpacts
 noStepsImpacts =
     { materials = Nothing
@@ -154,14 +158,7 @@ noStepsImpacts =
 
 
 type alias StepsColors =
-    { materials : String
-    , transform : String
-    , packaging : String
-    , transports : String
-    , distribution : String
-    , usage : String
-    , endOfLife : String
-    }
+    Steps String
 
 
 stepsColors : StepsColors
```

n1k0 (commented):
I've commented on the commit, my bad ([link](https://github.com/MTES-MCT/ecobalyse/pull/358/commits/4851a3b990ac87e87db2dd54100c8e9359a15f7c#r1354320428))

src/Data/Impact.elm
magopian: Yeah, makes sense, adding this, then merging. Thanks!

n1k0 (approved):

PR #365, opened 2023-10-18
Prettier git hook

paulboosz:
[User story](https://www.notion.so/prettier-pre-commit-hook-6f0ebc388b1744f8a70f5fe67795f260?pvs=4)

PR #367, opened 2023-10-23
Json format

ccomb:
removed pformat
reformatted with prettier


n1k0 (commented):
as discussed this morning this is taken care of by @paulboosz in #365

PR #380, opened 2023-11-08
Fix bogus github changelog decoding with ingredient editor.

n1k0:
The changelog page is broken because some of the github API response author object fields are missing (ingredient editor commit data, cc @ccomb). Note that it would be nice if the ingredient editor could have an associated avatar image, but I'm not sure what would be required for that (we don't want to create a github account for that). Gravatar maybe? Anyway that's unrelated to this patch.

image

### DIFF
--- a/src/Data/Github.elm
+++ b/src/Data/Github.elm
@@ -2,6 +2,7 @@ module Data.Github exposing (Commit, decodeCommit)
 
 import Iso8601
 import Json.Decode as Decode exposing (Decoder)
+import Json.Decode.Pipeline as Pipe
 import Time exposing (Posix)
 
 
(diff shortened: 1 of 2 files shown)


magopian (approved):
LGTM 🚢

ccomb:
Mmmh not sure what to do with the avatar... since the ingredient editor push commits without a real github account (although it's authentified with a token of mine).

n1k0:
@ccomb that's why I mentioned [gravatar](https://gravatar.com). In my days, it allowed hosting and serving an image associated to an email address. If the service still exists (or if another/new one provides the same), maybe we could use it.

ccomb:
You mean we put  dummy email address as the commit author and register an icon for it on gravatar ?

n1k0:
Yeah that was the idea. It's no big deal really, we could also hardcode one as a fallback in the Elm decoder for that filed when the commit author name is "Ingredient Editor" and then serve an image we host ourselves, which would avoid relying on a third party service <-- ok this is what we should actually do

n1k0:
Fixed in 098de8b6 with a pizza 

image

PR #381, opened 2023-11-08
Harmonize textile step disabling

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=f872a7b99cbd4a19b9963b857846e188&pm=s)

### DIFF
--- a/src/Views/Textile/Step.elm
+++ b/src/Views/Textile/Step.elm
@@ -528,28 +528,13 @@ stepActions { current, detailedStep, index, toggleStepDetails } label =
 
 
 stepHeader : Config msg modal -> Html msg
-stepHeader { current, inputs, toggleStep } =
+stepHeader { current, inputs } =
     label
         [ class "d-flex align-items-center gap-2"
         , class "text-dark cursor-pointer"
         , classList [ ( "text-secondary", not current.enabled ) ]
-        , title
-            (if current.enabled then
-                "Étape activée, cliquez pour la désactiver"
-
-             else
-                "Étape desactivée, cliquez pour la réactiver"
-            )
         ]
-        [ input
-            [ type_ "checkbox"
-            , class "form-check-input mt-0 no-outline"
-            , attribute "role" "switch"
-            , checked current.enabled
-            , onCheck (always (toggleStep current.label))
-            ]
-            []
-        , h2 [ class "h5 mb-0" ]
+        [ h2 [ class "h5 mb-0" ]
             [ current.label
                 |> Step.displayLabel
                     { knitted = Product.isKnitted inputs.product
--- a/src/Views/Textile/Step.elm
+++ b/src/Views/Textile/Step.elm
@@ -692,128 +692,124 @@ viewStepImpacts selectedImpact { impacts, complementsImpacts } =
 viewMaterials : Config msg modal -> Html msg
 viewMaterials ({ addMaterialModal, db, inputs, selectedImpact, setModal } as config) =
     ul [ class "CardList list-group list-group-flush" ]
-        (if List.isEmpty inputs.materials then
-            [ li [ class "list-group-item" ] [ text "Aucune matière première" ] ]
+        ((inputs.materials
+            |> List.map
+                (\materialInput ->
+                    let
+                        nextCountry =
+                            config.next
+                                |> Maybe.withDefault config.current
+                                |> .country
 
-         else
-            (inputs.materials
-                |> List.map
-                    (\materialInput ->
-                        let
-                            nextCountry =
-                                config.next
-                                    |> Maybe.withDefault config.current
-                                    |> .country
-
-                            transport =
-                                materialInput
-                                    |> Step.computeMaterialTransportAndImpact db nextCountry config.current.outputMass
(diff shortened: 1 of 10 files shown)


src/Views/Textile/Step.elm
```diff
@@ -680,128 +681,124 @@ viewStepImpacts selectedImpact { impacts, complementsImpacts } =
...
+                        inputs.materials
+                            |> List.map .material
+
+                    availableMaterials =
+                        db.materials
+                            |> List.filter
+                                (\element ->
+                                    not
+                                        (List.member element excluded)
+                                )
```
n1k0, line 747: nit: not sure all this indentation helps with readability

src/Views/Textile/Step.elm
```diff
@@ -680,128 +681,124 @@ viewStepImpacts selectedImpact { impacts, complementsImpacts } =
...
+                    , button
+                        [ class "AddElementButton btn btn-outline-primary flex-fill"
+                        , class "d-flex justify-content-center align-items-center gap-1 no-outline"
+                        , id "add-new-element"
+                        , onClick
+                            (setModal
+                                (addMaterialModal Nothing
+                                    (AutocompleteSelector.init .shortName availableMaterials)
+                                )
+                            )
```
n1k0, line 790: nit: this can be written as

```elm
    , availableMaterials
        |> AutocompleteSelector.init .shortName
        |> addMaterialModal Nothing
        |> setModal
        |> onClick
```

I prefer this form but the one is the patch is fine too, so pick whatever you prefer

styles.scss
```diff
@@ -712,6 +712,10 @@ q {
     }
   }
 }
+.ElementDelete {
+  margin-left: auto;
+  width: 31px;
+}
```
n1k0, line 718: Any reason this one isn't wrapped within `ElementFormWrapper` like the other element styling rules?

tests/Server/RouteTest.elm
```diff
@@ -284,11 +279,6 @@ textileEndpoints db =
             |> Maybe.map .materials
             |> Expect.equal results
             |> asTest "should map the /textile/simulator endpoint with the list of materials"
-        , testEndpoint db "GET" Encode.null "/textile/simulator?"
-            |> Maybe.andThen extractTextileErrors
-            |> Maybe.andThen (Dict.get "materials")
-            |> Expect.equal (Just "La liste des matières est vide.")
-            |> asTest "should validate an empty materials list"
```
n1k0, line 291: I really liked this validation check 😭

n1k0 (approved):
I liked the checkbox option more, it feels weird to have no materials and no impacts for next transformation steps, except for some fixed values. The current checkbox option in prod allows to simply remove the materials step impacts, which was more useful in my humble opinion; that behavior isn't reproducible anymore with this patch.

If that's what we want, the code looks good to me.

styles.scss
magopian: The `ElementFormWrapper` is for the management of base elements (food ingredients and textile materials), but the `ElementDelete` for the delete button is used throughout the simulators, not only in the "base elements steps"

ccomb (approved):
Nice path towards factorization between textile and food!

I've not followed all the discussions but here are a few comments:

- this is not part of this PR, but regarding the sum of parts of textile materials: isn't there a risk that the sum as Float be différent from 1, such as 0.7+0.1+0.1+0.1 = 0.9999999999999999 (I could not reproduce in the ui though)
- I don't realy understand the rationale behind disabling the materials step. I must have missed something
- regarding the allowEmptyList:  there is a (unprobable but) possible formal state with allowEmptyList=False and the list of ingredients=[] on the food side. But I don't see how to prevent it. It would require different types (List for food and NonEmptyList for textile). So anyway, forget it.

magopian:
>     * this is not part of this PR, but regarding the sum of parts of textile materials: isn't there a risk that the sum as Float be différent from 1, such as 0.7+0.1+0.1+0.1 = 0.9999999999999999 (I could not reproduce in the ui though)

This is dealt with using `round` as you can see [here](https://github.com/MTES-MCT/ecobalyse/blob/harmonize-textile-step-disabling/src/Views/Textile/Step.elm#L772-L773)

>     * I don't realy understand the rationale behind disabling the materials step. I must have missed something

The idea is to be able to disable each and every step to allow for better customization. However, the materials step is a bit different: we can't just allow deleting all the materials, as we need at least one to compute the waste at the spinning step, and some complements like the "end of life probability to be outside of europe". So the materials steps still needs a "disable step checkbox" instead of simply adding a way to remove everything from the step itself like we did for the others.

>     * regarding the allowEmptyList:  there is a (unprobable but) possible formal state with allowEmptyList=False and the list of ingredients=[] on the food side. But I don't see how to prevent it. It would require different types (List for food and NonEmptyList for textile). So anyway, forget it.

That may be a possibility, sure, but as you said, we might not need to deal with this now. The worse that would happen is that somehow all the materials would be removed (because the boolean enabled the last material's delete button), and the impact calculation would be off.

PR #386, opened 2023-11-15
Removed subingredients as we don't have complex ingredients now

ccomb:
Removed subingredients since we don't have complex ingredients

Complex ingredients created during food export have been replaced by constructed processes created during AGB import


magopian (approved):
LGTM 🚢

PR #389, opened 2023-11-16
Add two additional substances

ccomb:
These are provided by ecoinvent and seem to be present in ei 3.9.2.
they are used in EF3.1 adapted


paulboosz:
Nice ! Did you compute the impacts and it didn't change anything ?

ccomb:
I've done the exports, it looks like it didn't change the impacts.

PR #391, opened 2023-11-16
Rename "out of Europe" end of life complement

magopian:
[user story](https://www.notion.so/Renommer-Fin-de-vie-hors-Europe-en-Export-hors-Europe-c8e513a505c84276a7d57aef3f926212)


ccomb (approved):
Found that in Views/Textile/Step.elm: Probablilité → Probabilité

PR #394, opened 2023-11-20
Add new ennobling processes

magopian:
[user story](https://www.notion.so/Int-grer-inventaires-enrichis-v1-5e1d1aae5195434bb1471277afb0a9a7) and [user story](https://www.notion.so/Nomenclature-Explorteur-Textile-d32f272ab93b400690173c8856b06475)


ccomb (approved):
I was wondering: is there a real benefit of using a Wellknown database instead of just extracting the wanted process from the process database?

magopian:
> I was wondering: is there a real benefit of using a Wellknown database instead of just extracting the wanted process from the process database?

At this point in the code, we have processes that have been "validated", and are no longer mere results, so it's easier to just address them instead of doing the whole "is this a result or an error" dance.

PR #403, opened 2023-11-27
Harmonize simulator reset

magopian:
[user story](https://www.notion.so/Harmonisation-r-initialisation-de-la-simulation-9f64a9ade51c4690a4d0ed26e52ad5ba)

### DIFF
--- a/src/Page/Food.elm
+++ b/src/Page/Food.elm
@@ -65,6 +65,7 @@ import Views.Transport as TransportView
 type alias Model =
     { db : FoodDb.Db
     , impact : Definition
+    , initialQuery : Query
     , bookmarkName : String
     , bookmarkTab : BookmarkView.ActiveTab
     , comparisonType : ComparatorView.ComparisonType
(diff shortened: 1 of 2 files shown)


ccomb (commented):
For food, I feel like there are now two buttons that do the same thing, with different names : one in the top (Create a new Recipe) and one in the bottom (reset simulator).
For textile there is only one in the bottom.

IMO there should be only one button in the top, similar to the current "New recipe" button, and beside the template product selector.

magopian:
> For food, I feel like there are now two buttons that do the same thing, with different names : one in the top (Create a new Recipe) and one in the bottom (reset simulator). For textile there is only one in the bottom.
> 
> IMO there should be only one button in the top, similar to the current "New recipe" button, and beside the template product selector.

We've discussed this with Pascal, it's the way to go, for now at least. The next steps are to harmonize the two, and remove the buttons in the food simulator, and replace it with a selector of different "recipe examples".

I'll land this now, and build on it in another PR.

PR #408, opened 2023-11-30
Food examples autocomplete

magopian:
[user story](https://www.notion.so/Int-gration-d-exemples-de-produits-alimentaires-e2a5df9775874d10a1dbee8a8c5f9471)

### DIFF
--- a/src/Data/Food/Query.elm
+++ b/src/Data/Food/Query.elm
@@ -371,6 +371,7 @@ recipesAndNames =
     , ( "Carrot cake", carrotCake )
     , ( "Ratatouille en conserve", ratatouille )
     , ( "Raviolis en conserve", raviolis )
+    , ( "Épinards congelés", frozenSpinnach )
     ]
 
 
@@ -558,6 +559,33 @@ raviolis =
           }
         ]
     , distribution = Just Retail.ambient
+    , preparation =
+        [ Preparation.Id "pan-warming"
+        ]
+    }
+
+
+frozenSpinnach : Query
+frozenSpinnach =
+    { ingredients =
+        [ { id = Process.codeFromString "spinach"
+          , mass = Mass.grams 840
+          , complements = Nothing
+          , country = Nothing
+          , planeTransport = Ingredient.PlaneNotApplicable
+          }
+        ]
+    , transform =
+        Just
+            { code = Process.codeFromString "AGRIBALU000000003102449"
+            , mass = Mass.grams 840
+            }
+    , packaging =
+        [ { code = Process.codeFromString "AGRIBALU000000003114927"
+          , mass = Mass.grams 100
+          }
+        ]
+    , distribution = Just Retail.frozen
     , preparation =
         [ Preparation.Id "pan-warming"
         ]
--- a/src/Data/Food/Query.elm
+++ b/src/Data/Food/Query.elm
@@ -14,6 +14,7 @@ module Data.Food.Query exposing
     , emptyQuery
     , encode
     , parseBase64Query
+    , ratatouille
     , serialize
     , setDistribution
     , setTransform
@@ -27,7 +28,7 @@ module Data.Food.Query exposing
 
 import Base64
 import Data.Country as Country
-import Data.Food.Ingredient as Ingredient
+import Data.Food.Ingredient as Ingredient exposing (PlaneTransport(..))
 import Data.Food.Preparation as Preparation
 import Data.Food.Process as Process
 import Data.Food.Retail as Retail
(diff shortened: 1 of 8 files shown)


src/Data/Food/Query.elm
```diff
@@ -412,3 +360,381 @@ parseBase64Query =
...
+recipesAndNames : List Product
+recipesAndNames =
+    [ { name = "Produit vide", query = emptyQuery, category = "" }
+    , { name = "Carrot cake (643g)", query = carrotCake, category = "Pâtisserie" }
+    , { name = "Épinards congelés (815g)", query = frozenSpinach, category = "Produit surgelé" }
+    , { name = "Pizza jambon fromage congelée (474g)", query = frozenPizzaHamCheese, category = "Produit surgelé" }
+    , { name = "Pizza margharita congelée (662g)", query = frozenPizzaMargarita, category = "Produit surgelé" }
+    , { name = "Ratatouille en conserve (804g)", query = cannedRatatouille, category = "Conserve" }
+    , { name = "Raviolis en conserve (733g)", query = cannedRaviolis, category = "Conserve" }
+    ]
```
n1k0, line 385: I'd naturally expect this to live in a JSON file because we already have the appropriate decoder so that would be basically matter of loading over HTTP on the Food page's `init`… I'm pretty sure that wouldn't be much more code and complexity than what this patch features, though I've been away from the codebase for a few weeks so I may be wrong.

src/Page/Food.elm
```diff
@@ -1317,19 +1361,29 @@ mainView session model =
...
+            , class "form-label fw-bold"
             ]
-            [ text "Carrot Cake" ]
+            [ text "Produit" ]
         , button
-            [ class "btn btn-outline-primary"
-            , classList [ ( "active", query == Query.emptyQuery ) ]
-            , onClick (LoadQuery Query.emptyQuery)
+            [ class "form-select ElementSelector text-start"
+            , id <| "selector-example"
```
n1k0, line 1376: Nit: unnecessary use of pipe operator

src/Page/Food.elm
```diff
@@ -1317,19 +1361,29 @@ mainView session model =
...
+            , style "overflow" "hidden"
+            , style "white-space" "nowrap"
```
n1k0, line 1378: As we have applied a custom `ElementSelector` class, shouldn't it rather be updated/extended/specialized to feature these inline styles?

src/Page/Food.elm
```diff
@@ -1317,19 +1361,29 @@ mainView session model =
...
+            , onClick (SetModal (SelectExampleModal autocompleteState))
+            ]
+            [ span
+                [ style "display" "block"
+                , style "overflow" "hidden"
```
n1k0, line 1383: Same

src/Data/Food/Query.elm
magopian: Sure, we could, but what good would it do us? What's the drawback of having it as elm data?
Having it as elm data:
+ no cpu used to decode
+ no need for another json loading management
- makes the base bundle slightly larger, even for users that will never use the food simulator
- needs some manual typing the first time the example is added in the codebase (compared to simply copy/pasting the json from the "share" tab in the simulator

At the moment we have textile examples as elm structures in `src/Data/Textile/Inputs.elm`, so I guess for coherence it would mean also converting those to json files loaded on demand?

n1k0: > At the moment we have textile examples as elm structures in src/Data/Textile/Inputs.elm, so I guess for coherence it would mean also converting those to json files loaded on demand?

Yes. In the same spirit, why don't we store processes, products, materials, ingredients and so on as Elm code directly? 

In my humble opinion, storing data in external JSON files kinda officially marks them as external data sources, which helps them identifying them as such and designing tooling for administrating and handling them. Maybe tomorrow we might want an example editor? Maybe we'd like offering people to contribute recipes through the github UI? 

Sure these ideas are not "official" requirements from POs, but it basically costs **nothing** to make it possible with very few constraints… I also won't debate on the "CPU parsing cost" argument as I'm confident it's pretty much negligible, and the "another json loading management" one is basically three lines of code as we're already loading a bunch of them through the codebase and shouldn't be much of a big deal.

I won't die on this hill though so I'll let you decide whatever you eventually want to do with this simple idea.

n1k0 (approved):

src/Data/Food/Query.elm
magopian: I have a feeling that this will cost a bit more than **nothing** to change, so I'll land this one, and start a new PR straight away to experiment and see where this goes.
Thanks for the feedback

PR #414, opened 2023-12-04
Make material process waste field a Unit.Ratio.

n1k0:
Couldn't use a `Split` because "Fil de soie" waste ratio is 700% while `Split` maxes at 100%.


magopian (approved):
LGTM, 🚢

PR #418, opened 2023-12-06
Food queries in json

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=e510369577854cd1affff5a1b4dbb0a5&pm=s)


magopian:
It's obviously a bit more "fiddly" this way, as we can't rely on a default "carrot cake" product to always be there, but I hope that the ergonomics work ok like that, let me know what you think.

src/Data/Food/Query.elm
```diff
@@ -373,52 +375,64 @@ type alias Product =
...
-recipes : List Query
-recipes =
-    recipesAndNames
-        |> List.map .query
+decodeProduct : Decoder Product
+decodeProduct =
+    Decode.succeed Product
+        |> Pipe.required "name" Decode.string
+        |> Pipe.required "query" decode
+        |> Pipe.required "category" Decode.string
```
n1k0, line 388: Shouldn't these categories be typed instead of strings?

public/data/food/products.json
```diff
@@ -0,0 +1,248 @@
...
+          "code": "AGRIBALU000000003104019",
+          "mass": 105
+        }
+      ],
+      "distribution": "ambient",
+      "preparation": [
+        "refrigeration"
+      ]
+    },
+    "category": "Pâtisserie" }
```
n1k0, line 36: How about formatting this JSON file using the same format as other JSON data files (with prettier)? Also related: do you use the git pre-commit hook checking for JSON formatting? If so, why didn't it prevent you to commit this file?

n1k0: I would also put the `category` property just after the `name` one, so we start each object with two short lines and a big `query` blob, which may makes things slightly more legible (here the `category` field is a bit hidden).

src/Page/Food.elm
```diff
@@ -294,6 +308,43 @@ update ({ queries } as session) msg model =
...
+        ProductsLoaded (RemoteData.Success products) ->
+            let
+                initialQuery =
+                    if model.initialQuery /= Query.emptyQuery then
+                        model.initialQuery
+
+                    else
+                        -- If we didn't load a specific query, then default to the carrot cake... if we've loaded it!
+                        products
+                            |> List.filter (\product -> product.name == "Carrot cake (643g)")
```
n1k0, line 320: How about picking the first product in the list instead, ensuring it's always the carrot cake? Or alternatively, adding a `priority` field to have the desired default product identified as such? Relying on the label seems  unnecessarily prone to desyncing.

n1k0 (approved):
LGTM

> @n1k0 if we still feel this change is worth it, i'll start a new PR to do the same for the textile examples, and see where it takes us. Or maybe add onto this one, so we decide for both at the same time?

I see value (now and in the future) to pursue this work (here or in another PR, your call), but an important point is if don't see any value doing so, feel free to stop the spike here and move to something you identify as more important or useful.

magopian:
I'm actually uncomfortable at the way i've duplicated the carrotCake in the `tests/FoodFixtures.elm` file. It feels redundant, and once we move the textile products in a json file, we'll need to also add a `tests/TextileFixtures.elm` duplicating:
- tShirtCotonFrance
- tShirtCotonIndia
- tShirtCotonAsie
- jupeCircuitAsie

To remove the duplication, we should load the products.json file in the tests... but we can't do that with elm-tests, so we'd switch to using a static db as we're already doing for all the "dynamic" data (impacts, countries, processes...).

But to me the main issue with moving away from the current solution is that we're losing readily available products that we can use as default values in the dropdowns and in the tests.

Also, as we can see in [Inputs.elm](https://github.com/MTES-MCT/ecobalyse/blob/fab163a684242be99b29f0c34959edc18b4d0daf/src/Data/Textile/Inputs.elm#L930-L935) we can take advantage of "elm code" to factorize some data, which makes it easier to maintain (at least in the short term?).

And I don't see any upside to that, so I'll drop this, and close it for now, until maybe one day we need to either
So I feel that this is not worth pursuing until we some day need to either
- update the products frequently (modifying the current ones, or adding/deleting some)
- have several dozen/hundreds of them, which would make the code more difficult to maintain
- open the products to be crowd sourced (it's indeed easier to contribute to json than to a typed elm structure)

Feel free to pick this up if you feel strongly about it though!

n1k0:
no no it's ok, i don't have strong arguments here, closing is fine

PR #419, opened 2023-12-11
Handle the Matomo site id as an env var.

n1k0:
Migrating our use of Matomo to a new instance with a different site id has been tedious. This patch introduces a new env var for storing the instance host and site id as environment variables.


magopian (approved):
LGTM, will make our life easier the next time the matomo instance has to be changed 😉

PR #420, opened 2023-12-11
Replace refrigeration with freezing for frozen food products

magopian:
Fix the frozen products so they have the proper freezing process instead of refrigeration (asked by Pascal and Lisa)


magopian:
Validated by Pascal 👍

PR #421, opened 2023-12-11
Fix the textile material step link to the docs

magopian:
[user story](https://www.notion.so/Le-lien-vers-la-doc-de-l-tape-mati-res-pointe-vers-celle-de-la-filature-ac53b102540f429cbe7b866489a41658)


n1k0 (approved):
My bad, missed the new `TextileMaterial` entry. LGTM.

PR #423, opened 2023-12-12
Better docs for env variables, decode flags

magopian:
Just an idea, what do you think @n1k0 ?


magopian:
Closed in favor of #424

PR #424, opened 2023-12-12
Improve docs and checks for required env vars.

n1k0:
@magopian Not sure if I prefer this or your approach in #423, which seems more robust 🤔

If we go the #423 route, ideally we would have a `Maybe Matomo` so we could disable stats in dev and only have them in production when env vars are properly set. What do you think?


magopian (approved):
This is simpler, go for it!

🚢

PR #431, opened 2023-12-20
Improve textile examples autocomplete.

n1k0:
A very simple enhancement and a small bugfix. 

@magopian I was also thinking of searching into the category name as well, because atm if you search for "polo" it shows no results, while we might want to expose t-shirt examples from the *t-shirt/polo* category as a starting point… what do you think?

### DIFF
--- a/src/Data/Textile/Inputs.elm
+++ b/src/Data/Textile/Inputs.elm
@@ -897,6 +897,7 @@ exampleProductToCategory q =
 exampleProducts : List Query
 exampleProducts =
     productsAndNames
+        |> List.sortBy .name
         |> List.map .query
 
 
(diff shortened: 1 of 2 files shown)


magopian:
Yes indeed it would make sense to be able to search on the "category" (product type) also 👍

n1k0:
@magopian ok I'll implement it in this patch

src/Data/Textile/Inputs.elm
```diff
@@ -1049,7 +1051,7 @@ pullLaineAsie =
 jupePolyesterAsie : Query
 jupePolyesterAsie =
     { jupeCotonAsie
-        | materials = [ { id = Material.Id "pa", share = Split.full, spinning = Nothing, country = Nothing } ]
+        | materials = [ { id = Material.Id "pet", share = Split.full, spinning = Nothing, country = Nothing } ]
```
magopian, line 1054: ah, well spotted, was an error on my end 👍

magopian (approved):
Makes sense to order alphabetically instead of "semantically" (all the asia products first, then all the french products and so on...)

n1k0:
Searching from categories looks more complicated than initially considered, I'm landing this and will file a card so we don't forget about it

PR #434, opened 2023-12-20
Refactor durability.

n1k0:
[User Story](https://www.notion.so/Impl-menter-le-coefficient-de-durabilit-578c7257a3be43dc9bff0b97546ee53b)


src/Data/Textile/Inputs.elm
```diff
@@ -760,8 +743,8 @@ encodeQuery query =
     , ( "countryDyeing", query.countryDyeing |> Country.encodeCode |> Just )
     , ( "countryMaking", query.countryMaking |> Country.encodeCode |> Just )
     , ( "airTransportRatio", query.airTransportRatio |> Maybe.map Split.encodeFloat )
-    , ( "quality", query.quality |> Maybe.map Unit.encodeQuality )
-    , ( "reparability", query.reparability |> Maybe.map Unit.encodeReparability )
+
+    -- , ( "durability", Unit.encodeDurability query.durability )
```
magopian, line 747: Is this commented out on purpose? Do we plan on decoding it later?

src/Page/Textile/Simulator.elm
```diff
@@ -614,24 +629,44 @@ massField massInput =
...
+                , id "durability-field"
+                , class "d-block form-range"
+                , title "Un double-clic réinitialise la valeur"
+                , onInput
+                    (String.toFloat
+                        >> Maybe.map Unit.durability
+                        >> Maybe.withDefault Unit.standardDurability
+                        >> updateDurability
+                    )
+                , onDoubleClick (updateDurability Unit.standardDurability)
```
magopian, line 653: Clever, even if maybe redundant with the "reset" button (admitedly very far at the bottom).
Should we replicate this UX on all the sliders, in a future PR? If so, let's create a Notion card?

magopian (approved):
Excellent work, it looks pristine!
Would it make sense to only display the custom info if the score is different? Eg, only if the durability has changed?

Also, it seems you've gone with "option A" from the user story, is that final?
BTW I like the "custom info" display better than having an extra ecoscore without the durability, imho it makes the information more readily available.

src/Page/Textile/Simulator.elm
n1k0: This is the behavior in a few software I use, eg. Adobe Lightroom, where double-clicking a range slider resets it. I'm not entirely sure it's a general well-known behavior so I'm not sure if we should generalize it or not… I've created [a card](https://www.notion.so/G-n-raliser-la-r-initialisation-au-double-click-sur-les-rangesliders-227ae641eb8e4cfa9fdbce502410714b) about it.

PR #452, opened 2024-01-18
Add a textile product type selector

magopian:
[user story](https://www.notion.so/Repositionner-le-type-de-produit-dans-la-phase-d-usage-e7ad8846ff13420a92301ec081b198ce)

### DIFF
--- a/src/Page/Textile/Simulator.elm
+++ b/src/Page/Textile/Simulator.elm
@@ -620,12 +620,12 @@ selectProduct autocompleteState ( model, session, _ ) =
                 |> Maybe.withDefault Inputs.defaultQuery.product
 
         currentQuery =
-            model.initialQuery
+            session.queries.textile
 
         updatedQuery =
             { currentQuery | product = product }
     in
-    update session (SetModal NoModal) { model | initialQuery = updatedQuery }
+    update session (SetModal NoModal) model
         |> updateQuery updatedQuery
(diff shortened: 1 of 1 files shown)


src/Page/Textile/Simulator.elm
```diff
@@ -506,6 +538,13 @@ commandsForNoModal modal =
                     |> Task.attempt (always NoOp)
                 ]
 
+        SelectProductModal _ ->
+            Cmd.batch
+                [ Ports.removeBodyClass "prevent-scrolling"
+                , Dom.focus "selector-product"
+                    |> Task.attempt (always NoOp)
```
n1k0, line 545: This doesn't seem to set the scroll position at the right height, the jump making hard to understand why it's happening and losing context

n1k0 (commented):
LGTM with a minor issue with autoscrolling after selecting a choice

src/Page/Textile/Simulator.elm
n1k0: Sure thing, here's a screecast

Uploading REC-20240122144003.mp4…

n1k0:
Okay as discussed on Mattermost I'm gonna take over this patch and land it, I'll fix the remaining scrolling bug on a next PR. No big deal meanwhile as we're on v2, not in production.

PR #455, opened 2024-01-22
Fix tomato id

magopian:


ccomb:
The same fix should be applied on frozenPizzaHamCheese

ccomb (approved):

PR #461, opened 2024-01-25
Only aggregated impacts on the client when logged out

magopian:
This is a spike for [user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=d342f0a524f4475f8a1fe147836327bd&pm=s)


package.json
```diff
@@ -17,12 +19,12 @@
     "server:debug": "elm make src/Server.elm --output=server-app.js && nodemon server.js --config nodemon.json",
     "server:start": "node server.js",
     "start:parcel": "parcel serve index.html --no-cache",
-    "start:dev": "npm run build:init && concurrently -n \"watch,parcel,server\" -c \"green,cyan\" \"npm run start:parcel\" \"npm run server:dev\"",
```
ccomb, line 20: Is removing the watch option intended?

src/ComputeAggregated.elm
```diff
@@ -0,0 +1,117 @@
...
+        |> Decode.decodeString decoder
+        |> Result.map
+            (\processes ->
+                processes
+                    |> List.map
+                        (\process ->
+                            { process | impacts = Impact.updateAggregatedScores definitions process.impacts }
+                        )
+            )
+
```
ccomb, line 31: slighltly shorter version:
```
        |> Result.map  
            (List.map                                                                                                                                                                                                                         
                (\process ->                                                                                                                                                                                                                  
                    { process | impacts = Impact.updateAggregatedScores definitions process.impacts }                                                                                                                                         
                )                                                                                                                                                                                                                             
            ) 
```

package.json
magopian: My understanding is that the `watch` here is the **name** of a process that `concurrently` is running... but there are three **names** and only two **processes** (`start:parcel` and `server:dev`).

So I believe `watch` is just a leftover from some previous code.

ccomb (approved):
Impressive work, GG!
I feel like it would be a little simpler or shorter to compute both `processes.json` and `processes_impacts.json` during export. (even if I trust elm code much more than python)

paulboosz:
> Impressive work, GG! I feel like it would be a little simpler or shorter to compute both `processes.json` and `processes_impacts.json` during export. (even if I trust elm code much more than python)

I feel the same as we would have to run only `export.py` instead of `export.py` + `npm run computeAggregated`

I can make the necessary modifications to `export.py` if needed

magopian:
I'm fine with doing the computation in export.py. The bonus with doing it this way would be to have the computation code only in one place, instead of having to maintain it in python AND in elm. I'll let you decide what's best, you guys are the ones responsible for this data management 👍

paulboosz:
> I'm fine with doing the computation in export.py. The bonus with doing it this way would be to have the computation code only in one place, instead of having to maintain it in python AND in elm. I'll let you decide what's best, you guys are the ones responsible for this data management 👍

All in one place 🙌 All in one place 🙌

bin/build-db
```diff
@@ -6,6 +6,10 @@
  * is impossible in an Elm test environment.
  */
 const fs = require("fs");
+const textileProcesses = process.env.TEXTILE_PROCESSES || "public/data/textile/processes.json";
+const foodProcesses = process.env.FOOD_PROCESSES || "public/data/food/processes.json";
```
n1k0, line 10: What's the use case for using different output targets? Debugging?

n1k0 (commented):
This works, though I think we'll have to figure out how to improve processes loading and updating mechanisms…

src/Data/Food/Process.elm
```diff
@@ -238,6 +287,34 @@ decodeStringUnit =
...
+        "kg" ->
+            Encode.string "kilogram"
+
+        "km" ->
+            Encode.string "kilometer"
+
+        "kWh" ->
+            Encode.string "kilowatt hour"
+
+        "l" ->
```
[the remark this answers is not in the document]
magopian: it comes from the way we are [currently decoding the unit](https://github.com/MTES-MCT/ecobalyse/pull/461/files#diff-dcc8167aab4dea154209ef3e3faadf06455df08620266c5f34d675efe94844c3R277), which in turn comes from the way we get the data in the first place I'd say?

bin/build-db
magopian: I'm not sure what you mean by that, we store the food and textile processes in their own files, because they don't have the same structure, and it wouldn't make much sense to merge them both in the same file and use a `Decode.oneOf`.

Am I missing something?

magopian: We need to output EVERYTHING in those files for the test run (all the impacts), but not for the normal build ;)

src/Data/Food/Process.elm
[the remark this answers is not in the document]
ccomb: you're right I've seen `l` in the databases :-(

compute-aggregated.js
```diff
@@ -0,0 +1,38 @@
...
+const elmApp = Elm.ComputeAggregated.init({
+  flags: {
+    definitionsString: fs.readFileSync("public/data/impacts.json", "utf-8"),
+    textileProcessesString: fs.readFileSync("public/data/textile/processes_impacts.json", "utf-8"),
+    foodProcessesString: fs.readFileSync("public/data/food/processes_impacts.json", "utf-8"),
+  },
+});
+
+const exportJson = (filePath, json) => {
+  fs.writeFileSync(filePath, JSON.stringify(json, null, 2) + "\n");
```
n1k0, line 14: It would be nice to format the resulting JSON serialization with the prettier coding styles to avoid unwanted diffs; I did that for the examples editor here: https://github.com/MTES-MCT/ecobalyse/blob/3ca9f18f9a39b7e4bb8ed98d40eaff3a7e6c3fda/lib/github.js#L32-L34

package.json
```diff
@@ -8,7 +8,9 @@
   "scripts": {
     "build": "npm run server:build && rimraf dist && npm run build:init && parcel build index.html --public-url ./",
     "build:init": "./bin/update-version.sh && mkdir -p dist && cp -r public/* dist/ && npm run db:build",
+    "computeAggregated": "elm make src/ComputeAggregated.elm --output=compute-aggregated-app.js && node compute-aggregated.js",
     "db:build": "./bin/build-db",
+    "db:buildFull": "TEXTILE_PROCESSES=public/data/textile/processes_impacts.json FOOD_PROCESSES=public/data/food/processes_impacts.json ./bin/build-db",
```
n1k0, line 13: Nit: `build-full` for casing consistency

src/ComputeAggregated.elm
```diff
@@ -0,0 +1,115 @@
...
+                                decodeProcesses (FoodProcess.decodeList Impact.decodeWithoutAggregated) definitions foodProcessesString
+                        in
+                        Result.map2
+                            (\textileProcesses foodProcesses ->
+                                let
+                                    textileProcessesOnlyAggregated =
+                                        textileProcesses
+                                            |> keepOnlyAggregated
+
+                                    foodProcessesOnlyAggregated =
```
n1k0, line 73: Nit: How about moving this `toExport` fn out of the `let..in` statement, for legibility?

src/Main.elm
```diff
@@ -295,6 +298,57 @@ update rawMsg ({ state } as model) =
                 ( VersionPoll, _ ) ->
                     ( model, Request.Version.loadVersion VersionReceived )
 
+                -- Login
+                ( LoggedIn (Ok newProcessesJson), _ ) ->
+                    let
+                        newSession =
+                            Session.loggedIn session newProcessesJson
+                                |> Session.notifyInfo "Vous avez maintenant accès au détail des impacts" ""
```
n1k0: For some reason, this text is rendered in a very large fashion in the info notification

image

It uses a `h5` while it should use a `h6`, and needs removing the bottom margin by adding the `mb-0` class to the element.

Also if you log in and out multiple times, the same notification stacks

image

To solve this (and if we want to, I reckon we do) we could either make them disappeear after a short while or avoid adding another instance of an already present notification (eg. make the notification list a Set)

Note: this shouldn't block landing this PR. But please file a card so we don't forget to fix it.

src/Static/Db.elm
```diff
@@ -39,60 +26,39 @@ type alias Db =
...
-    Result.map5 Db impactDefinitions textileDb foodDb countries distances
+db : StaticJson.Processes -> Result String Db
+db procs =
+    StaticJson.db procs
+        |> Result.andThen
+            (\{ foodDb, textileDb } ->
+                Result.map3
+                    (\okImpactDefinitions okCountries okDistances ->
+                        Db okImpactDefinitions textileDb foodDb okCountries okDistances
+                    )
```
n1k0, line 37: Nit: this can be `(\okImpactDefinitions -> Db okImpactDefinitions textileDb foodDb)`

src/Views/Comparator.elm
```diff
@@ -152,11 +152,18 @@ comparatorView config =
...
+            [ ( "Sous-scores", Subscores )
+            , ( "Impacts", IndividualImpacts )
+            ]
+
+        else
+            []
+       )
+        ++ [ ( "Étapes", Steps )
+           , ( "Total", Total )
+           ]
```
n1k0, line 165: By default on the Food page with an empty product, we have an empty single "Étapes" tab with nothing at all under it:

image

We shouldn't render anything if we don't have anything to render yet.

tests/e2e-textile.json
```diff
@@ -176,7 +176,7 @@
       "tre": 0.13206487622856444,
       "wtu": 2.0035853668394106,
       "ecs": 859.7119944122998,
-      "pef": 875.7274702532203
+      "pef": 875.7274702532204
```
n1k0, line 179: No big deal but do you have any clue why these have changed? Looks like floating point issues but can't track it to a specific reason.

n1k0 (approved):
Ok, this patch is *huge* and I certainly couldn't review *all* of the +33k/-14k LoC changed (especially the ones in JSON obviously), but green tests make me confident it's allright from an impacted source data validity standpoint.

Focusing on testing the patch functionally, everything works as expected and we don't load any IP-protected impact data over HTTP when we're not logged in, so we're good I think.

I feel like I may be missing interesting questions especially regarding JSON data management on @ccomb's side though as he already approved the patch I assume everything's clear here.

src/Views/Comparator.elm
magopian: Indeed. As it's the case also on the current v2, I'll create a notion card to address that at a later stage, thanks for the heads up.

src/Main.elm
magopian: https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=27110f9f18914855afd272ee51bfc1dd&pm=s

magopian: @n1k0 did you by any chance already fix that?

n1k0: @magopian nope I don't think so

magopian: I don't see any h5 or h6 anymore on `staging`, and i can't reproduce the multiple notifications, maybe the changes you did to "make the UI nicer" in the `auth` pages?

PR #470, opened 2024-01-30
Complete durability index computation.

n1k0:
[User Story](https://www.notion.so/D-tailler-dans-la-V2beta-le-calcul-de-l-indice-de-durabilit-1fa1eb1d1a924178bbe56bad2207c8c1)

- [x] Add a new `business` field, compute related index
- [x] Add a new `traceability` field, compute related index
- [x] Compute materials type index
- [x] Apply weighting
- [x] Update API params, docs, add changelog entry
- [x] Update indices boundaries


magopian (approved):
This looks NICE, congratz!
I'm not sure I understand how the computation works though, it seems most of the parameters only matter inside a small "window" of meaningful values? Eg whenever a price is lower than 20 or higher than 30 then it's not changing anymore. Same for the other two parameter values. Is that on purpose?

n1k0:
> it seems most of the parameters only matter inside a small "window" of meaningful values?

@magopian that's exactly the specs indeed :) thanks for the review!

PR #476, opened 2024-02-05
Experiment: computation on the backend

magopian:
[user story](https://www.notion.so/fd0068a7db8348c584cd53a3eea399f2?v=af2332265058414da4bd3aa4d381ed5f&p=ce3d1eaae349449fa0ebd1a37a03fbb0&pm=s)

- [x] use simulations queried from the API for textile: it stores simulations in the textile bookmarks instead of simple queries
- [x] use simulations queried from the API for food


magopian:
Ok, so this should now be ready for review. The idea is that instead of the textile and food simulator calling the `compute` themselves, they instead request the computation from the API (using `getCompute`).

This doesn't use debouncing (yet?), we might want to add that if we see too much lag.
Also, if this works as expected, we'll have to track down each and every "computation" that's done in the views, and instead use the computation results from the API (maybe adding some more data if needed).

vjousse:
It was a sprint spike to test a POC. `master` has diverged a lot since this PR so it's better to close it.
If needed, we will reopen another one to continue the work done here.

PR #473, opened 2024-02-01
Allow adjusting ecotox weighting.

n1k0:
[User Story](https://www.notion.so/Rendre-param-trable-la-pond-ration-de-l-cotox-894d42e217c6448a883346203dff8db4)

PR #479, opened 2024-02-08
Static DB split

ccomb:
Extract countries, distances and definitions from textile db and split them into several components. It should allow to better dissociate food and textile countries.


magopian (commented):
This is a huge work! Impressive!

It does seem that this PR breaks the [ecotox weighting customization](https://github.com/MTES-MCT/ecobalyse/pull/473) (the input below the blue box at the top right in the textile and food simulators).

I'm not sure why we would not want to continue using the DB stored in the session? It's shared accross all the pages, and can be updated (to update the ecotox weighting, or to update the processes' impacts in the ["auth spike"](https://github.com/MTES-MCT/ecobalyse/pull/461)).

That being said, I like the idea of splitting the "common" DBs out, which might make future changes and maintenance easier.

magopian (approved):
Looks good, after a very quick second review, thanks for the changes 🙏

PR #462, opened 2024-01-25
Add the missing "fabricProcess" parameter to the API example

magopian:
Quick fix, missing `fabricProcess` parameter in the API example

PR #485, opened 2024-02-15
add-missing-fabricProcess-api-example

paulboosz:
same as https://github.com/MTES-MCT/ecobalyse/pull/462 but merge into master

Fix 
image


magopian (approved):
Yep, thanks! And I created a card on notion to add an automatic test for that ;)

PR #486, opened 2024-02-19
Update durability computation

n1k0:
[User Story](https://www.notion.so/Modification-de-la-formule-largeur-de-gamme-dans-le-calcul-de-durabilit-97503fdceb014386bf01540aec9e7efe)


ccomb (approved):
Difficult to see the details through the UI but the value seem to change at the right thresholds

PR #488, opened 2024-02-20
EI heat process

ccomb:
[story](https://www.notion.so/Int-gration-Ecoinvent-Mix-Chaleur-f250a03d16754bd2b191bc3738f1b4cc)


public/data/textile/processes.json
```diff
@@ -8458,7 +8458,7 @@
   },
   {
     "search": "",
-    "name": "(EI) Heat, natural gas, France",
+    "name": "(EI) Heat mix (FR)",
```
n1k0, line 8461: As the impacts didn't change, does that mean the previous label was wrong?

src/Data/Textile/Inputs.elm
```diff
@@ -139,7 +137,6 @@ type alias Query =
     , fading : Maybe Bool
     , dyeingMedium : Maybe DyeingMedium
     , printing : Maybe Printing
-    , ennoblingHeatSource : Maybe HeatSource
```
n1k0, line 142: This is a breaking change for the API, so we must add an entry to the Api changelog. Also, we should update the `openapi.yml` docs accordingly.

public/data/textile/processes.json
n1k0: Scratch this, code changed meanwhile.

src/Data/Textile/Inputs.elm
ccomb: Sure, thx for noticing

n1k0 (approved):
LGTM 👍

PR #489, opened 2024-02-20
Remove confusing DOM element autofocusing behavior.

n1k0:
The auto-DOM-scrolling strategy when adding or updating examples/ingredients in the Food calculator makes it really hard and counter-intuitive to use. This patch removes all related code so the behavior is back to normal (no autoscroll at all).


ccomb (approved):
Didn't find the scroll you're talking about but I suppose the change relevant

PR #503, opened 2024-02-27
Split the textile inputs & query modules.

n1k0:
[Tech Debt] This patch splits the textile `Query` stuff from `Inputs` and moves it to a new dedicated module. This avoids an annoying cycling dependency issue when importing the static `Db` in the `Inputs` module. Also this matches what is done for Food with the `Recipe` and `Query` modules.


n1k0:
I'm landing this because I need the patch to implement something else that requires it.

PR #508, opened 2024-02-28
Track scores evolution

n1k0:
[User Story](https://www.notion.so/Mettre-en-place-un-outil-process-de-suivi-des-volutions-7bdf9a73024d4c889138a74dd1bab96d)


n1k0:
I'll probably give this implementation another go when the card is repriorized.

PR #513, opened 2024-03-04
Allow editing example products.

n1k0:
[User Story](https://www.notion.so/Permettre-d-diter-ou-de-proposer-de-nouveaux-exemples-textiles-8780d1a337b747c1b4b2419ab537a139)


n1k0:
This diff on this has getting totally out of hands and I prefer closing. We shall revisit this patch the day we'll *actually* want that feature.

PR #517, opened 2024-03-06
Compute ecosystemicServices for animal products

paulboosz:
[User Story](https://www.notion.so/Int-gration-en-V2Beta-du-calcul-des-services-cosyst-miques-pour-les-produits-animaux-a6b279aefe2b4b7f8dcafaea06e38537?pvs=4)


src/Data/Food/Ingredient/Category.elm
paulboosz, line 1: @ccomb to make a test pass I have to add a value for my "dummy" category. What do you think ? Is there another way ?

ccomb: It looks weird but since we have a dummy category anyway I suppose we have to add it on the client side. I admit I'm feeling a bit lost on the ES computation.

PR #525, opened 2024-03-08
Store split internal value as float

n1k0:
This avoids **internally** rounding percentages while they should not.


ccomb (commented):
Since Splits look like a fraction, couldn't it be implemented as two Int instead of a Float? Such as Split Int Int (33% would be Split 1 3)

n1k0:
@ccomb I think we xwant to keep being able to express splits from floats like `airTransportRatio = Split.fromFloat 0.4576` and finding the equivalent fraction 1. isn't easy and 2. isn't useful.

paulboosz:
Yay for always using float 🙌

n1k0:
Landing with r=@paulboosz

ccomb:
This is because I've long ago been definitely traumatized by the usage of floats in management or accounting applications and the associated unsolvable rounding issues :-/ But ok we're not in accounting.

PR #530, opened 2024-03-20
Update wear duration and wash cycles from computed durability.

n1k0:
[User Story](https://www.notion.so/Moduler-la-dur-e-d-utilisation-en-fonction-de-la-durabilit-397aea2046e2465b94cdaa786e920e80)


magopian (approved):
Looks good to me 👍 
🚢

PR #538, opened 2024-03-27
fixed road using Train

ccomb:


n1k0:
@ccomb I can't see any usage for the `trainTransport` wellKnown process introduced by this patch; am I missing something?

PR #547, opened 2024-03-27
Fix transport bug between "Inconnu" and "Inde"

paulboosz:
modified script `transports.py`: 
- add regions (eg. Asia : RAS,...)
- add 500 km self-distance in distance matrix (distance China-China, RAS-RAS,...)
- test internal coherence : every country/region has a distance with every other country/region
- test coherence with `countries.json`: {set of countries/regions in `transports.py`} == {set of countries/regions in `countries.json` } 

TODO :
- [x] remove 500km self distance rule from elm code


data/common/distances/CountryDistances.py
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(self, distances):
+        self.distances = self._convert_distances(distances)
+        self._all_countries = set()
+        self._update_all_countries()
+
+    self_distance = {
+        "road": 500,
+        "sea": None,
+        "air": 500,
+    }
```
paulboosz, line 11: Not entirely sure about this, should self_distance (the distance from countryA to countryA) be that ?

n1k0: I'm slightly surprised that air and road distances are the same but tbh the final impact ratio for these is marginal so we probably shouldn't care too much

n1k0: (otherwise 500km for default transportation within a same country is what's currently used in the Elm app)

tests/Data/TransportTest.elm
```diff
@@ -53,6 +55,32 @@ suite =
...
+                                )
+                            |> List.length
+                            |> Expect.equal 0
+                  in
+                  describe "Data coverage exhaustivity checks"
+                    [ testAllCountriesCoveredFor Scope.Food
+                        |> asTest "should always give a distance greater than 0 between two food countries"
+                    , testAllCountriesCoveredFor Scope.Textile
+                        |> asTest "should always give a distance greater than 0 between two textile countries"
+                    ]
```
n1k0, line 83: Note: I've added a test to ensure we cover all food and textile country distances.

n1k0 (approved):
I've mostly scanned the Python code and thoroughly tested the patch locally and its behavior in the UI, which works as expected. Still, I added an Elm test to ensure all countries are covered with distances, so the build would fail on failure and we would be noticed about it.

Maybe @cccomb or @magopian would like to review the Python code?

public/data/transports.json
```diff
@@ -1,1021 +1,1946 @@
...
+      "sea": 8753,
+      "air": 7294
+    }
   },
   "BR": {
-    "CH": { "road": null, "sea": 8596, "air": 8857 },
-    "CN": { "road": null, "sea": 25166, "air": 16580 },
-    "CZ": { "road": null, "sea": 10074, "air": 9489 },
-    "DE": { "road": null, "sea": 9006, "air": 9122 },
-    "EG": { "road": null, "sea": 10472, "air": 10025 },
```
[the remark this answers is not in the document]
paulboosz: That would be much better indeed. Corrected

n1k0:
@paulboosz I've also removed the [programmatic handling of default distances inside a same country](https://github.com/MTES-MCT/ecobalyse/pull/547/commits/e476412d1abea18d4f876bc912b9ead2c979678d), so code now entirely relies on provided JSON data.

n1k0:
I heavily suggest everybody heavily test the patch on the deployed staging branch, as it's potentially impactful on Food as well.

n1k0:
From the conversation on Mattermost, I think this patch is in a good place for landing to v2. @paulboosz would you mind merging?

PR #548, opened 2024-03-27
Make product examples generic.

n1k0:
### This patch

- [x] Backports part of the work done in #513 where we introduce a convenient generic `Example query` type with associated views and optimize performances in the Examples Explorer page.
- [x] Uses actual UUIDs with `NoRedInk/elm-uuid` to decode and validate them (and later generate them like in #513)
- [x] Makes the Examples tab the default ones in both food and textile explorers (even with the food section being currently hidden, but still available at `/#/food` for validation)
- [x] Moves `Page.Textile.Simulator` to just `Page.Textile` to simplify and clarify module organization


magopian (approved):
Good ideas, LGTM
🚢

PR #549, opened 2024-03-28
Break recursion in distance computation

ccomb:
Don't recurse and display NaN pts if a distance is missing instead of failing in infinite recursion or silently defaulting to zero

image


src/Data/Transport.elm
```diff
@@ -197,9 +211,19 @@ getTransportBetween scope impacts cA cB distances =
...
+                            else
+                                getTransportBetween True scope impacts cB cA distances
+                )
+            |> Maybe.withDefault
+                (if stopRecursion then
+                    erroneous impacts
+
+                 else
+                    getTransportBetween True scope impacts cB cA distances
                 )
```
n1k0, line 226: We should really rather have the two required checks performed at once and exposing the result in a `Result`:

```elm
getTransportBetween :
    Impacts
    -> Country.Code
    -> Country.Code
    -> Distances
    -> Result String Transport
getTransportBetween impacts cA cB distances =
    case
        ( distances |> Dict.get cA |> Maybe.andThen (Dict.get cB)
        , distances |> Dict.get cB |> Maybe.andThen (Dict.get cA)
        )
    of
        ( Just transport, _ ) ->
            Ok { transport | impacts = impacts }

        ( _, Just transport ) ->
            Ok { transport | impacts = impacts }

        ( Nothing, Nothing ) ->
            Err <|
                "No distance found between "
                    ++ Country.codeToString cA
                    ++ " et "
                    ++ Country.codeToString cB
```

If we really don't want to spend the required time to leverage a `Result`, the same versy simple approach can be used with a fallback:

```elm
getTransportBetween :
    Impacts
    -> Country.Code
    -> Country.Code
    -> Distances
    -> Transport
getTransportBetween impacts cA cB distances =
    case
        ( distances |> Dict.get cA |> Maybe.andThen (Dict.get cB)
        , distances |> Dict.get cB |> Maybe.andThen (Dict.get cA)
        )
    of
        ( Just transport, _ ) ->
            { transport | impacts = impacts }

        ( _, Just transport ) ->
            { transport | impacts = impacts }

        ( Nothing, Nothing ) ->
            default impacts
```

src/Views/Format.elm
```diff
@@ -70,7 +70,10 @@ formatFloat decimals float =
             FormatNumber.format { frenchLocale | decimals = Exact dc }
                 >> String.replace "−" "-"
     in
-    if float == 0 then
+    if isNaN float then
+        "NaN"
```
n1k0, line 74: I'm super reluctant to display "NaN" as it's highly associated with poor code. I really insist we rather display a red "Error" string rather than NaN.

n1k0 (changes requested):

src/Views/Format.elm
n1k0: And while we're at testing if the score is `NaN`, we could as well render an explicit error message so users may have something to copy and paste when reporting the issue, which is likely to help maintainers a lot on the long run.

n1k0: AND, sorry to insist a lot, but we MUST at least file a tech-debt card to make this `getTransportBetween` fn returning a `Result` so we don't have to mess around with fake error results.

n1k0: AND sorry to be a pain in the butt.

ccomb: I agree returning a Result really is the better solution. This is option A.

For now in the short term we have the choice between :
- option B, the current proposal which is propagating a NaN up to display, without much harm (I've checked with bad data).
- option C, the current state, which is breaking the runtime with a weird behaviour when the distance data is not complete.

Which option do you prefer in the short term?

n1k0: At least replace the `"NaN"` text with either `"Err"` or `"---"`. I don't want `NaN` to be printed on the web page.

n1k0: btw I created the card I suggested so you don't have to do it https://www.notion.so/Afficher-une-erreur-l-utilisateur-en-cas-de-distance-entre-deux-pays-inconnue-e3866a56500d4dfd9c2fc1d3d44c58a1

PR #559, opened 2024-04-04
Restored e2e tests

ccomb:
with a few additions:
- added a way to display the error in case the tests is in 400 while it should be in 200
- don't write the output file if it's empty
- write the test anyway with empty impacts and score if the status is not 200


magopian (commented):
Good idea to not write the file if it's empty. I'd go one step further: if the status isn't 200, still write to the output file (not sure with which impacts), to make sure we never lose any tests?

tests/server.spec.js
```diff
@@ -544,8 +544,12 @@ afterAll(() => {
   // with their contents.
   function writeE2eResult(key) {
     const target = `${__dirname}/e2e-${key}-output.json`;
-    fs.writeFileSync(target, JSON.stringify(e2eOutput[key], null, 2) + "\n");
-    console.info(`E2e ${key} tests output written to ${target}.`);
+    if (e2eOutput[key].length === 0) {
```
magopian, line 547: I believe this shouldn't happen anymore, but it's still good to have I guess, to prevent any future accidents ;)

magopian (approved):
Not sure about the removal of the "market for electricity" from the processes_impacts.json, not sure if it's related to this current PR, and why we don't want it anymore (not used?).

tests/server.spec.js
```diff
@@ -577,6 +581,9 @@ async function expectListResponseContains(path, object) {
 }
 
 function expectStatus(response, code, type = "application/json") {
+  if (response.status === 400 && code != 400) {
```
n1k0, line 584: Nit: maybe rename the `code` arg to `expectedCode` as the function body is confusing to read

n1k0 (approved):
LGTM, thanks!

tests/server.spec.js
ccomb: The PT elec is unused, I'll remove it in another PR

PR #561, opened 2024-04-08
Remove obsolete textile use step related processes.

n1k0:
This is part of this [user story](https://www.notion.so/Migrer-la-consommation-lectrique-du-repassage-dans-les-donn-es-de-cat-gories-de-produits-e16d76f65b9b4f378e6475f18e8064b1)


paulboosz (approved):
Looks good !

PR #570, opened 2024-04-15
Fix style issue with the `selector-example`

EmileRolley:
Hi, there was a little style issue with the `selector-example`:

image

After:

image


n1k0 (approved):
Thanks!

PR #571, opened 2024-04-15
Prepare staging.

n1k0:
I did what I could.


n1k0:
I'm landing this. The django setup and auth doesn't work, but I addressed a bunch of tweaks needed on our staging instance.

@ccomb fixing the backend and auth stuff will have to be addressed on a separate PR

PR #573, opened 2024-04-17
Use API token in OpenAPI interactive documentation page.

n1k0:
This patch stores the user profile information in the session store, so we can pass it to the rapidoc component in the Api page, allowing to access detailed impacts when authenticated on the website.

### DIFF
--- a/src/Data/User.elm
+++ b/src/Data/User.elm
@@ -1,0 +1,28 @@
+module Data.User exposing
+    ( User
+    , decode
+    )
+
+import Json.Decode as Decode exposing (Decoder)
+import Json.Decode.Pipeline as Pipe
+
+
+type alias User =
+    { email : String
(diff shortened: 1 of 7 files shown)


src/Data/User.elm
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+
+decode : Decoder User
+decode =
+    Decode.succeed User
+        |> Pipe.required "email" Decode.string
+        |> Pipe.required "first_name" Decode.string
+        |> Pipe.required "last_name" Decode.string
+        |> Pipe.optional "organization" Decode.string ""
```
ccomb, line 28: This is a fix for the admin login bug, it's also fixed on python side by returning "" instead of null for the organization

ccomb (approved):
Looks very polished now :+1:

PR #574, opened 2024-04-17
Make organization a non-null, required model field.

ccomb:
More tests for login and profile retrieval. Don't store null for missing organization: align with elm expecting a mandatory field that can be blank


backend/authentication/forms.py
```diff
@@ -66,12 +76,13 @@ def get_token(self, user):
...
         super().save(commit)
         email = self.cleaned_data["email"]
-        for user in EmailLoginForm.get_users(self, email):
-            context = EmailLoginForm.get_mail_context(self, self.request, user)
-            EmailLoginForm.send_mail(self, email, context)
-            if "test" in sys.argv:
+        for user in MailauthEmailLoginForm.get_users(self, email):
+            context = MailauthEmailLoginForm.get_mail_context(self, self.request, user)
+            MailauthEmailLoginForm.send_mail(self, email, context)
+            if "test" in sys.argv or "backend:test" in sys.argv:
```
n1k0, line 85: If `test` is in `sys.argv`, it is as well in `backend:test`

backend/authentication/forms.py
```diff
@@ -16,6 +16,16 @@ def get_login_url(self, request, token, next=None):
...
 
+    def save(self):
+        """redefine just to transmit the login_url in test mode"""
+        email = self.cleaned_data["email"]
+        for user in self.get_users(email):
+            context = self.get_mail_context(self.request, user)
+            self.send_mail(email, context)
+            if "test" in sys.argv or "backend:test" in sys.argv:
+                # hack to pass the login url to the test suite
+                os.environ["login_url"] = context["login_url"]
```
n1k0, line 27: Wow, this is indeed an ugly hack! Environment variable to act as a global one in production code just for tests isn't a good pattern, can we imagine another strategy here? Maybe a settings? You know you can have different django settings modules for different execution environment?

n1k0 (changes requested):

backend/authentication/forms.py
ccomb: `sys.argv` is not a string but the list of command line arguments : `"test"` is in `["python", "manage.py", "test"]` but not in `["npm", "run", "backend:test"]`

backend/authentication/tests.py
```diff
@@ -68,17 +85,78 @@ def test_register_post(self):
...
+        # a successful login should redirect to the "next" url
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.url, "/")
+        # get json profile
+        response = self.client.get(reverse("profile"))
+        jsonresp = json.loads(response.content)
+        self.assertEqual(
+            list(jsonresp.keys())[:5],
+            ["email", "first_name", "last_name", "organization", "terms_of_use"],
+        )
```
n1k0, line 159: Why don't we test for the token here?

backend/authentication/forms.py
ccomb: We can replace a global variable with another, anyway this code is not executed in production. Do you see a better way to pass the login_url thrown in the console to the test suite?

ccomb: I'm investigating the `locmem` email backend that should allow to watch in mails to be sent

n1k0: Now I remember there's builtin django tooling for testing email sending and contents https://docs.djangoproject.com/fr/5.0/topics/testing/tools/#django.core.mail.django.core.mail.outbox

ccomb: Yes, the `locmem` backend is automatically used in testing. I pushed a change which looks in the generated mail (I cannot look in the original context that was used to generate the mail)

PR #575, opened 2024-04-17
Ensure loading detailed impacts from session store.

n1k0:
Right now on staging, refreshing the browser when authenticated keeps the authentication status but loses the detailed impacts. This patch ensures loading the detailed process impacts from localStorage when they're available at app init so there's no inconsistent behavior and authenticated users will always get their detailed impacts even if they close/reopen their browser.


ccomb (approved):
looks good, I tried to reload and disconnect

PR #580, opened 2024-04-23
Memoize processes on token

magopian:
Some benchmarks:

20 requests per second:

without_memoize_20rqs

with_memoize_20rqs

40 requests per second:

with_memoize_40rqs

50 requests per second:

with_memoize_50rqs

### DIFF
--- a/server.js
+++ b/server.js
@@ -136,6 +136,7 @@ api.all(/(.*)/, bodyParser.json(), async (req, res) => {
     }
   } catch (err) {
     console.error(err.message);
+    return res.status(500).send("Error while retrieving the processes");
   }
 
   elmApp.ports.input.send({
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -133,7 +142,7 @@ api.all(/(.*)/, bodyParser.json(), async (req, res) => {
     method: req.method,
     url: req.url,
     body: req.body,
-    processes: processes,
+    processes: result.processes,
```
n1k0, line 145: We can easily get `undefined` passed here if we fall into the `catch` clause a few lines before. We may want to return an error response in this `catch` clause and bypass the Elm app entirely when it happens.

n1k0 (approved):

server.js
n1k0: Perfect 👌

PR #586, opened 2024-05-06
feat: encrypt and decrypt textile processes.

n1k0:
WiP, spike, HIGHLY experimental:

- No detailed impacts versioned in git anymore, textile processes are encrypted using a secret `ENCRYPTION_KEY` known only by our team and the server environment.
- A new `npm run processes:decrypt` command is available to allow working on the JSON processes locally, provided you have set the `ENCRYPTION_KEY` env var correctly (which should match the one in production).
- A new `npm run processes:encrypt` command is provided to encrypt the processes file when you're done before pushing to a new branch for review/staging; `ENCRYPTION_KEY` is used as the secret key.
- `ENCRYPTION_KEY` should obviously never be published anywhere.
- Because it knows about `ENCRYPTION_KEY`, the newly introduced `lib/crypto.js` module should never be used in client code, only server one.

## TODO

- [x] add utility to encrypt and decrypt a JSON processes file
- [x] update the detailed textile processes file in the repo to be encrypted
- [ ] make the auth workflow serve the decrypted version of the processes file (pb is it's in Python…)
- [ ] make the API work with this encryption system
- [ ] ensure to never push a non-encrypted processes file; this could be a pre-commit hook…
- [ ] … which could also check that the commit diff **never** includes the value of `ENCRYPTION_KEY`

## Remaining questions

- How would this play with self-hosting? Could we make the decryption key the auth token? How? Would it make sense? Would it be secure?


n1k0:
Obsolete, closing.

PR #589, opened 2024-05-16
Auth final touches

magopian:


src/Data/Env.elm
```diff
@@ -21,6 +22,11 @@ betagouvUrl =
     "https://beta.gouv.fr/startups/ecobalyse.html"
 
 
+cguUrl : String
+cguUrl =
+    "https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire-draft/old/pages-textiles-old/conditions-dutilisation"
```
n1k0, line 27: It's a bit strange to link to page url containing the word "old". We should probably ask this page to move to something more permanent.

src/Page/Auth.elm
```diff
@@ -526,7 +526,10 @@ viewRegisterForm ({ user } as model) =
                                     )
                                 ]
                                 []
-                            , text "Je m'engage à ne pas exploiter les données pour une utilisation commerciale."
+                            , text "Je m’engage à respecter les conditions d’utilisation (cf. "
+                            , a [ href Env.cguUrl ]
+                                [ text "conditions d'utilisation" ]
+                            , text ")"
```
n1k0, line 532: We should probably use `Markdown.simple` here.

magopian: done

n1k0 (approved):

PR #607, opened 2024-05-27
Add a way to switch between regulatory and advanced modes.

n1k0:
[User Story](https://www.notion.so/Dissocier-deux-modes-d-affichage-r-glementaire-expert-09b4d9a097ea40438a7c4c2a6192520b)

### DIFF
--- a/src/Views/Textile/Step.elm
+++ b/src/Views/Textile/Step.elm
@@ -446,7 +446,12 @@ stepActions { current, showAdvancedFields, toggleStep } label =
         allowDisablingStep =
             showAdvancedFields
                 -- Regulatory mode only allow disabling Spinning, Fabric and Ennobling steps
-                || List.member current.label [ Label.Spinning, Label.Fabric, Label.Ennobling ]
+                || List.member current.label
+                    [ Label.Material
+                    , Label.Spinning
+                    , Label.Fabric
+                    , Label.Ennobling
+                    ]
     in
     div [ class "StepActions ms-2" ]
         [ div [ class "btn-group" ]
--- a/src/Data/Textile/Query.elm
+++ b/src/Data/Textile/Query.elm
@@ -220,9 +220,9 @@ isAdvancedQuery query =
         , query.disabledSteps
             |> List.any
                 (\label ->
+                    -- If these steps are disabled, it means we're in advanced mode
                     List.member label
-                        [ Label.Material
-                        , Label.Making
+                        [ Label.Making
                         , Label.Distribution
                         , Label.Use
                         , Label.EndOfLife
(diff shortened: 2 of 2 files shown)


paulboosz:
UX review :
- really clean and clear !
- The warning message when switching mode is really helpful
- tooltip message on hover on both mode are also helpful. I kind of expected the tooltip to be clickable with a nodal because the other one in the steps are (and link to the doc) but that's fine.
- Didn't find any bug when switching mode and playing with comparator

src/Page/Textile.elm
```diff
@@ -85,12 +86,19 @@ type Modal
     = NoModal
     | ComparatorModal
     | AddMaterialModal (Maybe Inputs.MaterialInput) (Autocomplete Material)
+    | ConfirmSwitchToRegulatoryModal
```
ccomb, line 89: This seems a bit weird to have such a behaviour on switching tabs. One usually expect tabs to display different informations. But anyway although weird it's clear with the confirmation modal.

src/Views/Textile/Step.elm
```diff
@@ -130,36 +127,37 @@ airTransportRatioField { current, updateAirTransportRatio } =
...
-                )
-            |> select
-                [ id "dyeing-medium"
-                , class "form-select form-select-sm w-75"
-                , onInput
-                    (DyeingMedium.fromString
-                        >> Result.withDefault inputs.product.dyeing.defaultMedium
-                        >> updateDyeingMedium
+dyeingMediumField { inputs, updateDyeingMedium, showAdvancedFields } =
+    showIf showAdvancedFields <|
```
ccomb, line 131: Would be better to move up the `showIf (...) <|` outside each view function and gather them in the calling view, to keep the upper layout at the same place

src/Page/Textile.elm
```diff
@@ -968,109 +971,147 @@ lifeCycleStepsView db { detailedStep, impact } simulator =
...
-                , div [ class "card-body py-2 row g-3 align-items-start flex-md-columns" ]
-                    [ div [ class "col-md-6" ]
-                        [ inputs.price
-                            |> Maybe.withDefault inputs.product.economics.price
-                            |> productPriceField
+            , div [ class "d-flex flex-column sticky-md-top" ]
+                [ CardTabs.view
+                    { tabs =
+                        [ { label =
+                                "Mode règlementaire"
```
ccomb, line 1097: Both forms are OK (réglementaire / règlementaire). Choose your side! :-D

ccomb (changes requested):
Be prepared to introduce a "Level 1" tab, "Level 2" tab, then "Level 3" tab! ;-)

src/Views/Textile/Step.elm
n1k0: done in https://github.com/MTES-MCT/ecobalyse/pull/607/commits/298cb78c55fb254cad72a2407211f1917fe4e5e5

ccomb (approved):
👍

PR #608, opened 2024-05-28
feat: store history of scores for example products

paulboosz:
Store history of textile and food examples score in db. Visualize evolution in Viz webapp

https://superset.ecobalyse.fr/login/

[User Story](https://www.notion.so/Historiser-et-permettre-de-suivre-l-volution-des-scores-de-exemples-de-produit-df286f347e0d46aeaac772ac48b453c5?pvs=4)

# Mechanic

- `tests/examplesScore.spec.js` compute the score of our examples each time the tests are run. Tests results are stored in `tests/textile-examples-score.json` and `tests/food-examples-score.json`
- `.github/workflows/score_history.yml` when triggered runs `data/common/score_history/score_history.py`
  Score history trigger :
  -  push to branch `staging` or `master`
  - when `score_history` workflow is manually triggered
  - when any commit from any branch contains the `[score_history]` tag
  
- `data/common/score_history/score_history.py` will get the new score from `tests/XXX-examples-score.json` compare it with the old score in the `score_history` postgresql table in staging db. If scores are different it will store the new score in the db.


vjousse:
We should land https://github.com/MTES-MCT/ecobalyse/pull/622 before and then, rebase this one on top of it.

vjousse (approved):
Tested locally and on the old `staging` branch. Works as expected. The Read Only user has been created on production. Shipping it.

PR #609, opened 2024-05-31
Token from env

ccomb:
Allow to specify a token through scalingo environment, to allow some admin users to have always the same token


paulboosz:
Work as expected :)

ccomb:
Thx for merging!

PR #610, opened 2024-06-03
Reorganize the textile simulator UI.

n1k0:
No user story but the basic idea is to move fields around in the UI for better general UX.


src/Page/Textile.elm
```diff
@@ -754,17 +754,17 @@ productCategoryField { products } query =
...
-            , class "col-sm-4 col-form-label text-truncate"
-            ]
+    div [ class "d-flex flex-column" ]
+        [ label [ for "selector-product", class "form-label text-truncate" ]
             [ text "Catégorie" ]
         , button
-            [ class "col-sm-8 flex-fill form-select ElementSelector text-start w-auto"
-            , id "selector-product"
+            [ id "selector-product"
+            , class "form-select ElementSelector text-start w-auto"
```
vjousse, line 767: Just for my understanding, this TitleCase (`ElementSelector`) looks strange from the outside, why not use the same style with `-` than the other classes?

vjousse (approved):
LGTM :+1: Just a little question, we don't show "Teinture sur" anymore in non advanced mode but we did before. Is it the expected behavior?

src/Page/Textile.elm
n1k0: It's the ongoing convention for custom elements in styles.scss

n1k0:
> we don't show "Teinture sur" anymore in non advanced mode but we did before. Is it the expected behavior?

Yes, [it's a requested bug fix](https://www.notion.so/Introduire-un-champ-autonome-cat-gorie-fc0e802f1fb149e7a1699b64343defcb?pvs=4#6066295c477a4ad3aebb39ac08dabb21)

n1k0:
I'll hold on landing this patch because Pascal has discovered [a limitation in the way we reset the fields](https://www.notion.so/Introduire-un-champ-autonome-cat-gorie-fc0e802f1fb149e7a1699b64343defcb?pvs=4#c0063204904842abbbfa2d8013e495c4) (I still have to understand what he means precisely haha)

n1k0:
@vjousse would you mind reviewing the latest changes? The `fabricProcess` is no more a required parameter of the textile query, so we can more easily fallback on the default product fabric process when switching product category. Sorry if that sounds complicated, because it is 😅

public/data/textile/examples.json
vjousse, line 1: If it's optional (and not removed) shouldn't we keep some examples with the "fabricProcess" parameter set?

openapi.yaml
vjousse, line 1: Is this file automatically generated?

tests/Server/RouteTest.elm
vjousse, line 1: I may miss something, but it looks like we don't test that removing the parameter works as expected in the generated output. Is it expected?

vjousse (approved):
Works as expected :+1: 
Just some nits regarding the test for my personal understanding.

public/data/textile/examples.json
n1k0: At some point we'll have to allow POs to edit example, which has been a long time requested feature. There was previously no custom fabric process used in examples as they all featured the default one associated with their category. Also we should not update examples in production before we announce a new version as discussed today.

openapi.yaml
n1k0: Nope, manually creafted and maintained with love since the very beginning. (Help).

tests/Server/RouteTest.elm
n1k0: Indeed I'll add that.

PR #612, opened 2024-06-03
docs: enhance README

vjousse:
# Proposed changes

- [x] As `env.sample` should be used for dev purposes, having debug enabled by default seems to make sense.
- [x] Add simple docker setup for PG in the README
- [x] Allow ALLOWED_HOSTS to be overridden by a config value
- [x] Listen only on `127.0.0.1` when starting the backend to be aligned with the default `ALLOWED_HOSTS` 
- [x] Add ALLOWED_HOSTS env variable to scalingo

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,7 @@
+ALLOWED_HOSTS=localhost,127.0.0.1
 BACKEND_ADMINS=foo@bar.baz,baz@bar.foo
 DEFAULT_FROM_EMAIL=bar@foo.baz
-DJANGO_DEBUG=False
+DJANGO_DEBUG=True
 DJANGO_SECRET_KEY=please-change-this
 DJANGO_BYPASS_AUTH=False
 EMAIL_HOST_PASSWORD=please-change-this
--- a/backend/backend/settings.py
+++ b/backend/backend/settings.py
@@ -36,13 +36,11 @@ SECRET_KEY = config("DJANGO_SECRET_KEY", "dev_not_so_secret_key")
 # SECURITY WARNING: don't bypass auth in production
 BYPASS_AUTH = config("DJANGO_BYPASS_AUTH", cast=bool, default=False)
 
-ALLOWED_HOSTS = [
-    HOSTNAME,
-    "localhost",
-    "127.0.0.1",
-    ".osc-fr1.scalingo.io",
-    "staging-ecobalyse.incubateur.net",
-]
+ALLOWED_HOSTS = config(
+    "ALLOWED_HOSTS",
+    f"{HOSTNAME},localhost,127.0.0.1",
+    cast=lambda v: [s.strip() for s in v.split(",")],
+)
 
 
 # Application definition
--- a/README.md
+++ b/README.md
@@ -8,18 +8,34 @@ L'application est accessible [à cette adresse](https://ecobalyse.beta.gouv.fr/)
 
 ## Socle technique et prérequis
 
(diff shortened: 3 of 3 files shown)


.env.sample
```diff
@@ -1,13 +1,13 @@
...
+DJANGO_DEBUG=True
 DJANGO_SECRET_KEY=please-change-this
 EMAIL_HOST_PASSWORD=please-change-this
 EMAIL_HOST_USER=please-change-this
 ENABLE_FOOD_SECTION=True
 MATOMO_HOST=stats.beta.gouv.fr
 MATOMO_SITE_ID=57
 MATOMO_TOKEN=xxx
-NODE_ENV=production
+NODE_ENV=dev
```
n1k0, line 11: I think it's `development` by convention

backend/backend/settings.py
```diff
@@ -33,13 +33,16 @@
...
+ALLOWED_HOSTS = config(
+    "ALLOWED_HOSTS",
+    [
+        HOSTNAME,
+        "localhost",
+        "127.0.0.1",
+        ".osc-fr1.scalingo.io",
+        "staging-ecobalyse.incubateur.net",
+    ],
+)
```
n1k0, line 45: So we should add an example of it in `.env.sample`, and probably set the defaults as just `HOSTNAME` and `127.0.0.1` so the settings module is agnostic to our specific infrastructure requirements and therefore makes the project easier to self-host.

n1k0 (commented):
I felt pretty good having a production configuration by default — mostly for safety reasons — but I'm not entirely opposed to switching to a dev one. I still think tweaking a local `.env` file to override some defaults is no big deal but I might be mistaken here. Your call.

I'll leave others commenting if they feel like so.

backend/backend/settings.py
vjousse: Agree. For the defaults I suppose it will require an update of the ENV variables on scalingo, that's why I left it untouched for now.

vjousse:
> I felt pretty good having a production configuration by default — mostly for safety reasons — but I'm not entirely opposed to switching to a dev one. I still think tweaking a local `.env` file to override some defaults is no big deal but I might be mistaken here. Your call.
> 
> I'll leave others commenting if they feel like so.

Usually `.env` files are not used in production, only for dev purposes. I don't know if it's better for the frontend to run in production mode by default, but for Django it doesn't make sense as, for example, statics are not collected and served so the admin can't be used properly.

ccomb (approved):
Thx for improving the readme!
No opinion on dev/prod default value.

backend/backend/settings.py
vjousse: Done here https://github.com/MTES-MCT/ecobalyse/pull/612/commits/8b3ba114acf359d13f97dc960b9d7b05c6414f1b

n1k0:
In theory we should have this landed on staging first, but I see staging is currently desynchronized from master. I don't know what to do, thoughts @ccomb @vjousse?

vjousse:
I've resync staging with master. Why would you merge it to staging before? For people to test? Testing on https://ecobalyse-pr612.osc-fr1.scalingo.io is not enough?

n1k0:
> I've resync staging with master. @n1k0 why would you merge it to staging before? For people to test? Testing on https://ecobalyse-pr612.osc-fr1.scalingo.io/ is not enough?

This is part of the questioning we should address during the dedicated meeting. I'm not sure what should be expected from staging now Pascal mentioned that we should have distinct stagings for methodology and features 😅

PR #614, opened 2024-06-05
Add missing `gettext` in CI build

vjousse:
## :wrench: Problem

Currently, there is a silent error in the CI due to the absence of `gettext`.

image

## :cake: Solution

Install `gettext` in the CI build.


n1k0 (approved):
Thanks 👍 it may be worth adding a note in the README about `gettext` being installed and available on the system.

PR #615, opened 2024-06-05
Fix backend tests

vjousse:
## :wrench: Problem

- Backend tests are not run in the CI
- `npm` command/scripts naming is inconsistent between the front and the back
- We display a stack track in the Django tests and don't check for the exact exception thrown

## :cake: Solution

Run the tests in the CI by adding an entry into the `node.js.yml` file and rename the inconsistent commands in `package.json`.

As the authenticate can fail for many reasons (expired token, user missing) we test that it's the appropriate exception that is raised in the python logs and silent the output at the same time.

- The CI should pass
- `npm test`, `start:dev` and `npm test:server-ci` should work as usual

### DIFF
--- a/backend/authentication/tests.py
+++ b/backend/authentication/tests.py
@@ -88,11 +88,14 @@ class DjangoAuthenticationTests(TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(json.loads(response.content).get("success"), True)
 
-        # wrong json login url
-        response = self.client.get(
-            "/accounts/login/1::1ru4fl:Z3XQ1tyORtolai5tycqK99BjUgzefc7o-mfui0DQFa0?next=/"
-        )
-        self.assertEqual(response.status_code, 403)
+        with self.assertLogs(logger="mailauth.backends", level="ERROR") as cm:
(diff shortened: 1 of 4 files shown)


backend/authentication/tests.py
```diff
@@ -88,11 +88,14 @@ def test_register_post(self):
...
-        # wrong json login url
-        response = self.client.get(
-            "/accounts/login/1::1ru4fl:Z3XQ1tyORtolai5tycqK99BjUgzefc7o-mfui0DQFa0?next=/"
-        )
-        self.assertEqual(response.status_code, 403)
+        with self.assertLogs(logger="mailauth.backends", level="ERROR") as cm:
+            # wrong json login url
+            response = self.client.get(
+                "/accounts/login/1::1ru4fl:Z3XQ1tyORtolai5tycqK99BjUgzefc7o-mfui0DQFa0?next=/"
+            )
```
n1k0, line 95: How about just

```python
response = self.client.get("/accounts/login/invalid-token?next=/")
```

n1k0 (approved):
Sweet 👌

PR #620, opened 2024-06-17
chore: draw the difference between Brightway and SimaPro

ccomb:
Just by doing `make compare_food`. A bunch of PNG files are written in the `food` folder


paulboosz (approved):
There is a bug on the reformatting at the end of computeAggregated.js. The bug doesn't impact the results.

Actually we don't need this script and can just modify export.py. [Card created here](https://www.notion.so/Supprimer-computeAggregated-js-dc602f0a7dc1468b99fff08e0394f35c?pvs=4)

PR #622, opened 2024-06-17
fix: remove fake details

vjousse:
## :wrench: Problem

Currently, we use a complex system of "fakeDetails" to serve files with some fake impacts to the test suite in order to obfuscate the computations. It doesn't make sense while encryption is not present.

## :cake: Solution

Remove the fakeDetails process and add an environment variable to bypass authentication process in the CI.

End 2 end output JSON files were updated with the "real" computation results.

The CI should still be green.


package.json
```diff
@@ -10,6 +10,7 @@
     "build:init": "./bin/update-version.sh && mkdir -p dist && cp -r public/* dist/ && npm run db:build",
     "processes:build": "elm make src/ComputeAggregated.elm --output=compute-aggregated-app.js && node compute-aggregated.js",
     "db:build": "./bin/build-db",
+    "db:build:full": "TEXTILE_PROCESSES=public/data/textile/processes_impacts.json FOOD_PROCESSES=public/data/food/processes_impacts.json ./bin/build-db",
```
n1k0, line 13: Wondering out loud, couldn't we get rid of the `TEXTILE_PROCESSES` and `FOOD_PROCESSES ` env vars entirely? And/or we could infer the file paths by checking the `DJANGO_BYPASS_AUTH` value (and if so maybe find a better name for it?)

server.js
```diff
@@ -115,9 +115,7 @@ const getProcesses = async (token) => {
   if (token) {
     headers["token"] = token;
   }
-  if (process.env.NODE_ENV == "test") {
-    headers["fakeDetails"] = "true";
-  }
```
n1k0, line 120: 🙏

src/ComputeAggregated.elm
```diff
@@ -72,46 +71,19 @@ toExport { definitionsString, textileProcessesString, foodProcessesString } =
...
-                            foodProcessesFakeDetails =
-                                fakeDetails foodProcesses
                         in
                         Encode.object
                             [ ( "textileProcesses", Encode.list TextileProcess.encode textileProcesses )
                             , ( "foodProcesses", Encode.list FoodProcess.encode foodProcesses )
                             , ( "textileProcessesOnlyAggregated", Encode.list TextileProcess.encode textileProcessesOnlyAggregated )
                             , ( "foodProcessesOnlyAggregated", Encode.list FoodProcess.encode foodProcessesOnlyAggregated )
-                            , ( "textileProcessesFakeDetails", Encode.list TextileProcess.encode textileProcessesFakeDetails )
-                            , ( "foodProcessesFakeDetails", Encode.list FoodProcess.encode foodProcessesFakeDetails )
```
n1k0, line 88: 🙏

src/ComputeAggregated.elm
```diff
@@ -72,46 +71,19 @@ toExport { definitionsString, textileProcessesString, foodProcessesString } =
...
-        Unit.impact 1.2
-
-
-fakeDetails : List { a | impacts : Impacts } -> List { a | impacts : Impacts }
-fakeDetails processes =
-    processes
-        |> List.map
-            (\process ->
-                { process | impacts = Impact.mapImpacts fakeImpact process.impacts }
-            )
```
n1k0, line 112: I think I'm gonna cry 🥲

tests/Data/Food/RecipeTest.elm
```diff
@@ -180,16 +180,16 @@ suite =
                                     |> Expect.within (Expect.Absolute 0.0001) (Unit.impactToFloat scoring.all)
                                     |> asTest "should expose coherent scoring"
                                 , Unit.impactToFloat scoring.biodiversity
-                                    |> Expect.within (Expect.Absolute 0.01) 26269.888595064378
+                                    |> Expect.within (Expect.Absolute 0.01) 191.70923964701296
```
n1k0, line 183: 🥲

tests/TestUtils.elm
```diff
@@ -21,19 +20,8 @@ asTest label =
...
-                -- in the DB.
-                fakeDetailedDb =
-                    { db
-                        | food = { food | processes = fakeDetails food.processes }
-                        , textile = { textile | processes = fakeDetails textile.processes }
-                    }
-            in
-            describe name (suite fakeDetailedDb)
+        Ok db ->
+            describe name (suite db)
```
n1k0, line 24: I know this is highly unprofessional, but : 🥲

n1k0 (commented):
Thanks for this work 🥲 I have a feeling we could go even further and get rid the fake-processes-related env vars completely, what do you think?

n1k0 (approved):
Ok LGTM and we'll probably remove the compute-aggregated script later at some point, but for now this is good to land.

PR #624, opened 2024-06-18
Fix impact distribution computations in the sidebar.

n1k0:
[User story](https://www.notion.so/Bug-sur-les-calculs-de-dans-la-vue-d-taill-e-70d1116d1b8d4e5cbf1798513d6accac)

### DIFF
--- a/src/Data/Textile/LifeCycle.elm
+++ b/src/Data/Textile/LifeCycle.elm
@@ -8,6 +8,7 @@ module Data.Textile.LifeCycle exposing
     , getNextEnabledStep
     , getStep
     , getStepProp
+    , getTotalImpactsWithoutComplements
     , init
     , sumComplementsImpacts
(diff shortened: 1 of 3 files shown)


vjousse (approved):
Nothing to say regarding the code :+1: 
Regarding the tests, I'm surprised that, changing a computation formula doesn't require any test update. Shouldn't we cover this case somewhere or maybe testing this behavior doesn't make sense?

n1k0:
> Regarding the tests, I'm surprised that, changing a computation formula doesn't require any test update. Shouldn't we cover this case somewhere or maybe testing this behavior doesn't make sense?

These functions are used in the UI and involve so many complex data fixtures that it usually makes tests hard to write, to read and provide not much confidence about them. But you're right, it may be worth it just for the sake of knowing when they break. I'll add some.

PR #627, opened 2024-06-20
feat: serve multiple app versions

vjousse:
## :wrench: Problem

We want to be able to easily go back in time and serve multiple versions of the app.

## :cake: Solution

Download the zipped versions available on the releases page on Github, extract the versions to the `./versions` directory on scalingo build. Read the `./versions` directory when starting the node server and load in memory the different versions (and their detailed or not detailed impacts files). Serve the static files when getting to `/versions/<version_number>` url.

For the moment, it support encrypted and non encrypted files. We may want to skip a version if we don't have the encrypted files but we do have the non encrypted ones…

The tokens are validated with the current database and not the one that existed at the time.

Version number is displayed at the bottom starting with version `v1.3.0`.

Some documentation is available here: https://www.notion.so/01-08-2024-Modification-des-releases-publi-es-pour-comptabilit-avec-le-syst-me-de-versioning-5ffafb4e20cc4a2a88c8619348366b0b

## :memo: Todos

- [x] Get the correct `version.json` file from Elm
- [x] Redirect to the correct version after login
- [x] Provide the created version number to the build script in release please workflow (because fetching it from github is impossible at this stage as the release has not been created)
- [x] Add script to build static version for a specific tag including the encrypted files and an up to date `version.json` file including the changes of this PR: https://github.com/MTES-MCT/ecobalyse/pull/686
- [x] Rebuild all the assets of the already published releases including the changes above and test their behavior locally
- [x] Check local storage behavior between versions
- [x] Upload new assets to Github
- [x] Deploy review app with all versions

https://ecobalyse-pr627.osc-fr1.scalingo.io/versions/v1.1.0/
https://ecobalyse-pr627.osc-fr1.scalingo.io/versions/v1.1.1/
https://ecobalyse-pr627.osc-fr1.scalingo.io/versions/v1.2.0/
https://ecobalyse-pr627.osc-fr1.scalingo.io/versions/v1.3.2/
https://ecobalyse-pr627.osc-fr1.scalingo.io/versions/v2.0.0/

Should serve the expected versions.


server.js
```diff
@@ -149,8 +157,96 @@ api.all(/(.*)/, bodyParser.json(), async (req, res) => {
...
 });
 
+version.use("/:versionNumber", (req, res, next) => {
+  const versionNumber = req.params.versionNumber;
+
+  // Construct the directory path based on the versionNumber path segment
+  const staticDir = path.join(__dirname, "versions", versionNumber);
+
+  // Verify that the file path is under the static directory for security reasons
+  filePath = fs.realpathSync(path.resolve(staticDir));
```
n1k0, line 167: You need to declare this using either `const` or `let`, the latter being for if you intend to mutate it. Same for other similarily introduced code blocks in this file.

server.js
```diff
@@ -149,8 +157,96 @@ api.all(/(.*)/, bodyParser.json(), async (req, res) => {
...
+  if (!filePath.startsWith(staticDir)) {
+    res.statusCode = 403;
+    res.end();
+    return;
+  } else {
+    const openApiContents = yaml.load(fs.readFileSync(path.join(staticDir, "openapi.yaml")));
+    res.status(200).send(openApiContents);
+  }
```
n1k0, line 196: Seems like this block of code is copied three times in the file, maybe it could be factored out

n1k0 (commented):
Super early feedback just trying to test the patch on my box: I couldn't make it work locally

bin/build-specific-app-version.sh
```diff
@@ -0,0 +1,106 @@
...
+
+# Clean dir if it already exists
+if [ -d $VERSION_DIR ]; then
+  rm -rf $VERSION_DIR
+  mkdir -p $VERSION_DIR
+fi
+
+mv $GIT_CLONE_DIR/dist/* $VERSION_DIR
+mv $GIT_CLONE_DIR/server-app.js $VERSION_DIR
+cp $GIT_CLONE_DIR/openapi.yaml $VERSION_DIR
```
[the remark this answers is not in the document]
vjousse: Could you try with the latest commit?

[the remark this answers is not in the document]
n1k0: It now works 👍

n1k0 (commented):
I think we should rather list the available versions before the server even starts and put each string in some in-memory list and query this list on each incoming requests. Much more efficient and risk-free.

server.js
n1k0: Actually, rather see my latest review comment

servers.conf.erb
```diff
@@ -34,6 +34,12 @@ server {
...
     proxy_set_header X-Real-IP $remote_addr ;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for ;
+	}
+	location /versions {
+		proxy_pass http://localhost:8001;
+    proxy_set_header Host $host ;
+    proxy_set_header X-Real-IP $remote_addr ;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for ;
 	}
 }
```
n1k0, line 45: Nit: Indentation looks messed up here

server.js
```diff
@@ -93,6 +105,55 @@ app.get("/accessibilite", (_, res) => res.redirect("/#/pages/accessibilité"));
 app.get("/mentions-legales", (_, res) => res.redirect("/#/pages/mentions-légales"));
 app.get("/stats", (_, res) => res.redirect("/#/stats"));
 
+// Versions
+const versionsDir = "./versions";
+let availableVersions = [];
+
+// Loading existing versions in memory
+if (fs.existsSync(versionsDir)) {
```
n1k0, line 113: Do we really need this check? I think we always want to have this directory available, so let it crash :)

server.js
```diff
@@ -93,6 +105,55 @@ app.get("/accessibilite", (_, res) => res.redirect("/#/pages/accessibilité"));
...
+  const dirs = fs.readdirSync(versionsDir);
+  for (const dir of dirs) {
+    const foodNoDetails = path.join(versionsDir, dir, "data/food/processes.json");
+    const textileNoDetails = path.join(versionsDir, dir, "data/textile/processes.json");
+
+    const foodDetailed = path.join(versionsDir, dir, "data/food/processes_impacts.json");
+    const textileDetailed = path.join(versionsDir, dir, "data/textile/processes_impacts.json");
+
+    const foodDetailedEnc = path.join(versionsDir, dir, "processes_impacts_food.json.enc");
+    const textileDetailedEnc = path.join(versionsDir, dir, "processes_impacts_textile.json.enc");
```
n1k0, line 123: that's a lot of repeated `path.join(versionsDir, dir`, maybe we should factor this out

server.js
```diff
@@ -93,6 +105,55 @@ app.get("/accessibilite", (_, res) => res.redirect("/#/pages/accessibilité"));
...
+    } else if (fs.existsSync(foodDetailed) || fs.existsSync(textileDetailed)) {
+      // Or use old files
+      processesImpacts = {
+        foodProcesses: fs.readFileSync(foodDetailed, "utf8"),
+        textileProcesses: fs.readFileSync(textileDetailed, "utf8"),
+      };
+    }
+
+    availableVersions.push({
+      dir: dir,
```
n1k0, line 147: Nit: just `dir,` is enough

n1k0 (commented):
The patch and features look promising, though unfortunately I won't have the time to fully review and test them before I'm on holiday, so I'll just leave a few comments, mostly nits

server.js
vjousse: Well, when running it locally we may not have this `versions` directory. And I love the idea that the app will still serve the current version even if no other versions are available, looks more fail safe to me.

bin/download-github-releases.sh
```diff
@@ -0,0 +1,50 @@
...
+
+ARCHIVE_FILE_NAME_TEMPLATE="TAG-dist.tar.gz"
+DOWNLOAD_URL_TEMPLATE="https://github.com/$ECOBALYSE_GIT_REPO/releases/download/TAG/$ARCHIVE_FILE_NAME_TEMPLATE"
+
+
+TAGS=$(curl -sSL \
+  https://api.github.com/repos/MTES-MCT/ecobalyse/releases \
+  | grep tag_name \
+  | sed 's/ //g; s/"//g; s/,//' \
+  | cut -d':' -f 2)
```
n1k0, line 21: This really really wants to be written in at least Python

server.js
vjousse: Should we consider that not having the `/versions` directory is a problem per se? I mean, when you clone the project and start the server you will not have the old versions locally, they are only available after running the `bin/download-github-releases.sh` script.

n1k0: Ok I now understand what you meant by running locally, where you don't necessarily care *at all* about versions.

n1k0 (approved):
LGTM, we can always keep migrating bash parts to Python in the future, but for now this unlocks other important work 👍

PR #629, opened 2024-06-20
Improve textile product category related behaviors.

n1k0:
[User story](https://www.notion.so/Requestionner-les-param-tres-attach-s-au-champ-cat-gorie-ac862d35339b4f70abde3901d7dd32f4#8ececf00a8d9413d8c5aef569fe79383)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,7 +1,0 @@
-#!/usr/bin/env sh
-
-# Check if code is formatted
-if ! npm run format:check ; then
-  echo "Code is not formatted. Run 'npm run format:json' to format before committing."
-  exit 1
-fi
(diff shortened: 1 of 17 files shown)


n1k0:
@vjousse the patch has been privately approved by Pascal

vjousse (approved):
Code wise, looks good to me. I can't judge the functional side as the User Story looks too complicated to me.
Regarding the `describe "Product update"` that was removed, shouldn't we add the opposite test now?

n1k0:
> Regarding the describe "Product update" that was removed, shouldn't we add the opposite test now?

Testing that the `mass` field doesn't exist anymore in the model? What for? 🤔

I'm landing this!

PR #631, opened 2024-06-24
Update master with staging

vjousse:
## :wrench: Problem

Maintaining a coherent `staging` is difficult and can slow down development and merge into `master` because we don't really know if the code in `staging` is ready to go into production or not. 

## :cake: Solution

Remove `staging` completely and rely on review apps to check the changes.


n1k0 (approved):
Disclaimer: I did **not** thoroughly review this 14802 lines diff line by line, though we should be good to go as this is basically the current state of the staging branch.

I'd really appreciate POs to spend some time validating that critical features and workflows still work as expected before we land the patch though.

n1k0:
I took some time to manually test a few core features like user auth and the api, everything seems to be working just fine. I can't say anything about the ongoing state of work in the food area though, but as it's not intended to be publicly available just yet, I think we're fine anyway. Ship it.

PR #638, opened 2024-06-25
fix: fix fading UI activation status bug.

n1k0:
[User story](https://www.notion.so/Bug-activation-d-sactivation-du-d-lavage-63d94a65203a4bc7be7ff8ab43e4c111)


vjousse (approved):
LGTM :+1:

PR #641, opened 2024-06-27
fix: yarnSize API param wasn't parsed when provided as float.

n1k0:
This patch partially address [this user story](https://www.notion.so/Bug-API-52a9084b9f404f1faf14b9d27a2339cf), most notably the bug regarding the `yarnSize` API param parsing. Yes at some point we'll want one single card per bug.

## 🐛 Understanding the bug

```elm
> import Json.Decode as D
> import Data.Unit as U
> "27.04"|>D.decodeString U.decodeYarnSize
Err (Failure ("Expecting an INT") <internals>)
    : Result D.Error U.YarnSize
> "27"|>D.decodeString U.decodeYarnSize
Ok (Quantity 27000) : Result D.Error U.YarnSize
```

Yarn size cannot be expressed as floats while they should.

## 💡 Solution

Accept a yarn size to be expressed as a float.

### DIFF
--- a/src/Data/Textile/Formula.elm
+++ b/src/Data/Textile/Formula.elm
@@ -465,7 +465,7 @@ computeThreadDensity surfaceMass yarnSize =
             1.08
     in
     toFloat (Unit.surfaceMassInGramsPerSquareMeters surfaceMass)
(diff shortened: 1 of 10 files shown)


vjousse (approved):
:+1: (Elm 😍 )

PR #642, opened 2024-06-27
fix: don't export legacy fake details

vjousse:
## :wrench: Problem

`npm run processes:build` fails as it tries to export legacy fake details files.

## :cake: Solution

Remove fake details export.

### DIFF
--- a/compute-aggregated.js
+++ b/compute-aggregated.js
@@ -27,16 +27,12 @@ elmApp.ports.export.subscribe(
     foodProcesses,
     textileProcessesOnlyAggregated,
     foodProcessesOnlyAggregated,
-    textileProcessesFakeDetails,
-    foodProcessesFakeDetails,
   }) => {
     try {
       exportJson("public/data/textile/processes_impacts.json", textileProcesses);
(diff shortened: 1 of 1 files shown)


.github/workflows/score_history.yml
```diff
@@ -69,6 +72,13 @@ jobs:
       - name: Install dependencies
         run: pip install numpy pandas requests python-dotenv GitPython sqlalchemy psycopg2
 
+      - name : Setup PostgreSQL tunnel to scalingo
+        env:
+          SCALINGO_REGION: ${{ secrets.SCALINGO_REGION }}
+          SCALINGO_APP: cdb-${{ matrix.service }}${{ secrets.APP_SUFFIX }}
+          SCALINGO_API_TOKEN: ${{ secrets.SCALINGO_API_TOKEN }}
+        run : echo "coucou"
+
```
n1k0, line 81: I don't think this should be part of this specific PR

n1k0 (approved):

.github/workflows/score_history.yml
vjousse: Oops ^^

PR #643, opened 2024-06-27
chore: connect to PG using tunnel

vjousse:
## :wrench: Problem

Currently we use a direct connection to the PostgreSQL exposed publicly. As we want to close the public access, we have to find another way to connect to PG.

## :cake: Solution

Use the scalingo CLI to create a tunnel between Github Actions and PG.

I am using a personal access token for the scalingo API access. We should change it as soon as we got the new `dev@ecobalyse.beta.gouv.fr` email and service account on Scalingo.

The `score_history` action should not fail.

### DIFF
--- a/.github/workflows/score_history.yml
+++ b/.github/workflows/score_history.yml
@@ -84,9 +84,12 @@ jobs:
           SCALINGO_API_TOKEN: ${{ secrets.SCALINGO_API_TOKEN }}
         run: |
           npm run server:build
+          scalingo db-tunnel DATABASE_URL &
           npm run server:start &
           npm run start:backend &
+
           for attempt in {1..20}; do sleep 1; if curl -s http://localhost:8001/ > /dev/null; then echo "-> Node server is up and ready on port 8001"; break; fi; echo "-> Waiting for the Node server to boot..."; done
           for attempt in {1..20}; do sleep 1; if curl -s http://localhost:8002/ > /dev/null; then echo "-> Python server is up and ready on port 8002"; break; fi; echo "-> Waiting for the Python server to boot..."; done
-          scalingo db-tunnel DATABASE_URL &
+          for attempt in {1..20}; do sleep 1; if lsof -i:10000 > /dev/null; then echo "-> PG Tunnel to Scalingo is up and ready on port 10000"; break; fi; echo "-> Waiting for the PG tunnel to Scalingo..."; done
+
           python data/common/score_history/score_history.py http://localhost:8001 $GITHUB_REF_NAME $LAST_COMMIT_HASH $SCALINGO_POSTGRESQL_SCORE_URL
--- a/.github/workflows/score_history.yml
+++ b/.github/workflows/score_history.yml
@@ -20,6 +20,9 @@ jobs:
 
       - uses: actions/checkout@v4
 
+      - name: Install Scalingo CLI
+        uses: ./.github/actions/scalingo-cli
+
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/score_history.yml
```diff
@@ -69,17 +72,37 @@ jobs:
...
         run: pip install numpy pandas requests python-dotenv GitPython sqlalchemy psycopg2
 
+
+      - name: Add scalingo to know hosts
+        run: echo "KNOWN_HOSTS=$(ssh-keyscan -H ssh.osc-fr1.scalingo.com)" >> $GITHUB_ENV
+
+      - name: Install SSH key
+        uses: shimataro/ssh-key-action@v2
+        with:
+          key: ${{ secrets.SCALINGO_SSH_KEY }}
```
n1k0, line 82: I'm cursious as to where does this `SCALINGO_SSH_KEY` value come from?

.github/workflows/score_history.yml
```diff
@@ -69,17 +72,37 @@ jobs:
...
         env:
           GITHUB_REF_NAME: ${{ github.ref_name }}
           LAST_COMMIT_HASH: ${{ github.sha }}
-          SCALINGO_POSTGRESQL_SCORE_URL: ${{ secrets.SCALINGO_POSTGRESQL_SCORE_URL }}
+          SCALINGO_POSTGRESQL_SCORE_URL: ${{ secrets.SCALINGO_POSTGRESQL_TUNNEL_SCORE_URL }}
           DJANGO_BYPASS_AUTH: True
           NODE_ENV: "test"
+          SCALINGO_REGION: ${{ secrets.SCALINGO_REGION }}
+          SCALINGO_APP: ecobalyse
+          SCALINGO_API_TOKEN: ${{ secrets.SCALINGO_API_TOKEN }}
```
n1k0, line 97: Same question

.github/workflows/score_history.yml
```diff
@@ -69,17 +72,37 @@ jobs:
...
         run: |
           npm run server:build
+          scalingo db-tunnel DATABASE_URL &
           npm run server:start &
           npm run start:backend &
+
           for attempt in {1..20}; do sleep 1; if curl -s http://localhost:8001/ > /dev/null; then echo "-> Node server is up and ready on port 8001"; break; fi; echo "-> Waiting for the Node server to boot..."; done
           for attempt in {1..20}; do sleep 1; if curl -s http://localhost:8002/ > /dev/null; then echo "-> Python server is up and ready on port 8002"; break; fi; echo "-> Waiting for the Python server to boot..."; done
+          for attempt in {1..20}; do sleep 1; if lsof -i:10000 > /dev/null; then echo "-> PG Tunnel to Scalingo is up and ready on port 10000"; break; fi; echo "-> Waiting for the PG tunnel to Scalingo..."; done
```
n1k0, line 106: how about sleeping 1 second **after** failing to connect?

n1k0 (approved):
I don't really know how I could efficiently review this patch so I'm just gonna ask dumb questions here and there 😅

paulboosz (approved):
score_history action didn't fail on this branch

.github/workflows/score_history.yml
vjousse: This is a private key generated with `ssh-keygen -t ed25519`. I've added the corresponding public key to my scalingo account. we need to do the same with a scalingo dedicated user. Discussion about it is currently ongoing here: https://mattermost.incubateur.net/betagouv/pl/mdpjyet487nqmdw8iw8gzdmrfc

.github/workflows/score_history.yml
vjousse: `SCALINGO_REGION` is an env variable needed by the scalingo cli tool https://doc.scalingo.com/platform/internals/regions. This value could have been public and don't really need a secret for it.
` SCALINGO_API_TOKEN` has been generated on scalingo using my user. We should  replace it with ah API key created by the "dedicated scalingo user" that we will create.

vjousse:
I've changed the scalingo token to the one of `ecobalyse-dev` and it works as expected :+1: 
https://github.com/MTES-MCT/ecobalyse/actions/runs/9746774745/job/26897925099

trigger-commit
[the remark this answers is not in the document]
vjousse: Oh thanks, didn't know about this one :sweat_smile:

PR #646, opened 2024-07-01
refactor: don't use Django anymore to serve the files

vjousse:
## :wrench: Problem

Using Django to serve the files based on a token will prevent us to fully version the app using Git and will make the file encryption harder to manage.

## :cake: Solution

We serve the files directly from node by reading them on the disk. It allow us to encrypt those files in the future.

This is the first step to fully encrypt the processes files. We will need to remove the detailed impacts from the repo and move them to a private repo instead (using git submodules) in another PR.

It should be an idempotent change. Be sure that the app is behaving as usual when logged in and not logged in. Do the same for the API.


.proxyrc.json
```diff
@@ -6,7 +6,7 @@
     "target": "http://127.0.0.1:8002/"
   },
   "/processes/": {
-    "target": "http://127.0.0.1:8002/"
+    "target": "http://127.0.0.1:8001/"
```
n1k0, line 9: Just to be sure I'm getting this right, `8001` is the expressjs port and `8002` the django one, and we now serve the processes endpoint from express instead of django?

server.js
```diff
@@ -8,14 +8,12 @@ const helmet = require("helmet");
 const Sentry = require("@sentry/node");
 const { Elm } = require("./server-app");
 const lib = require("./lib");
-const memoize = require("fast-memoize");
```
n1k0, line 11: You'll have to update `package.json` and `packag-lock.json` files accordingly.

n1k0 (changes requested):

.proxyrc.json
vjousse: That's it :+1:

server.js
```diff
@@ -110,28 +118,29 @@ api.get(/^\/products$/, (_, res) => res.redirect("textile/products"));
 const cleanRedirect = (url) => (url.startsWith("/") ? url : "");
 api.get(/^\/simulator(.*)$/, ({ url }, res) => res.redirect(`/api/textile${cleanRedirect(url)}`));
 
-const getProcesses = async (token) => {
-  let headers = {};
-  if (token) {
-    headers["token"] = token;
+const getProcesses = async (encrypted = true) => {
```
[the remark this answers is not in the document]
vjousse: Here, when reading the files for the API: https://github.com/MTES-MCT/ecobalyse/pull/646/files#diff-a4c65ede64197e1a112899a68bf994485b889c4b143198bac4af53425b38406fR143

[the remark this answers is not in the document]
n1k0: Thanks, I'm gonna get glasses soon

server.js
```diff
@@ -141,7 +150,7 @@ api.all(/(.*)/, bodyParser.json(), async (req, res) => {
     method: req.method,
     url: req.url,
     body: req.body,
-    processes: result.processes,
+    processes: processes,
```
n1k0, line 155: Nit: you can simply write `processes` here:

```js
{
    // …
    body: req.body,
    processes
    // …
}
```

server.js
```diff
@@ -110,28 +128,21 @@ api.get(/^\/products$/, (_, res) => res.redirect("textile/products"));
...
-  }
+const getProcesses = async (encrypted = true) => {
+  const { ENCRYPTION_KEY } = process.env;
 
-  const processesUrl = `http://127.0.0.1:${django_port}/processes/processes.json`;
-  const processesRes = await fetch(processesUrl, { headers: headers });
-  const processes = await processesRes.json();
-  return { processes: processes, status: processesRes.status };
+  if (encrypted) {
+    return encrypt(JSON.stringify(processes), ENCRYPTION_KEY);
```
n1k0, line 135: Even this could be computed before server starts, provided `ENCRYPTION_KEY` is never dynamically updated at runtime

server.js
```diff
@@ -92,19 +94,38 @@ const apiTracker = lib.setupTracker(openApiContents);
...
+  foodProcesses: JSON.parse(fs.readFileSync(foodImpactsFile, "utf8")),
+  textileProcesses: JSON.parse(fs.readFileSync(textileImpactsFile, "utf8")),
+};
 
 const processes = {
-  foodProcesses: fs.readFileSync(foodFile, "utf8"),
-  textileProcesses: fs.readFileSync(textileFile, "utf8"),
+  foodProcesses: JSON.parse(fs.readFileSync(foodFile, "utf8")),
+  textileProcesses: JSON.parse(fs.readFileSync(textileFile, "utf8")),
 };
```
n1k0, line 110: It really feels like it shouldn't be too hard to dynamically set all unwanted impacts to zero and get rid of the `processes.json` file.

n1k0 (approved):
I've been playing **a lot** with the refactored auth flow and checked detailed impacts so many time that I feel safe to approve this patch.

PR #652, opened 2024-07-08
chore: give uv a try

vjousse:


vjousse:
Archiving this one, it works as expected, but Scalingo doesn't support uv for now. Modifying the buildpack to support uv would be too cumbersome.

PR #654, opened 2024-07-09
test: use pytest instead of Django builtin

vjousse:
## :wrench: Problem

We are using the Django specific unittest and, even if it works as expected, it's not the most up to date method to test apps in the python world. Moreover, it's specific to Django and will not be usable for any other python code in the project.

## :cake: Solution

Use the python standard `pytest`. Some pros and cons are detailed here: https://djangostars.com/blog/django-pytest-testing/

Nothing special, tests should be green.


n1k0 (approved):
LGTM 👍

PR #658, opened 2024-07-10
refactor: encrypt processes

vjousse:


bin/build-db
```diff
@@ -11,11 +11,20 @@ console.log(`Building database for the ${process.env.NODE_ENV} env…`);
...
+    console.error("-> Exiting the build-db process.\n");
+    process.exit(1);
+  }
 } else {
-  foodProcesses = "public/data/food/processes.json";
-  textileProcesses = "public/data/textile/processes.json";
+  foodProcesses = process.env.FOOD_PROCESSES_IMPACTS_PATH || "public/data/food/processes.json";
+  textileProcesses =
+    process.env.TEXTILE_PROCESSES_IMPACTS_PATH || "public/data/textile/processes.json";
 }
```
n1k0, line 28: My understanding is that if we're not in a test environment, eg. we're in production, which defines `FOOD_PROCESSES_IMPACTS_PATH`, we'll build the db using the detailed impacts version instead of the non-detailed version — which is not what we want I think. Am I misunderstanting something?

ccomb (commented):
I've added a PR on top of this PR https://github.com/MTES-MCT/ecobalyse/pull/664/

.github/workflows/node.js.yml
```diff
@@ -68,6 +69,12 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y gettext
 
+    - name: Clone the detailed impacts
+      run: |
+        eval `ssh-agent -s`
+        ssh-add - <<< '${{ secrets.SCALINGO_SSH_KEY }}'
```
n1k0, line 75: I might be confused but where does this `secrets.SCALINGO_SSH_KEY` value come from, especially as we're in a github action and see it defined nowhere? In any case it might be worth adding a comment here and/or an entry in the README about it.

.github/workflows/release-please.yml
```diff
@@ -40,12 +40,25 @@ jobs:
         run: |
           npm run build
           npm run server:build
+          cp server-app.js dist/
+          cp openapi.yaml dist/
```
n1k0, line 44: This could totally be some `ci:build-app` npm command or whatever better name

.pre-commit-config.yaml
```diff
@@ -36,3 +36,10 @@ repos:
         types_or: [python]
         entry: |
           bash -c 'npm --prefix . run fix:ruff:format "${@}"' --
+
+      - id: dont-commit-detailed-processes
+        name: check that we don't commit processes_impacts.json
+        language: system
+        files: ^processes_impacts.json$
+        entry: |
+          bash -c 'exit 1'
```
n1k0, line 45: Nit: we may want to print some custom error message here, eg. "Prevented pushing detailed impacts data to public git repository. To update these, clone and push to the ecobalyse-private repository."

bin/build-db
```diff
@@ -11,8 +11,16 @@ console.log(`Building database for the ${process.env.NODE_ENV} env…`);
...
 if (process.env.NODE_ENV === "test") {
-  foodProcesses = "public/data/food/processes_impacts.json";
-  textileProcesses = "public/data/textile/processes_impacts.json";
+  foodProcesses = process.env.FOOD_PROCESSES_IMPACTS_PATH;
+  textileProcesses = process.env.TEXTILE_PROCESSES_IMPACTS_PATH;
+
+  if (!foodProcesses || !textileProcesses) {
+    console.error(
+      "\n🚨 ERROR: For the tests to work properly, you need to specify FOOD_PROCESSES_IMPACTS_PATH and TEXTILE_PROCESSES_IMPACTS_PATH env variables. They need to point to the detailed versions of the processes files. Please, edit your .env file accordingly.",
+    );
```
n1k0, line 20: We might benefit from template strings to format this long line over multiple ones for legibility here

compute-aggregated.js
```diff
@@ -24,11 +32,16 @@ elmApp.ports.export.subscribe(
...
+      exportJson(process.env.FOOD_PROCESSES_IMPACTS_PATH, foodProcesses);
       exportJson("public/data/textile/processes.json", textileProcessesOnlyAggregated);
       exportJson("public/data/food/processes.json", foodProcessesOnlyAggregated);
-      console.log("EXPORTED!");
+      console.log(
+        `\n4 files exported to\n- ${process.env.TEXTILE_PROCESSES_IMPACTS_PATH}\n- ${process.env.FOOD_PROCESSES_IMPACTS_PATH}\n- public/data/textile/processes.json\n- public/data/food/processes.json!`,
+      );
+      console.log(
+        "\n🚨 Be sure to commit the detailed impacts `processes_impacts.json` files in the `ecobalyse-private` repo.",
+      );
```
n1k0, line 44: Same remark as per using a single template strings with natural newlines for legibility and use a single `console.log` call.

n1k0 (commented):
This is a first batch of review comments as I still need to checkout the branch and test it locally.

.github/workflows/node.js.yml
vjousse: It's defined in the Github repo secrets but sure, I'll add a comment :+1:

vjousse: Bat as it's not a private key used only for Scalingo, I should rename it to clarify things

server.js
```diff
@@ -25,6 +25,19 @@ if (process.env.NODE_ENV !== "test" && (!MATOMO_HOST || !MATOMO_SITE_ID || !MATO
   process.exit(1);
 }
 
+if (process.env.NODE_ENV === "test") {
+  if (!process.env.ECOBALYSE_DATA_DIR) {
```
n1k0, line 29: This could be `if (process.env.NODE_ENV === "test" && !process.env.ECOBALYSE_DATA_DIR)`

Even better, if you destructure vars along the ones at the beginning of the file:

```js
// Env vars
const {
  ECOBALYSE_DATA_DIR,
  MATOMO_HOST, 
  MATOMO_SITE_ID, 
  MATOMO_TOKEN, 
  NODE_ENV,
  SENTRY_DSN, 
} = process.env;

// …

if (NODE_ENV === "test" && !ECOBALYSE_DATA_DIR) {
```

(also yeah I like we order things alphabetically everywhere we can ^^)

n1k0: Also, when `ECOBALYSE_DATA_DIR` is defined in env, we should test that the directory and detailed processes data files exist, and raise/exit early if they don't — meaning we don't ever care about that later on.

compute-aggregated.js
```diff
@@ -2,11 +2,26 @@ require("dotenv").config();
 const fs = require("fs");
 const { Elm } = require("./compute-aggregated-app");
 
+if (!process.env.ECOBALYSE_DATA_DIR) {
```
n1k0, line 5: See the remark made in server.js for the same check; it feels like some check code could be factored out.

.github/workflows/release-please.yml
```diff
@@ -38,14 +38,24 @@ jobs:
...
+          eval `ssh-agent -s`
+          ssh-add - <<< '${{ secrets.PRIVATE_SSH_KEY }}'
+          git clone git@github.com:MTES-MCT/ecobalyse-private.git
+
+      - name: Encrypt the impacts files
+        env:
+          ENCRYPTION_KEY: ${{ secrets.ENCRYPTION_KEY }}
+        run : |
+          npm run encrypt ./ecobalyse-private/data/textile/processes_impacts.json dist/processes_impacts_textile.json.enc
+          npm run encrypt ./ecobalyse-private/data/food/processes_impacts.json dist/processes_impacts_food.json.enc
```
n1k0, line 54: Nit: took me a while to realize these encoded files are bundled in the release so self-hosters can configure the app to use them if they provide the encryption key. Might worth adding a comment :)

lib/index.js
```diff
@@ -57,6 +58,38 @@ function setupTracker(spec) {
   };
 }
 
+function checkDataFiles() {
+  const { ECOBALYSE_DATA_DIR } = process.env;
```
n1k0, line 62: How about passing the env as an arg? It could make it easy to unit test, even if it's not mandatory for this specific function.

```js
function checkDataFiles(env = {}) {
  const { ECOBALYSE_DATA_DIR } = env;
```

lib/index.js
```diff
@@ -57,6 +58,38 @@ function setupTracker(spec) {
...
+  if (!ECOBALYSE_DATA_DIR) {
+    console.error(
+      `
+🚨 ERROR: You need to specify ECOBALYSE_DATA_DIR env variable.
+   It must point to a directory with the detailed versions of the processes files. Please, edit your .env file accordingly.",
+   -> Exiting the process.
+`,
+    );
+    process.exit(1);
+  }
```
n1k0, line 73: In the same spirit, how about throwing instead of printing and exiting here?

In the same spirit, how about throwing instead of printing and exiting here?
```js
if (!ECOBALYSE_DATA_DIR) {
  throw new Error("You need to specify…");
}
```

So later you can catch and exit:

```js
try {
  checkDataFiles(process.env);
} catch (err) {
  console.error(`🚨 ERROR: ${err.message}`);
  process.exit(1);
}
```

server.js
```diff
@@ -91,13 +102,26 @@ const openApiContents = yaml.load(fs.readFileSync("openapi.yaml"));
...
+// If detailed impacts files doesn't exist,
+// we should default to the non detailed files
+if (!fs.existsSync(textileImpactsFile)) {
+  textileImpactsFile = textileFile;
+}
+
+if (!fs.existsSync(foodImpactsFile)) {
+  foodImpactsFile = foodFile;
+}
+
```
n1k0, line 124: Note that this wouldn't be required anymore with the new suggested `checkDataFile` implementation.

lib/index.js
```diff
@@ -57,6 +58,24 @@ function setupTracker(spec) {
...
+  const processesImpactsTextilePath = "data/textile/processes_impacts.json";
+  const processesImpactsFoodPath = "data/food/processes_impacts.json";
+
+  for (const file of [processesImpactsTextilePath, processesImpactsFoodPath]) {
+    if (!fs.existsSync(`${ecobalyseDataDir}/${file}`)) {
+      throw new Error(`No '${file}' file found in your ECOBALYSE_DATA_DIR '${ecobalyseDataDir}' directory.
+Be sure that the file structure inside '${ecobalyseDataDir}' is correct.`);
+    }
+  }
+}
```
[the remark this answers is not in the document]
vjousse: Well, sometimes, like in `compute-aggregated` we want to throw if `ecobalyseDataDir` is not set. But I may provide an additional flag to explicitly throw in this case :+1:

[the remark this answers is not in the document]
n1k0: Yeah, there are distinct use cases for these checks and they should be handled explicitly 👍

lib/index.js
```diff
@@ -57,6 +58,40 @@ function setupTracker(spec) {
   };
 }
 
+function getDataFiles(ecobalyseDataDir, checkDataDirIsSet = false) {
+  if (!ecobalyseDataDir && checkDataDirIsSet) {
+    throw new Error(`You need to specify ECOBALYSE_DATA_DIR env variable.
+It must point to a directory with the detailed versions of the processes files. Please, edit your .env file accordingly.`);
+  }
```
n1k0, line 65: Nit: "Well, actually" 😅 I think this shouldn't be part of that function with the added bloat, but rather checked in context before calling `checkDataFiles`.

I understand this is reaching nitpicking territory so I'll leave it up to you. Code works and that's probably all what matters 👍

n1k0 (approved):
Deployed on the review app with the ECOBALYSE_DATA_DIR env properly set, the app works as expected from my testing.

PR #661, opened 2024-07-11
feat(ui): allow searching ingredients and materials by process name

n1k0:
[Follow-up to this card](https://www.notion.so/Augmenter-la-visibilit-du-choix-ingr-dient-ICV-source-360b90ee35d24afdaca58cbe9194eb90)

image


n1k0:
Vincent C doesn't want this.

PR #664, opened 2024-07-11
refactor: Use only one env var to specify the private repo

ccomb:
The export in python needs to access the private repo inside a docker container, so we need a bind mount to this repo. It's simpler to use a single environment variable to this repo, then point to different files in this repo. I've propagated this one-env-var behaviour to the JS files as well


n1k0 (commented):
I'm pretty sure @vjousse should be reviewing this one. I'm not against using a single env var for simplicity, I'm just slightly cautious with the name, though I don't currently have a much better proposition.

ccomb:
obsolete

PR #666, opened 2024-07-14
fix(api): Fix invalid openapi format

n1k0:
We've been serving an invalid openapi spec for long it seems. This patch fixes it and adds OpenAPI validation to CI.


vjousse (approved):
Oh boy, we need to find something to generate the OpenApi Spec automatically :scream: Otherwise LGTM :+1:

n1k0:
> Oh boy, we need to find something to generate the OpenApi Spec automatically 😱

That would imply moving away from Elm for tha API I'm afraid :(

PR #670, opened 2024-07-16
fix(api): Fix material shares sum rounding precision error.

n1k0:
[user story](https://www.notion.so/BUG-API-Contr-le-de-la-somme-des-mati-res-1af0370afefa43ae97438295b771b5b1)

### DIFF
--- a/src/Data/Textile/Query.elm
+++ b/src/Data/Textile/Query.elm
@@ -19,6 +19,7 @@ module Data.Textile.Query exposing
     , updateMaterialSpinning
     , updateProduct
     , updateStepCountry
+    , validateMaterials
     )
 
 import Base64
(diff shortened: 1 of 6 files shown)


src/Data/Textile/Query.elm
```diff
@@ -361,6 +362,29 @@ updateStepCountry label code query =
...
+
+    else
+        let
+            total =
+                materials
+                    |> List.map (.share >> Split.toFloat)
+                    |> List.sum
+        in
+        -- Note: taking care of float number rounding precision errors https://en.wikipedia.org/wiki/Round-off_error
+        if not (List.member total [ 1, 0.6 + 0.3 + 0.1 ]) then
```
vjousse, line 378: Instead of having to deal with floats and this kind of JS silliness, why is the `Split` type not using `Int` by default?

```elm
type Split
    = Split Int
```
To my understanding it's modeling shares that can be expressed as an `Int` percentage. So here we should check if the sum is equal to `100`.

It looks like it would solve this problem and converting from an `Int` to a `Float` when really needed should be easier than dealing with rounding stuff.

vjousse (changes requested):

src/Data/Textile/Query.elm
n1k0: The internal repr can be either int or float, hence float. Also, reverting #525 would reintroduce related bugs. I think it's safe to go the route provided with this patch.

vjousse: Thanks for the precision!

vjousse (approved):

PR #672, opened 2024-07-17
chore: provide a way to synchronise a branch of `ecobalyse-private` with a PR

vjousse:
## :wrench: Problem

We need a way to be able to test a PR with a specific branch of the `ecobalyse-private` repo.

## :cake: Solution

Parse the body of the PR for a line with `ecobalyse_data: <branch_name>` pattern. When deploying on scalingo, check if the current deployed commit is part of an opened PR and check if there is a `ecobalyse_data: <branch_name>` line. If so, we checkout the corresponding branch of `ecobalyse-private`.

We do the same when running the tests and score_history in the Github Workflows.

Create a branch in the `ecobalyse-private`, create a PR (or use this one) with the specific line. Check that the app contains the data of the `ecobalyse-private` branch.

## :memo: Data

ecobalyse_data: organic_tomato

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -75,6 +75,13 @@ jobs:
         # The corresponding public key must be set as a deploy key of the private repo
         ssh-add - <<< '${{ secrets.PRIVATE_SSH_KEY }}'
         git clone git@github.com:MTES-MCT/ecobalyse-private.git
+        DATA_BRANCH_NAME=$(./bin/extract-data-branch-from-pr.sh $GITHUB_SHA)
+        if [ $? -eq 0 ]; then
+          cd ecobalyse-private
+          echo "-> Checkout branch $DATA_BRANCH_NAME of ecobalyse-private"
+          git checkout $DATA_BRANCH_NAME && git pull
+          cd ..
+        fi
 
     - name: Build app
       run: npm run build --if-present
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -69,18 +69,24 @@ jobs:
         sudo apt-get install -y gettext
 
     - name: Clone the detailed impacts
+      env:
+        GITHUB_SHA: ${{ github.sha }}
       run: |
         eval `ssh-agent -s`
         # Private ssh key used by this Github action to clone the private repo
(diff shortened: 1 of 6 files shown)


.github/workflows/node.js.yml
```diff
@@ -76,6 +76,17 @@ jobs:
...
+    - name: Checkout specific ecobalyse-private branch
+      if: ${{ github.event_name == 'pull_request' }}
+      env:
+        LAST_COMMIT_SHA: ${{ github.event.pull_request.head.sha }}
+      run: |
+        eval `ssh-agent -s`
+        # Private ssh key used by this Github action to clone the private repo
+        # The corresponding public key must be set as a deploy key of the private repo
+        ssh-add - <<< '${{ secrets.PRIVATE_SSH_KEY }}'
+        ./bin/checkout-ecobalyse-private-branch.sh $LAST_COMMIT_SHA
```
ccomb, line 88: I understand that this is done during the build?
I created a PR with a merge of this one and the organic_tomato one (#665) to test the behaviour. I initially forgot to specify the `ecobalyse_data: <commit>` in the PR, and corrected while it was running. The test is failing by not finding a new process id which should be in the private repo. So we might need a way to detect and restart the tests/checks after modifying the description of the PR, a bit like the semantic PR title check. But would it be relevant?
```
on:
  pull_request_target:
    types:
      - opened
      - edited
      - synchronize
```

vjousse: Now it's computed on every test run. It was a bug in branch name detection, it should be fixed now.

buildpack-run.sh
```diff
@@ -1 +1,6 @@
+#!/usr/bin/env bash
+
 git clone git@github.com:MTES-MCT/ecobalyse-private.git
+
+# Scalingo should set the $SOURCE_VERSION variable
+./bin/checkout-ecobalyse-private-branch.sh $SOURCE_VERSION
```
n1k0, line 6: What calls this bash script? I couldn't find any docs about it the Scalingo one

I really wish we could use anything else than bash and sed for advanced scripting and parsing here :/

ccomb (approved):
Tested at least on one case that worked :+1:

vjousse:
Merging it now so we can synchronize branches. Refactoring using Python will come later on

PR #673, opened 2024-07-18
feat(textile,api,ui,data)!: update durability index computation

n1k0:
[user story](https://www.notion.so/Durabilit-Suppression-de-crit-res-mati-res-et-dur-e-de-commercialisation-371f0f41e03f446683a7f8c6ab952552)


ccomb (approved):
LGTM except the new checkbox in the story : `service de réparation ou de garantie`→`service de réparation`

package.json
```diff
@@ -39,7 +39,7 @@
     "test:backend": "pipenv run pytest",
     "test:client": "NODE_ENV=test npm run db:build && npx elm-test",
     "test:review": "npx elm-review",
-    "test:server": "npm run server:build && jest",
+    "test:server": "npm run server:build && NODE_ENV=test jest",
```
vjousse, line 42: Why not setting the env var for the whole command? `NODE_ENV=test npm run server:build && jest`

n1k0:
@ccomb as I commented on the card and [on the mattermost thread](https://mattermost.incubateur.net/fabnum-mte/pl/nanmghj7zpddxfby1sfzmn151w), I think the checkbox thing should be another card as it doesn't have any connection with the tasks at hand.

vjousse (approved):
The US is not enough for me to understand the reasoning behind the new computations, so I can only approve the code but not the business logic. LGTM :+1:

package.json
n1k0: because `jest` wouldn't see it, that's the way env var definition scopes work

n1k0:
sorry I was confused reading the diff on mobile, scratch my comment

PR #677, opened 2024-07-22
feat: display current version in the footer

vjousse:
## :wrench: Problem

We want to be able to know the current tagged version of the app on the website. It will allow people doing computations to know the version number they are using (and later on, come back to this version when the versioning system will be online).

## :cake: Solution

When deploying on Scalingo, call the Github API to get the tag corresponding to the current commit, if any. If we find a tag, we display it (and link to the release page) if not, the behavior is the same as before (display the commit number and a link to this commit on Github).

We are using calls to the public Github API to get the tag corresponding to the commit number. There is a rate limit of 60 calls per hour to the public Github API. We may want to get an app token to avoid such problems.

If you have one, you can set the `GITHUB_ACCESS_TOKEN` env variable and it will be used to make the Github calls and bypass this limitation.

It's checking the existence of the `SOURCE_VERSION` env variable usually set by Scalingo to call the Github API. If there is none set it assumes you are working locally and uses git commands to get the same information instead of calling the Github API.

Locally, start the app with `npm run start:dev`. If the current commit of the app doesn't have any tag associated, the commit number should be displayed at the bottom of the app.

Set the `SOURCE_VERSION` env variable (like Scalingo does) to a commit corresponding to a tag. For example the commit for the `v1.2.0` tag: 

    export SOURCE_VERSION=4ea8af325a81a73f5429c1525fce449c08c2447a

The run `npm run start:dev` in the same terminal. You should now see a link to the `v1.2.0` tag at the bottom of the app instead of a link to a commit.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -37,6 +37,9 @@ jobs:
 
       - name: Build app
         if: ${{ steps.release.outputs.release_created }}
+        env:
+          # Specify the created SHA to correctly update version.json
+          SOURCE_VERSION: ${{ steps.release.outputs.sha }}
         run: |
           npm run build:standalone-app
 
--- a/Pipfile
+++ b/Pipfile
@@ -16,5 +16,6 @@ GitPython = "3.1.43"
 pre-commit = "*"
 numpy = ">=1,<2"
 pytest-django = "*"
+pygithub = "*"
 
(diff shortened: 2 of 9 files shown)


ccomb (approved):
:+1:  Nice path towards a version selector
Just need to find a solution for dev env with the orange bottom banner

PR #685, opened 2024-07-23
refactor: use python instead of bash to get data branch

vjousse:
## :wrench: Problem

We want to rely as less as possible on bash for business logic.

## :cake: Solution

Replace the bash script getting the data branch name by a python one.

Refactor of https://github.com/MTES-MCT/ecobalyse/pull/672

Create a branch in the `ecobalyse-private`, create a PR (or use this one) with the specific line `ecobalyse_data: <ecobalyse_private_branch_name>`. Check that the app contains the data of the `ecobalyse-private` branch.

## :books: Data
ecobalyse_data: test1


lib/ecobalyse_github.py
```diff
@@ -0,0 +1,36 @@
...
+    else:
+        g = Github(retry=None)
+
+    try:
+        return (g.get_repo("MTES-MCT/ecobalyse"), g)
+    except github.RateLimitExceededException:
+        print(
+            "-> ⚠️ Github rate limit exceeded, can't get the tag for the current commit. You should setup GITHUB_ACCESS_TOKEN env variable to a valid Github access token.",
+            file=sys.stderr,
+        )
```
n1k0, line 36: Pondering if we shouldn't re-raise here, so the caller could try/except and retrieve this error message and do whatever it wants with it 🤔

Also that would eventually make the function slightly easier to mock/test if we ever want to

pyproject.toml
```diff
@@ -1,2 +1,5 @@
 [project]
 requires-python = ">= 3.11"
+
+[tool.ruff]
+extend-select = ["I"]
```
n1k0, line 5: Out of curiosity, what does this do?

bin/update-version.sh
```diff
@@ -11,7 +12,7 @@ DATA_DIR_HASH=null
 # Using the github API
 if [[ ! -z "${SOURCE_VERSION}" ]]; then
   HASH=$SOURCE_VERSION
-  TAG_FOUND=$(python $SCRIPT_DIR/get-tag-for-commit.py $HASH)
+  TAG_FOUND=$(python $SCRIPT_DIR/get_tag_for_commit.py $HASH)
```
n1k0, line 15: Do you think this file will eventually move to Python?

n1k0 (approved):
LGTM with a few nits and remarks 👍

pyproject.toml
vjousse: It adds automatic import ordering

bin/update-version.sh
vjousse: Yes why not, what do you think? We should maybe enforce the use of python instead of bash everywhere we start to have some logic that should be tested.

n1k0: Totally!

.github/workflows/node.js.yml
```diff
@@ -6,6 +6,7 @@ name: Node.js CI
 env:
   NODE_ENV: test
   ECOBALYSE_DATA_DIR: ./ecobalyse-private
+  PYTHONPATH: .
```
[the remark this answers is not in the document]
vjousse: Indeed, Python needs to know where the libs are located and by default, the CI doesn't add the current directory to the Python path (and I don't want to make our own lib just for that)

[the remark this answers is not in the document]
vjousse: Fixed by using a local python package, it's much cleaner now!

lib/ecobalyse_github.py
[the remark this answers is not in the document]
vjousse: See other comment, using a python package is much cleaner.

PR #686, opened 2024-07-23
chore: keep track of data-dir hash

vjousse:
## :wrench: Problem

As our versions depend on a private repository, we may need later on to be sure what commit hash of this private repository was used to build the specific version. For example, it will be useful for the version history system.

## :cake: Solution

Just add an entry in the `public/version.json` file containing the hash of the private repository.

We don't use it/decode it on the elm side as it's not useful for the end user.

Once deployed go to https://ecobalyse-pr686.osc-fr1.scalingo.io/version.json and check that `dataDirHash` key is present and that the sha is the same than the latest commit on https://github.com/MTES-MCT/ecobalyse-private

ecobalyse_data: test1


ccomb (approved):
:+1:

PR #687, opened 2024-07-25
docs: update readme ecobalyse private

paulboosz:
Added more detail on deployment mecanism between ecobalyse and ecobalyse_private


vjousse (approved):
`docs` would be more accurate I think for the PR name :+1:

PR #697, opened 2024-08-13
fix: add missing run command for score history [score_history]

vjousse:
## :wrench: Problem

The CI build fails on master because the `run` parameter is missing in the `score_history` workflow.

## :cake: Solution

We add the missing parameter.

A bug in the python code was fixed at the same time when a PR is opened without a body.

Run the `score_history` workflow dispatch by hand in the Github Actions.


n1k0 (approved):
👍

PR #700, opened 2024-08-14
feat: add a version selector widget.

n1k0:
[User story](https://www.notion.so/Comment-mat-rialiser-le-s-lecteur-de-version-dans-l-UI-d27ff90c71ab4f959163fe0441e72d68)

### DIFF
--- a/src/Views/Page.elm
+++ b/src/Views/Page.elm
@@ -318,18 +318,6 @@ pageHeader { session, activePage, openMobileNavigation, switchVersion } =
                 ]
                 [ img [ class "HeaderLogo", alt "République Française", src "img/republique-francaise.svg" ] []
                 , h1 [ class "HeaderTitle" ] [ text "Ecobalyse" ]
-                ]
-            , div [ class "d-none d-sm-flex align-items-center flex-nowrap gap-3" ]
-                [ a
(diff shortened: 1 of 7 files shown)


src/Views/Page.elm
```diff
@@ -293,35 +295,49 @@ versionLink version =
...
                 , h1 [ class "HeaderTitle" ] [ text "Ecobalyse" ]
                 ]
+            , session.releases
+                |> RemoteData.map
+                    (List.map
+                        (\release ->
+                            option [ selected <| Version.is release session.currentVersion ]
+                                [ text release.tag ]
+                        )
+                    )
```
vjousse, line 330: What if the current version is not found on github? I know it can't happen in production but it's a little bit strange when developing locally we have the `v2.0.0` selected by default. In this case, maybe we could add a "Unreleased" entry or something along the lines?

n1k0: @vjousse addressed in https://github.com/MTES-MCT/ecobalyse/pull/700/commits/8673fab287d9858391f105b897721dc7a6e6db9b

Do you want me to patch the injected js as well?

vjousse: I don't think that it makes sense as we are only patching released versions.

vjousse (approved):

PR #705, opened 2024-08-22
feat: allow downloading explorer data as CSV.

n1k0:
## :wrench: Problem

[user story](https://www.notion.so/Permettre-de-t-l-charger-en-csv-excel-les-donn-es-de-l-explorateur-a459d3be30cc44cb8977788073768df0)

Users want to be able to download the explorer data as CSV, so they can inspect and use them in their favorite software like Excel or Numbers.

## :cake: Solution

This patch adds a button to download explorer data as CSV on every tab, at the bottom right of the data table, for food and textile data.

Click on the download buttons and open the CSV file in your preffered CSV viewer.

image

image

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -1,8 +1,6 @@
 {
     "type": "application",
-    "source-directories": [
-        "src"
-    ],
+    "source-directories": ["src"],
     "elm-version": "0.19.1",
     "dependencies": {
         "direct": {
--- a/elm.json
+++ b/elm.json
@@ -1,6 +1,8 @@
 {
     "type": "application",
-    "source-directories": ["src"],
+    "source-directories": [
+        "src"
+    ],
     "elm-version": "0.19.1",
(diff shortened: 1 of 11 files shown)


vjousse:
It looks like `"` may not be encoded as expected: `"Pull viscose (550g) - Chine - Mode ""fast fashion""","Pull","2970.400526195585","540.0728229446518","N/A"`
(I'm downloading the examples tab)

n1k0:
@vjousse this is the way this specific character is escaped. 

It opens fine in Numbers:

image

It also opens just fine in Excel:

image

It even works on default mobile viewer (Safari):

image

What is the exact problem you're having with the generated file?

vjousse:
If it works it's ok for me, I was just doing a simple `cat` on the generated file.

vjousse (approved):

PR #706, opened 2024-08-26
fix: broken comparator charts on desynced cache data.

n1k0:
## :wrench: Problem

Sometimes when processes data are updated server side, the locally cached client data are desynced, which triggers weird issues — most notably in the comparator with missing chart segments, those corresponding to missing data or linked objects.

## :cake: Solution

This patch removes all use of the processes data cached in localStorage, and loads detailed data on client app start when authenticated. That way, we're always sure to start the client app with the latest available data from the server.

Very hard to simulate data changed from the server while a locally cached previous version of it lives in memory. I suspect you'll have to trust me on this one.

### DIFF
--- a/src/Data/Food/Db.elm
+++ b/src/Data/Food/Db.elm
@@ -1,8 +1,6 @@
 module Data.Food.Db exposing
     ( Db
     , buildFromJson
-    , updateIngredientsFromNewProcesses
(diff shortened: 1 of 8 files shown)


vjousse (approved):
A lot cleaner! Good job :+1:

PR #710, opened 2024-08-27
feat: implement upcycling.

n1k0:
## :wrench: Problem

We need to allow computing impacts for upcycled garments. Specs are [here](https://www.notion.so/Permettre-l-upcycling-en-niveau-1-e04de9780076444bac13a68f96c982f6).

## :cake: Solution

image

The patch adds a checkbox which:

- when checked:
  * disables the material, spinning, fabric and ennobling steps
  * sets the making complexity to "High"
- when unchecked:
  * enables all previously disabled steps
  * don't update the making complexity previously set

Toy around with the new checkbox.


src/Data/Textile/Simulator.elm
```diff
@@ -111,7 +108,8 @@ compute db query =
             else
                 identity
     in
-    init db query
+    Query.handleUpcycling query
+        |> init db
```
vjousse, line 112: Just a nit but having the `handleUpcycling` function called here looks strange to me (because everything else is not related to the query). Why not calling the `handleUpcycling` function in the `init` function directly?

src/Page/Api.elm
```diff
@@ -55,7 +55,16 @@ getApiServerUrl { clientUrl } =
 
 changelog : List News
 changelog =
-    [ { date = "18 juillet 2024"
+    [ { date = "28 août 2024"
+      , level = "minor"
+      , domains = [ "Textile" ]
+      , md =
+            """Ajout du champ booléen `upcycled` qui permet de préciser si un vêtement est remanufacturé.
+               En pareil cas, les étape Matières, Filature, Tissage/Tricotage et Ennoblissement sont
```
vjousse, line 63:
```suggestion
               En pareil cas, les étapes Matières, Filature, Tissage/Tricotage et Ennoblissement sont
```

src/Data/Textile/Simulator.elm
n1k0: addressed in https://github.com/MTES-MCT/ecobalyse/pull/710/commits/d81f54b8268b9de23b1260bf01f85bc3bb087734

n1k0:
> I have some strange behaviors (checkboxes are not displayed as checked but can be unchecked, checked boxes can be unchecked but they are still displayed as checked) when playing with the checkboxes when the upcycled is checked, see the screencast below.

This has been addressed in https://github.com/MTES-MCT/ecobalyse/pull/710/commits/547c5f19d0dd59956b4c6a40b499bb753b03598a

Would you mind reviewing the latest changes?

vjousse (approved):
With the latest changes, everything works as expected! :+1:

PR #712, opened 2024-08-28
fix: use fabric processes to compute fabric waste

vjousse:
## :wrench: Problem

Fabric waste is not updated when the fabric process is changed.

[Notion card](https://www.notion.so/BUG-Non-prise-en-compte-changement-de-masse-transformation-993e741e606c409c838d4f443e0a90bd)

## :cake: Solution

Use the right fabric process to compute waste instead of the product default fabric process.
This is mainly a revert of this revert https://github.com/MTES-MCT/ecobalyse/pull/588 ;)

Select "pull en coton 550g majorant par défaut" then "mode exploratoire" and in "Transformation - Tissage / Tricotage" select the process "Tissage".
The mass output of the previous step ("Transformation - Filature") should be "Masse : 0,863 kg" and not "Masse : 0,855 kg" as it is currently.


src/Data/Textile/Simulator.elm
```diff
@@ -583,16 +583,14 @@ computeFabricStepWaste { textile } ({ inputs, lifeCycle } as simulator) =
             lifeCycle
                 |> LifeCycle.getStepProp Label.Making .inputMass Quantity.zero
                 |> Formula.genericWaste
-                    -- FIXME: we should rather use inputs.fabricProcess here, but we must
-                    --        avoid updating results in production until next milestone.
-                    --        @see https://github.com/MTES-MCT/ecobalyse/pull/577
-                    (inputs.product.fabric
+                    (inputs.fabricProcess
+                        |> Maybe.withDefault inputs.product.fabric
```
n1k0, line 587: Nice catch! 👏

n1k0 (commented):
Nice work! I would expect at least a test to fail, but none do. Maybe we could add one?

tests/server.spec.js
```diff
@@ -336,6 +336,35 @@ describe("API", () => {
...
+        expect(fabricLifeCycle.waste).toEqual(0.045271493212669676);
+
+        const jeanQueryKnittingMix = jeanQuery.map((input) =>
+          input == "fabricProcess=weaving" ? "fabricProcess=knitting-mix" : input,
+        );
+
+        response = await makeRequest("/api/textile/simulator/detailed", jeanQueryKnittingMix);
+        expectStatus(response, 200);
+        fabricLifeCycle = response.body.lifeCycle.find((l) => l.label == "Tissage & Tricotage");
+        expect(fabricLifeCycle.waste).toEqual(0.039095022624434386);
```
n1k0, line 364: Maybe what we really want is simply checking that the waste is *less* than with the previous process used? So we won't have to maintain the exact numbers here

n1k0 (approved):
🚀

PR #714, opened 2024-08-31
chore: upgrade dependencies to their latest stable versions.

n1k0:


vjousse (approved):
Tested locally, works as expected, good work!

PR #719, opened 2024-09-03
fix: don't add disabled step impacts to lifecycle totals.

n1k0:
[Spec](https://www.notion.so/Bug-tshirt-coton-remanufactur-dans-le-comparateur-c4a1e484e8ea475788c5bf9678e7b540)


paulboosz (approved):
Looks good :)

PR #724, opened 2024-09-05
chore: upgrade browserlist.

n1k0:
This is actually for trying to fix this scalingo deployment issue:

image


n1k0:
Missed the `malloc` error line. Will try to scale the VM ram up instead.

PR #729, opened 2024-09-05
fix: ensure express app is properly monitored by Sentry.

n1k0:
We have `[server] [Sentry] express is not instrumented. This is likely because you required/imported express before calling Sentry.init()` log errors which is fixed by this patch — which follows the [official integration guide](https://docs.sentry.io/platforms/javascript/guides/express/).

### DIFF
--- a/lib/instrument.js
+++ b/lib/instrument.js
@@ -14,9 +14,8 @@ if (SENTRY_DSN) {
 
 function monitorExpressApp(app) {
   if (SENTRY_DSN) {
-    Sentry.setupExp;
+    Sentry.setupExpressErrorHandler(app);
   }
-  Sentry.setupExpressErrorHandler(app);
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


lib/instrument.js
```diff
@@ -0,0 +1,24 @@
...
+    integrations: [nodeProfilingIntegration()],
+    tracesSampleRate: 1.0,
+    profilesSampleRate: 1.0,
+  });
+}
+
+function monitorExpressApp(app) {
+  if (SENTRY_DSN) {
+    Sentry.setupExp;
+  }
```
vjousse, line 18: It looks strange, what is it exactly?

vjousse (approved):
Looks good but why don't you put everything in the `monitorExpressApp` function?

n1k0:
> Looks good but why don't you put everything in the monitorExpressApp function?

Because the Sentry module **must** be required before the express one, as stated in the doc linked i the description :)

PR #733, opened 2024-09-09
fix(ci): check for ecobalyse-private when extracting the branch name

vjousse:
## :wrench: Problem

Currently, we are checking for `ecobalyse_data` in the content of a PR to synchronize withe the `ecobalyse-private` repo. For consistency, check for `ecobalyse-private`.

Github token was changed to my personal token to avoid 403 responses. We may need to use an app token at some point.
Tests have been moved to the package directory for consistency.

Merge this PR in a PR requiring a different `ecobalyse-private` branch and check that everything is working as expected.


paulboosz (approved):
Tested it in another branch and it worked. Thanks :)

PR #734, opened 2024-09-10
chore: optimize scalingo build

vjousse:
## :wrench: Problem

Builds are some time taking ages on Scalingo and the image size is more than 1GB.

## :cake: Solution

Remove unneeded files from the image using [Scalingo `.slugignore`](https://doc.scalingo.com/platform/app/slugignore), cache elm files, move python packages not needed for production in `dev-packages` and remove .map files from old versions.

The env variable `ELM_HOME` has been set to `.elm` in scalingo so that it can be properly cached. Moving python packages into `dev-packages` requires to install them with `pipenv install -d` locally.
When doing `npm install` we should force the `NODE_ENV` to `development` so that `devDependencies` are always installed (by default, it is set to `production` in the `.env` file)

Scalingo should be green and image size smaller.


n1k0:
Where can we see the images size?

ccomb (approved):
1GB -> 751MB  nice job :+1:

PR #736, opened 2024-09-11
refactor: sort most record properties and constructors.

n1k0:
## :wrench: Problem

The codebase is sometimes painful to read on long pattern matching cases, long records with many properties, especially as they're not always ordered alphabetically to easily visually scan and find what you're after.

## :cake: Solution

Sorting long list of things alphabetically is well known for helping seeking information in such long lists. Let's use [fysiweb/elm-review-sorted](https://package.elm-lang.org/packages/fysiweb/elm-review-sorted/latest/) to report CS violations when important files don't comply.

Note: a few modules where excluded from applying the sorted record properties rule because we're using them as constructors, so position of properties used as arguments is required at given positions in the record definition, and were so left intact.

There's nothing to test really, as this is plain and simple code refactoring.


review/src/ReviewConfig.elm
```diff
@@ -44,6 +45,14 @@ config =
         |> Rule.ignoreErrorsForDirectories [ "tests/" ]
     , NoRedundantConcat.rule
     , NoRedundantCons.rule
+    , NoUnsortedRecordFields.rule
+        |> Rule.ignoreErrorsForDirectories [ "tests/" ]
+        |> Rule.ignoreErrorsForDirectories [ "src/Page" ]
+        |> Rule.ignoreErrorsForFiles [ "src/Data/Impact/Definition.elm" ]
+        |> Rule.ignoreErrorsForFiles [ "src/Views/Page.elm" ]
```
n1k0, line 52: Note to reviewers: these are excluded because they feature record types used as constructors where the order of fields is important and so we can't reorder them.

vjousse (approved):
Awesome work :heart_eyes: Tested locally, works as expected :+1:

PR #745, opened 2024-09-12
fix: fixed brightway explorer notebook error (wrong key)

ccomb:


n1k0 (approved):

n1k0:
@ccomb could you just please updating the commit label for something a bit more specific?

PR #752, opened 2024-09-16
chore: filter sentry errors by env

vjousse:
## :wrench: Problem

Production and review app errors are mix and matched in Sentry.

[Notion card](https://www.notion.so/Ne-loguer-les-v-nements-Sentry-qu-en-production-43f188ca8a624f08b6b7c7fd8e31254d)

## :cake: Solution

Use a different environment in Sentry for production and review-apps errors.

I've introduced a new file `scalingo.json` used to configure review apps https://doc.scalingo.com/platform/app/review-apps#run-a-task-after-each-deployment-of-a-review-app

I have CORS issue with the envelope call on the review-app without knowing why. Could be a server config/version issue cf. https://github.com/getsentry/sentry/issues/24637

Clicking on the footer should throw an error (the code will be removed before merging).

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@ if (process.env.SENTRY_DSN) {
       // Most often due to DOM-aggressive browser extensions
       /_VirtualDom_applyPatch/,
     ],
-    environment: process.env.IS_REVIEW_APP ? "review-app" : NODE_ENV || "development",
+    environment: process.env.IS_REVIEW_APP ? "review-app" : process.env.NODE_ENV || "development",
   });
 }
 
@@ -67,6 +67,15 @@ app.ports.appStarted.subscribe(() => {
   _paq.push(["disableCookies"]);
   _paq.push(["setSiteId", process.env.MATOMO_SITE_ID]);
   loadScript(u + "matomo.js");
+
+  var footer = document.getElementsByClassName("Footer");
+  footer[0].addEventListener(
+    "click",
+    function (e) {
+      throw new Error("Sentry Test");
+    },
+    false,
+  );
 });
 
 app.ports.loadRapidoc.subscribe((rapidocScriptUrl) => {
--- a/index.js
+++ b/index.js
@@ -11,11 +11,14 @@ if (process.env.SENTRY_DSN) {
     allowUrls: [
       /^https:\/\/ecobalyse\.beta\.gouv\.fr/,
       /^https:\/\/staging-ecobalyse\.incubateur\.net/,
+      // Review apps
+      /^https:\/\/ecobalyse-pr.*\.osc-fr1\.scalingo\.io/,
     ],
     ignoreErrors: [
       // Most often due to DOM-aggressive browser extensions
       /_VirtualDom_applyPatch/,
     ],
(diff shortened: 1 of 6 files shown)


vjousse:
At least it works for sentry errors from node
image

Can't test for the errors in the browser as there are CORS errors with the sentry hosted by beta.

scalingo.json
```diff
@@ -0,0 +1,7 @@
+{
+  "env": {
+    "IS_REVIEW_APP": {
+      "value": "true"
+    }
+  }
```
n1k0, line 6: Wouldnt that mean the value is always at `true` even on non PR/branch-based deploys?

index.js
```diff
@@ -11,11 +11,14 @@ if (process.env.SENTRY_DSN) {
...
       /^https:\/\/ecobalyse\.beta\.gouv\.fr/,
       /^https:\/\/staging-ecobalyse\.incubateur\.net/,
+      // Review apps
+      /^https:\/\/ecobalyse-pr.*\.osc-fr1\.scalingo\.io/,
     ],
     ignoreErrors: [
       // Most often due to DOM-aggressive browser extensions
       /_VirtualDom_applyPatch/,
     ],
+    environment: process.env.IS_REVIEW_APP ? "review-app" : process.env.NODE_ENV || "development",
```
n1k0, line 21: Ideally we should never collect events on `development` environment, don't you think? How could we proceed to ensure we're only collecting them on `production`, and maybe eventually on `staging`?

n1k0 (commented):
A few remarks and questions

index.js
```diff
@@ -6,16 +6,17 @@ import Charts from "./lib/charts";
 if (process.env.SENTRY_DSN) {
   Sentry.init({
     dsn: process.env.SENTRY_DSN,
-    integrations: [Sentry.browserTracingIntegration()],
-    tracesSampleRate: 0,
     allowUrls: [
       /^https:\/\/ecobalyse\.beta\.gouv\.fr/,
       /^https:\/\/staging-ecobalyse\.incubateur\.net/,
+      // Review apps
+      /^https:\/\/ecobalyse-pr.*\.osc-fr1\.scalingo\.io/,
```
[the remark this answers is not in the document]
vjousse: Why not if there are using the appropriate tag so we can filter production and non production errors on sentry? (this is what this PR does)
It may help debugging?

scalingo.json
vjousse: I thought so but… no. This file is only processed on review apps.

> The manifest is an optional file named scalingo.json you can add at the root of your repository to configure [review apps](https://doc.scalingo.com/platform/app/review-apps) and [one-click deploy button](https://doc.scalingo.com/platform/deployment/one-click-deploy). Its intent is to configure how an application should be created.

https://doc.scalingo.com/platform/app/app-manifest

index.js
vjousse: As localhost is not in the allowedUrls we should not collect dev errors. But… `localhost` seems to be enabled on `sentry.incubateur.net`
image

For the server side sentry, we could init it only if `NODE_ENV` is `production`, what do you think?

n1k0: Sounds good to me!

n1k0 (approved):
Looks great!

PR #753, opened 2024-09-16
fix: check db integrity after building it

n1k0:
## :wrench: Problem

We didn't check if the static Elm database built was valid, allowing deploying to production a broken app.

## :cake: Solution

Introduce a script to check the build database integrity.

Tweak the static JSON data sources and run `npm run db:build && db:check`:

image


.github/workflows/node.js.yml
```diff
@@ -91,9 +91,6 @@ jobs:
     - name: Build app
       run: npm run build --if-present
 
-    - name: Build Elm static Db
-      run: npm run db:build
```
n1k0, line 95: Note to reviewers: this is not necessary as this is called by the `npm run build` call above.

package.json
```diff
@@ -16,10 +16,10 @@
     "build": "npm run server:build && rimraf dist && npm run build:init && parcel build index.html --public-url ./",
     "build:init": "./bin/update-version.sh && mkdir -p dist && cp -r public/* dist/ && npm run db:build",
     "build:standalone-app": "npm run build && npm run server:build && cp server-app.js dist/ && cp openapi.yaml dist/",
-    "processes:build": "elm make src/ComputeAggregated.elm --output=compute-aggregated-app.js && node compute-aggregated.js",
```
n1k0, line 19: Note to reviewers: this has been moved and reordered alphabetically.

vjousse (approved):
Tested on scalingo with an `ecobalyse-private` branch with a missing process, works as expected: the build is failing.

PR #755, opened 2024-09-17
chore: render food api docs conditionally from env.

n1k0:
## :wrench: Problem

While we introduced the `ENABLE_FOOD_SECTION` env var to hide or show the Food section, associated menus and links, we hide the API docs for Food completely and there's no current way to access these API docs. 

## :cake: Solution

This patch checks for the value of the `ENABLE_FOOD_SECTION` var and filter the openapi docs appropriately.

Check that the API docs for food are available for this review app.


vjousse (approved):

PR #757, opened 2024-09-17
feat: airTransportRatio should depend on durability

vjousse:
[Notion card](https://www.notion.so/Adapter-la-part-de-transport-en-avion-automatiquement-en-fonction-de-la-durabilit-76a7514bedb44d3298f401d69f09672f)


src/Data/Textile/Simulator.elm
```diff
@@ -210,6 +214,37 @@ computeDurability ({ inputs } as simulator) =
...
 
+computeMakingAirTransportRatio : Simulator -> Simulator
+computeMakingAirTransportRatio ({ inputs } as simulator) =
+    simulator
+        |> updateLifeCycleStep Label.Making
+            (\({ country } as step) ->
+                { step
+                    | airTransportRatio =
+                        let
+                            iSEuropeOrTurquey =
```
n1k0, line 225: Nit: isEuropeOrTurquey (lower s)

Also, it's `Turkey` or `Türkiye`

src/Data/Textile/Simulator.elm
```diff
@@ -210,6 +214,37 @@ computeDurability ({ inputs } as simulator) =
...
+                                country.zone == Zone.Europe || country.code == Country.codeFromString "TR"
```
n1k0, line 226: Also, how about moving this to `Data.Country` as a helper?

src/Data/Textile/Simulator.elm
```diff
@@ -210,6 +214,37 @@ computeDurability ({ inputs } as simulator) =
...
+                                step.airTransportRatio
+
+                            -- If Making country is Europe or Turquey, airTransportRatio should always be 0
+                            ( _, True, _ ) ->
+                                Split.zero
+
+                            -- Else it should depend on the computed durability
+                            ( _, _, durabilty ) ->
+                                if durabilty >= 1 then
+                                    Split.fromFloat 0.33 |> Result.withDefault step.airTransportRatio
```
n1k0, line 240: We may want to introduce `Split.third` here

src/Data/Textile/Simulator.elm
```diff
@@ -210,6 +214,37 @@ computeDurability ({ inputs } as simulator) =
...
+
+                                else
+                                    Split.full
```
n1k0, line 243: Nit: Probably matter of taste but I'd find this block more readable using plain old if statements:

```suggestion
                        if inputs.airTransportRatio /= Nothing then
                            -- User-provided value always takes precedence
                            step.airTransportRatio

                        else if country.zone == Zone.Europe || country.code == Country.codeFromString "TR" then
                            -- If Making country is Europe or Turquey, airTransportRatio is always 0
                            Split.zero

                        else if Unit.floatDurabilityFromHolistic simulator.durability >= 1 then
                            -- Durable garments outside of Europe and Turquey
                            Split.fromFloat 0.33
                                |> Result.withDefault step.airTransportRatio

                        else
                            Split.full
```

n1k0 (approved):
The code looks good to me and the implementation seems to work as expected by the spec.

src/Data/Country.elm
```diff
@@ -152,6 +153,11 @@ getAquaticPollutionRatio scenario =
             Split.fromPercent 65 |> Result.withDefault Split.full
 
 
+iSEuropeOrTurkey : Country -> Bool
+iSEuropeOrTurkey country =
+    country.zone == Zone.Europe || country.code == codeFromString "TR"
```
n1k0, line 158: I swear I added a comment about s/iSEuropeOrTurkey/isEuropeOrTurkey here 🤷

n1k0 (commented):
I realize we could add a test ensuring that the simulator behaves accordingly to provided inputs and defaults regarding the air transport ratio

n1k0 (changes requested):
Hold on landing this patch, after a more careful review and toying around with the code, I realize we always override the default country airTransportRatio value when it's been used in `updateFromInputs` in `Step.elm` ([here](https://github.com/MTES-MCT/ecobalyse/blob/master/src/Data/Textile/Step.elm#L351-L352)). Ideally we should handle the airTransportRatio in a single place when we have the final durability as you have done in the patch, and so remove the assignment in `Step.elm`. I've asked what we should we do with default coutry airTransportRatio to POs.

vjousse:
I'm holding tight :smile:

n1k0:
I've pushed 90948a1d560c1eeeaec4c60ea34656ed3a7e18a0 which fallback on country air transport default when no condition is match. Let's wait for POs to know what they think.

vjousse:
> I realize we could add a test ensuring that the simulator behaves accordingly to provided inputs and defaults regarding the air transport ratio

Done in https://github.com/MTES-MCT/ecobalyse/pull/757/commits/f5e155b9d844ea3e1e148010f3b4f54c49013dd3

n1k0:
@vjousse can we remove the `airTransportRatio` data from the `countries.json` database and from the `Country` type?

vjousse:
> @vjousse can we remove the `airTransportRatio` data from the `countries.json` database and from the `Country` type?

Done in https://github.com/MTES-MCT/ecobalyse/pull/757/commits/5b0c9fd5b59978c4abfae87480ef068f11ff3cf4

vjousse:
Hmmmm should we completely remove the information from the country tab in the explorer?

n1k0:
> Hmmmm should we completely remove the information from the country tab in the explorer?

Not, we still want to expose all the supported country codes.

n1k0 (approved):
I took the liberty to fix the remaining conflicts and compilator issues. This is ready to land imho.

vjousse:
Approved by Pascal on [mattermost](https://mattermost.incubateur.net/fabnum-mte/pl/y7im6no4b7g33g49dyp3syd35a)

PR #758, opened 2024-09-18
chore: fix npm error when make_export_food

paulboosz:
Bug : 
image

Fix :
- The previous  chown command in the Dockerfile is setting ownership to user ID 1000, but the error message suggests the container is running as user ID 1001. -> use the actual jovyan user ID in the container
- clear the npm cache before running any npm commands


ccomb (approved):
👍

PR #760, opened 2024-09-18
fix: stricter validation of POST json body passed to the textile API.

n1k0:
~~Funnily enough, while we were thinking there wasn't any validation applied to some input POST parameters in the API, it was in fact our unit test env which was unable to execute these tests, because it didn't have access to the native JSON body to execute the correct assertions. Validation was working in e2e tests just fine… I should probably have spotted that earlier.~~

Edit: There actually wasn't any validation performed on some optional parameters passed as a JSON  body to POST requests, because we were using the infamous `Json.Decode.maybe` decoder which fallbacks to returning `Nothing` instead of a failure in case the decoded value is invalid. This patch leverages [Json.Decode.Extra.optionalField](https://package.elm-lang.org/packages/elm-community/json-extra/latest/Json-Decode-Extra#optionalField) instead:

> If a field is missing, succeed with Nothing. If it is present, decode it as normal and wrap successes in a Just.
>
> When decoding with [maybe](http://package.elm-lang.org/packages/elm-lang/core/latest/Json-Decode#maybe), if a field is present but malformed, you get a success and Nothing. optionalField gives you a failed decoding in that case, so you know you received malformed data.

### DIFF
--- a/src/Data/Textile/Query.elm
+++ b/src/Data/Textile/Query.elm
@@ -110,33 +110,37 @@ buildApiQuery clientUrl query =
 
 decode : Decoder Query
 decode =
-    -- Note: We're using Json.Decode.Extra's optionalField here because we want
-    -- a failure when an optional decoded field value is invalid
+    let
+        -- Note: Using Json.Decode.Extra's optionalField here because we want
+        -- a failure when a Maybe decoded field value is invalid
+        strictOptional field decoder =
+            DE.andMap (DE.optionalField field decoder)
+    in
     Decode.succeed Query
-        |> DE.andMap (DE.optionalField "airTransportRatio" Split.decodeFloat)
-        |> DE.andMap (DE.optionalField "business" Economics.decodeBusiness)
-        |> DE.andMap (DE.optionalField "countryDyeing" Country.decodeCode)
-        |> DE.andMap (DE.optionalField "countryFabric" Country.decodeCode)
-        |> DE.andMap (DE.optionalField "countryMaking" Country.decodeCode)
-        |> DE.andMap (DE.optionalField "countrySpinning" Country.decodeCode)
+        |> strictOptional "airTransportRatio" Split.decodeFloat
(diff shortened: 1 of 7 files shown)


n1k0:
@vjousse I'd like a first preliminary look over this patch and know what you think :)

src/Data/Textile/Query.elm
```diff
@@ -110,39 +111,46 @@ buildApiQuery clientUrl query =
...
         |> Pipe.optional "upcycled" Decode.bool False
-        |> Pipe.optional "yarnSize" (Decode.maybe Unit.decodeYarnSize) Nothing
+        |> strictOptional "yarnSize" Unit.decodeYarnSize
+
+
+strictOptional : String -> Decoder a -> Decoder (Maybe a -> b) -> Decoder b
+strictOptional field decoder =
+    -- Note: Using Json.Decode.Extra's optionalField here because we want
+    -- a failure when a Maybe decoded field value is invalid
+    DE.andMap (DE.optionalField field decoder)
```
vjousse, line 144: Nice catch! :heart:

n1k0: This has bitten so many legs and I'm still bitten. Amazing.

tests/server.spec.js
```diff
@@ -619,7 +614,7 @@ async function expectListResponseContains(path, object) {
 
 function expectStatus(response, expectedCode, type = "application/json") {
   if (response.status === 400 && expectedCode != 400) {
-    expect(response.body).toHaveProperty("errors", "");
+    expect(response.body).not.toHaveProperty("errors", "");
```
vjousse, line 617: Why are we checking that we don't have the `errors` field in the response here?

n1k0: Oops that's a leftover from some debugging session, nice catch!

vjousse (approved):
Nice work and good catch for the Maybe decoding! Tested locally and it works as expected.

PR #766, opened 2024-09-23
fix: check JSON dbs primary keys uniqueness.

n1k0:
## :wrench: Problem

We have unnoticed duplicate PK in `public/data/textile/examples.json`.

## :cake: Solution

We check for JSON dbs interity at build time.

This PR should fail with an appropriate error message. Once reviewed, the culprit dupliucate should be updated so the PR pass and can be merged.

### DIFF
--- a/bin/build-db
+++ b/bin/build-db
@@ -22,35 +22,60 @@ try {
   process.exit(1);
 }
 
-function getJson(path) {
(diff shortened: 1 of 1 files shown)


n1k0:
The CI build failed as expected:

image

PR #767, opened 2024-09-24
chore: add PR template

vjousse:
## :wrench: Problem

PRs descriptions have different formats depending on the author of the PR and don't always provide context and/or how to test.

## :cake: Solution

Setup a PR template so that every PR created starts with the same structure.

We can't test it before merging it into `master`. Once merged into `master`, the template should be the default content of a new PR description.


n1k0 (approved):
Awesome.

PR #768, opened 2024-09-24
chore: improve changelog by using `git-cliff`

vjousse:
## :wrench: Problem

We can't customize the changelog generated by [release-please](https://github.com/googleapis/release-please), so it's impossible to filter technical and non technical changes.

Moreover only `fix` and `feat` commits are parts of the changelog, sometimes it would make sense to add `chore`, `refactor` and `doc` changes to the changelog.

## :cake: Solution

Use [`git-cliff`](https://git-cliff.org/) to generate the changelog and create the tags too. The process is the same than with `release-please`, it will create an automated PR that, when merged, will tag the appropriate version and create a release.

Commits that are only relative to `data/` are excluded from the changelog.

We will be able to fully test this PR only once it will be merged into master and once a release will be created. I did all the tests that I could without merging, see below.

The generated PR will look like this one: https://github.com/MTES-MCT/ecobalyse/pull/797
> [!IMPORTANT]
> This automated PR was updated using the commits of this PR, so once this is merged into `master`, the changelog will only contain the squashed commits of the  `master` branch.

The different workflows were tested extensively here: https://github.com/vjousse/git-cliff-release-creation-exemple

I did a test on this PR by enabling release creation on commit push here (instead of merging the changelog PR). The tag and the release (configured as draft in my test) was created as expected, see the screenshots below.

Screenshot 2024-10-09 at 15-53-53 Tags · MTES-MCT_ecobalyse
Screenshot 2024-10-09 at 15-53-32 Releases · MTES-MCT_ecobalyse

I have removed the tag and the draft PR since.


.github/workflows/create_pr_for_changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+          args: --verbose --exclude-path "data/" --bump
+        env:
+          OUTPUT: CHANGELOG.md
+
+      - name: Generate latest changes
+        uses: orhun/git-cliff-action@v4
+        id: git_cliff_latest
+        with:
+          config: cliff.toml
+          args: --verbose --latest --strip header --exclude-path "data/" --bump --unreleased
```
n1k0, line 35: Not entirely sure about excluding `data/` entirely as we may still want to list data updates in our changelog; what do @ccomb and @paulboosz think?

.github/workflows/create_pr_for_changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Create Pull Request
+        if: steps.git_cliff.outputs.version != 'null'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          add-paths: |
+            CHANGELOG.md
+          base: master
+          commit-message: "chore: update CHANGELOG.md for ${{ steps.git_cliff.outputs.version }}"
+          signoff: false
```
n1k0, line 45: Would setting this to `true` fail? What would be required to have a signed-off commit here?

cliff.toml
```diff
@@ -0,0 +1,89 @@
...
+  # Replace issue numbers
+  { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "([#${2}](<REPO>/issues/${2}))"},
+  # Check spelling of the commit with https://github.com/crate-ci/typos
+  # If the spelling is incorrect, it will be automatically fixed.
+  #{ pattern = '.*', replace_command = 'typos --write-changes -' },
+]
+# regex for parsing and grouping commits
+commit_parsers = [
+  { message = "^feat", group = "<!-- 0 -->🚀 Features" },
+  { message = "^fix", group = "<!-- 1 -->🐛 Bug Fixes" },
```
n1k0, line 68: nit: May I suggest 🪲 as it's what's already used in Notion, for consistency?

n1k0 (approved):
LGTM, final validation will occur when this is landed onto master I guess

.github/workflows/create_pr_for_changelog.yml
paulboosz: I'd like to include changes in data but, as they are somehow different, what about including them in a dedicated data section (like features, bug fixes) ?

ccomb: The data generation pipeline and experiments is of no interest to end users. Actually I'm in favor of moving `data/` into the ecobalyse-data repo.

.github/workflows/create_pr_for_changelog.yml
vjousse: Good point. I'll add commit signing too.

.github/workflows/create_pr_for_changelog.yml
vjousse: This is something we could change latter on. I'm in favor of merging this PR as it is to be able to use it for the next release.

PR #769, opened 2024-09-24
fix(api): handle ingredient plane transport in food POST api.

n1k0:
## :wrench: Problem

The food API wasn't properly handling the `byPlane` ingredient parameter in `POST`, see [Notion card](https://www.notion.so/Inconsistance-UI-API-sur-mango-non-ue-9da4a84239b741efafd8c6620622c466).

## :cake: Solution

The patch fixes the issue.

There's a lot of tech debt in the way we parse the food query, but this would be HUGE work to properly refactor and should probably be assessed and prioritized better at a more appropriate time, so… this patch basically just patches and fixes the issue in the *least* dirty way.

The test suite still passes, which is good. To test the patch works, ask for the Non-EU Mango to the patched food API (I'm running it againt the local branch and server):

```
curl -sS -X POST "http://localhost:1234/api/food" \
 -H "accept: application/json"\
 -H "content-type: application/json"\
 -d '{"ingredients":[{"id":"mango-non-eu","mass":1000}]}' | jq .results.total.ecs
 1090.6833471290029
```

### DIFF
--- a/src/Data/Food/Ingredient.elm
+++ b/src/Data/Food/Ingredient.elm
@@ -69,6 +69,9 @@ byPlaneAllowed planeTransport ingredient =
         ( ByPlane, PlaneNotApplicable ) ->
             Err "Impossible de spécifier un acheminement par avion pour cet ingrédient, son origine par défaut ne le permet pas."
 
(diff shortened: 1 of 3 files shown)


vjousse (approved):
Tested locally, LGTM :+1:

PR #771, opened 2024-09-26
chore: don't download draft releases

vjousse:
## :wrench: Problem

When deploying on scalingo we download all releases without checking if it's a draft one, it may cause problems if we want to use the draft functionality of releases to use `git-cliff` for example https://github.com/MTES-MCT/ecobalyse/pull/768.

## :cake: Solution

Filter out releases that are draft.

The deploy to scalingo of this PR should download 11 releases and not more (we currently have 11 releases and some drafts).

### DIFF
--- a/bin/download_github_releases.py
+++ b/bin/download_github_releases.py
@@ -68,6 +68,11 @@ if __name__ == "__main__":
     nb_releases_extracted = 0
     for release in releases:
         if release.tag_name is None:
+            logger.info("Skipping release without a tag.")
+            continue
+
+        if release.draft:
+            logger.info("Skipping draft release.")
             continue
 
         for asset in release.assets:
--- a/bin/download_github_releases.py
+++ b/bin/download_github_releases.py
@@ -26,8 +26,12 @@ def download_file(url, destination_directory=None):
     logger.debug(f"-> Downloading {url} to {local_filename}")
 
     with requests.get(url, stream=True) as r:
-        with open(local_filename, "wb") as f:
-            shutil.copyfileobj(r.raw, f)
+        if r.status_code == 200:
+            with open(local_filename, "wb") as f:
+                shutil.copyfileobj(r.raw, f)
+        else:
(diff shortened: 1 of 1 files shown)


bin/download_github_releases.py
```diff
@@ -26,8 +26,12 @@ def download_file(url, destination_directory=None):
     logger.debug(f"-> Downloading {url} to {local_filename}")
 
     with requests.get(url, stream=True) as r:
-        with open(local_filename, "wb") as f:
-            shutil.copyfileobj(r.raw, f)
+        if r.status_code == 200:
+            with open(local_filename, "wb") as f:
+                shutil.copyfileobj(r.raw, f)
+        else:
+            logger.error(f"Status: {r.status_code}, body: {r.text}")
```
n1k0, line 33: 🙏

bin/download_github_releases.py
```diff
@@ -68,13 +72,24 @@ def download_file(url, destination_directory=None):
...
             continue
 
         for asset in release.assets:
             if "-dist.tar.gz" in asset.browser_download_url:
                 file_path = download_file(
                     asset.browser_download_url, args.destination_directory
                 )
+                if not file_path:
+                    logger.error(
+                        f"Error downloading {asset.browser_download_url}, skipping."
```
n1k0, line 89: 👌

n1k0 (approved):
Will the logged error show up in Sentry? I suppose so! All good 👍

vjousse:
> Will the logged error show up in Sentry? I suppose so! All good 👍

As far as I know, sentry as never been enabled on the python side :grimacing:

n1k0:
> As far as I know, sentry as never been enabled on the python side 😬

Argh, you're right! Would be nice to be notified when release archives are unavailable or things like that… but I suspect we'd get many errors as it would depend on GH infrastructure availability, which has proven *slightly* unreliable over time.

PR #781, opened 2024-10-01
feat: add bookmarks for objects

vjousse:
## :wrench: Problem

The first version of the object interface doesn't manage bookmarks.

## :cake: Solution

Implement the bookmark mechanism already available for the food and textile sections.

The `toString` on the `Query` object to find a default name for the bookmark may be questionnable.

Go on the review app https://ecobalyse-pr781.osc-fr1.scalingo.io/#/object/simulator and try to save/load simulations for the object interface using the bottom right tab.

Screenshot 2024-10-01 at 14-41-57 Simulateur Ecobalyse

### DIFF
--- a/src/Page/Object.elm
+++ b/src/Page/Object.elm
@@ -12,12 +12,12 @@ import Autocomplete exposing (Autocomplete)
 import Browser.Dom as Dom
 import Browser.Events
 import Browser.Navigation as Navigation
+import Data.Bookmark as Bookmark exposing (Bookmark)
 import Data.Dataset as Dataset
 import Data.Example as Example
 import Data.Impact as Impact
 import Data.Impact.Definition as Definition exposing (Definition)
 import Data.Key as Key
-import Data.Object.Db exposing (Db)
 import Data.Object.Process as Process exposing (Process)
 import Data.Object.Query as Query exposing (Query)
 import Data.Object.Simulator as Simulator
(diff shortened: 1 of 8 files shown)


src/Views/Comparator.elm
```diff
@@ -121,17 +122,29 @@ addToComparison session label query =
...
+        Bookmark.Object objectQuery ->
+            objectQuery
+                |> ObjectSimulator.compute session.db
+                |> Result.map
+                    (\_ ->
+                        { complementsImpact = Impact.noComplementsImpacts
+                        , impacts = Impact.empty
+                        , label = label
+                        , stepsImpacts = Impact.noStepsImpacts
+                        }
```
n1k0, line 134: Why don't we assign the computed impacts to `impacts` here?

src/Page/Object.elm
```diff
@@ -115,6 +144,14 @@ initFromExample session uuid =
     )
 
 
+findExistingBookmarkName : Session -> Query -> String
+findExistingBookmarkName { store } query =
+    store.bookmarks
+        |> Bookmark.findByObjectQuery query
+        |> Maybe.map .name
+        |> Maybe.withDefault (query |> Query.toString)
```
n1k0, line 152: Nit: `Query.toString query`

n1k0 (commented):
Looks promising! We should have the example name in the bookmark name text input when one is selected, as in textile and food though.

src/Views/Comparator.elm
vjousse: Good question 😅 Fixed in the latest commit.

n1k0 (approved):
Let's land this so we can iterate further.

PR #782, opened 2024-10-02
feat: add object api

vjousse:
## :wrench: Problem

We have no way to query the object computations using an API.

## :cake: Solution

Like for food or textile, implement API endpoints for objects.

_**TODO**_


vjousse:
Already done in other PRs

PR #785, opened 2024-10-07
chore: remove `airTransportRatio` from examples

vjousse:
## :wrench: Problem

Following the merge of https://github.com/MTES-MCT/ecobalyse/pull/757, `airTransportRatio` field is not needed anymore in exemples.

[Notion card](https://www.notion.so/Supprimer-la-valeur-fixe-airtransportratio-des-exemples-06d8fb81da934bb3a9c1f8783063e302)

## :cake: Solution

Remove the field from the examples file.

Tests should still be green and scores should still be the same.
The test scenario described in this [Notion card](https://www.notion.so/Adapter-la-part-de-transport-en-avion-automatiquement-en-fonction-de-la-durabilit-76a7514bedb44d3298f401d69f09672f) should still work.

### DIFF
--- a/public/data/textile/examples.json
+++ b/public/data/textile/examples.json
@@ -64,7 +64,6 @@
       "countryFabric": "RAS",
       "countryDyeing": "RAS",
       "countryMaking": "RAS",
-      "airTransportRatio": 1,
       "business": "large-business-without-services",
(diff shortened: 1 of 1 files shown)


n1k0 (approved):
LGTM

PR #786, opened 2024-10-07
fix: reset physical durablility in regulatory mode

vjousse:
## :wrench: Problem

When switching from exploratory mode to regulatory mode, the physical durability is not reset.

[Notion card](https://www.notion.so/Correction-du-coeff-de-durabilit-227d9ba113354f4388de9a6e49cff9ff)

## :cake: Solution

Add `physicalDurability` to the advanced query check.

- Go to the regulatory mode
- Choose "Pull coton chine fast fashion"
- Put the physical durability to 0.67
- Switch to regulatory tab and check that this popup appears
image
- Confirm
- On the regulatory tab,  the durability coefficient should be back to 0.92
image


n1k0 (approved):
LGTM

PR #794, opened 2024-10-09
Agregate in python

ccomb:
## :wrench: Problem

## :cake: Solution

### DIFF
--- a/data/common/export.py
+++ b/data/common/export.py
@@ -2,6 +2,8 @@
 import functools
 import json
 import logging
+from collections import OrderedDict
+from copy import deepcopy
 
 import bw2data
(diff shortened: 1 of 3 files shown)


paulboosz (approved):

data/common/export.py
```diff
@@ -2,6 +2,8 @@
 import functools
 import json
 import logging
+from collections import OrderedDict
```
ccomb, line 5: actually I'm no more sure it's usefull to use an OrderedDict. We remove it?

paulboosz: Im using to keep the same key order in the impacts object, is it bad ?

PR #799, opened 2024-10-10
fix: improve object simulator.

n1k0:
## :wrench: Problems

- The simulator doesn't show masses (total and per-item)
- The "add item" button is missing when the product items list is empty.
- When selecting an existing example or bookmark, we'd like to fill its name in the new bookmark text input. 
- We'd also like a more meaningful query string representation than what we currently have.
- We want to render the steps impact distribution chart in the sidebar.

## :cake: Solution

Implement aforementioned tasks.

- Check that the simulator renders total and per-item masses
- Check that we always have an "add item" button rendered.
- Check that an existing bookmark or example name is filled in the bookmark input text.
- Check that suggested bookmark names are more convenient than what we have now.
- Check that the step impacts distribution chart is rendered properly.

### DIFF
--- a/public/data/object/examples.json
+++ b/public/data/object/examples.json
@@ -1,5 +1,13 @@
 [
   {
+    "id": "0d08a9ff-5683-4d3d-a7b8-dce74f37cd3b",
+    "name": "Produit vide",
+    "category": "",
+    "query": {
+      "processes": []
+    }
+  },
+  {
     "id": "7d78d30e-7c35-451f-b8ab-e590f39ed0e8",
     "name": "Chaise",
     "category": "",
--- a/src/Data/Object/Simulator.elm
+++ b/src/Data/Object/Simulator.elm
@@ -1,15 +1,28 @@
 module Data.Object.Simulator exposing
-    ( availableProcesses
+    ( Results(..)
+    , availableProcesses
     , compute
     , computeItemImpacts
+    , emptyResults
+    , extractImpacts
+    , extractMass
     )
 
 import Data.Impact as Impact exposing (Impacts)
 import Data.Object.Process as Process exposing (Process)
 import Data.Object.Query as Query exposing (Item, Query)
+import Mass exposing (Mass)
 import Quantity
 import Result.Extra as RE
 import Static.Db exposing (Db)
+
+
+type Results
+    = Results
+        { impacts : Impacts
+        , items : List Results
+        , mass : Mass
+        }
 
 
 availableProcesses : Db -> Query -> List Process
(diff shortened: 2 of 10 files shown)


n1k0:
@vjousse this is now ready for review.

src/Data/Object/Simulator.elm
```diff
@@ -22,19 +35,67 @@ availableProcesses { object } query =
...
+                (\(Results { impacts, mass }) (Results acc) ->
+                    Results
+                        { acc
+                            | impacts = Impact.sumImpacts [ impacts, acc.impacts ]
+                            , items = Results { impacts = impacts, items = [], mass = mass } :: acc.items
+                            , mass = Quantity.sum [ mass, acc.mass ]
+                        }
+                )
+                emptyResults
+            )
```
vjousse, line 54: Nice one!

public/data/object/examples.json
```diff
@@ -1,4 +1,12 @@
 [
+  {
+    "id": "0d08a9ff-5683-4d3d-a7b8-dce74f37cd3b",
+    "name": "Produit vide",
```
vjousse, line 4: Maybe we should use "objet" here and not "produit" (and in the selector title too)?

src/Page/Object.elm
```diff
@@ -411,58 +433,62 @@ addItemButton db query =
...
     ]
 
 
-itemView : Db -> Definition -> ( Query.Amount, Process ) -> Html Msg
-itemView db selectedImpact ( amount, process ) =
-    li [ class "list-group-item d-flex align-items-center gap-2" ]
-        [ div [ class "input-group w-33" ]
+itemView : Definition -> ( Query.Amount, Process ) -> Results -> Html Msg
+itemView selectedImpact ( amount, process ) itemResults =
+    tr []
```
vjousse, line 487: Maybe having a `<thead>` somewhere could ease the understanding the columns?

vjousse (approved):
Some nits but code wise it looks good to me. Just a little question: why can't we add more than 2 processes? We may allow having twice the same process?

src/Page/Object.elm
n1k0: I thought the data were rather self-explanatory? 
image

public/data/object/examples.json
n1k0: Frankly I *hate* the term "object" and would apprcoate avoiding it as much as possible in the UI, but that's just… me, so…

n1k0:
> Some nits but code wise it looks good to me. Just a little question: why can't we add more than 2 processes? We may allow having twice the same process?

@vjousse because we only have two processes available and having two different amount of plastic would look weird I think… but that's probably some feedback we should get from POs.

vjousse:
Shouldn't we display the density to to help understand the computation?

n1k0:
I think it should be rendered in the explorer but here... I don't know, that would also clutter the ui. Let's wait for user feedback here.

PR #802, opened 2024-10-14
refactor: reorder and sync json data output between python and elm

ccomb:
## :wrench: Problem

- The json keys are in a different order between private and public data
- The elm aggregate outputs a different order than python
- → This is a permanent nightmare as it leads to big diffs in PRs

## :cake: Solution

The order has been synchronized between python and elm, so that aggregate does'nt change the order

Check the output keys of json have the same order between public and private repo.
Also check that the order is a decent one, with ids and names at the beginning, impacts at the end.

ecobalyse_private: export_reorder


n1k0 (commented):
Would it be possible to order props alphabetically as it's done in Elm? it really helps maintaining JSON files manually.

ccomb:
> Would it be possible to order props alphabetically as it's done in Elm? it really helps maintaining JSON files manually.

I remember that @paulboosz (so do I) prefered seeing some keys, such as alias, id, name, display_name at the top, because it's more convenient when parsing the file at sight

n1k0:
Oh, so we have conflicting opinions here. What do @vjousse think? And how can we make a decision about that?

ccomb:
Keep in mind that we don't have to maintain files by hand, as they are generated through the `export_food` and soon the `export_textile` procedure.

vjousse:
Well, I'm a lot in favor of sorting the keys by alphabetical order like we do in the elm code (and like I do everywhere I can). It's easier to find the keys when you don't know what the expected keys are. It's also easier to generate such files automatically as sorting the keys is easy to do automatically.

paulboosz:
I understand the preference for alphabetical sorting, it's more robust and easier to maintain.

But it makes life harder for us because it's harder to parse with the eye (mainly because of the huge `impacts` object).

Also people (alban, peter koch, v.colomb, other users I guess) are using the `processes.json` and convert it to csv for excel analysis. In this case they have to reorder the columns by hand to have the metadata first, the impacts after. This is cumbersome.

That's why I prefer the custom order

n1k0:
@paulboosz I don't understand why they have to convert json to csv while we already offer this feature in the explorer. Am I missing a use case I'm not aware of?

image
image

paulboosz:
> @paulboosz I don't understand why they have to convert json to csv while we already offer this feature in the explorer. Am I missing a use case I'm not aware of?
> 
> image image

Actually I forgot we had this feature !

That could definitely help with the excel analysis usecase. But right now the impacts are not in it

n1k0:
> That could definitely help with the excel analysis usecase. But right now the impacts are not in it

Sounds like a nice card to create!

paulboosz (approved):
✅ Order is coherent between ecobalyse and ecobalyse-private
✅  Custom order is pertinent

ccomb:
This PR is invalidated by PR794

PR #804, opened 2024-10-15
chore: standardize number formatting across codebase

paulboosz:
## :wrench: Problem

- Number formatting is unstable and is generating unwanted diff : 

- Also there seems to be some kind of computation instability a the 14th decimal between server and local

- It's hard to parse with the eye numbers with 14 decimals

## :cake: Solution

We format as best as we can using python when generating json. Prettier reformat it.

There should be 2 functions in python, js.

### Python

For a number `n` in python we use `f"{n:.6g}"` to produce a number in general format. But it's a string. If we write it in our json it would be enclosed by quotes "0.00123457". So we have to convert it to a float using `float()`. So to produce the "python formatted number" number the function is  `float(f"{n:.6g}")`. Then prettier reformats it.

Here are examples of what happens :

| Original Number   | Python formatted | Prettier Formatted |
|-------------------|------------------|---------------|
| 0.000123456789    | 0.000123457      | 0.000123457 |   
| 0.0000123456789    | 1.23457e-05      | 1.23457e-5 |
| 0.0000000123456789| 1.23457e-08      | 1.23457e-8 |
| 123.456789999     | 123.457          |123.457          |
| 1234567899999     | 123457000000.0      | 123457000000.0      |
| 1.23456789999e-7  | 1.23457e-07      |1.23457e-7      |
| 0.1               | 0.1              | 0.1 |
| 1.0                 | 1.0                | 1.0 |
| 1.00001           | 1.00001          |1.00001 |
| 1.000001          | 1.0                | 1.0                |

### JS

In JS its more straightforward. We can directly obtain the desired prettier-compatible formatting using a js function :

| Original Number    | Js Formatted |
|-------------------|---------------|
| 0.000123456789        | 0.000123457 |   
| 0.0000123456789      | 1.23457e-5 |
| 0.0000000123456789      | 1.23457e-8 |
| 123.456789999     |123.457          |
| 1234567899999   | 123457000000.0      |
| 1.23456789999e-7   |1.23457e-7      |
| 0.1                   | 0.1 |
| 1.0               | 1.0 |
| 1.00001      |1.00001 |
| 1.000001    | 1.0   |

## 📎  Todo

- [x] format_number function in python
- [x] tests in python
- [x] format_number function in js
- [x] tests in js

The tests in python and in js are different because the tests for js are post-prettier whereas in python we can't produce prettier-like numbers so the checks are pre-prettier

- `make export` && `make format`
- check that there is no diff in ecobalyse and ecobalyse-private

ecobalyse-private: json-number-format


data/common/__init__.py
```diff
@@ -212,3 +212,38 @@ def fix_unit(unit):
...
+        # Convert to scientific notation with 6 significant digits
+        num_py_g = f"{num:.6g}"
+        # Clean up the exponential notation
+        return float(
+            num_py_g.replace("e+0", "e").replace("e-0", "e-").replace("e+", "e")
+        )
+    return str(num)
+
+
+def format_numbers_recursively(data):
```
ccomb, line 241: The formatting can be done during the json encoding, using the cls argument of json.dump with something like that:
```
class CustomEncoder(json.JSONEncoder):
    def encode(self, obj):
        def format_numbers(obj):
            if isinstance(obj, (int, float)):
                return f"{obj:.6g}"
            elif isinstance(obj, dict):
                return {k: format_numbers(v) for k, v in obj.items()}
            elif isinstance(obj, list):
                return [format_numbers(v) for v in obj]
            else:
                return obj
        
        return super().encode(format_numbers(obj))
```
then
```
>>> json.dumps({"a": 1.23456789e-5, "b": [1.23456789e-5]}, cls=CustomEncoder)
'{"a": "1.23457e-05", "b": ["1.23457e-05"]}'
```

lib/json.js
```diff
@@ -0,0 +1,60 @@
...
+    indent,
+  )
+    .replaceAll(`"${token}`, "")
+    .replaceAll(`${token}"`, "");
+}
+
+module.exports = {
+  formatNumber,
+  serialize,
+};
```
ccomb, line 60: Like in Python, wouldn't it be much shorter to use the `replace` argument of JSON.Stringify ?
Something like:
```
function formatJSON(obj, precision = 6) {
  return JSON.stringify(obj, (key, value) => {
    if (typeof value === "number") {
      return Number(value.toPrecision(precision));
    }
    return value;
  }, 0);
}
```
then
```
> formatJSON({"a": 1.23456789e-5, "b": [1.23456789e-5], "c": 1.23456789e-5})
'{"a":0.0000123457,"b":[0.0000123457],"c":0.0000123457}'
```

ccomb (changes requested):
This is  much needed change, thanks for working on it!

lib/json.js
paulboosz: ToPrecision doesn't implement the same behavior, it goes from decimal to scientific at 1e-6
Screenshot 2024-12-05 at 16 10 20
Screenshot 2024-12-05 at 16 12 28
https://jsfiddle.net/9knf4j7x/

data/common/__init__.py
paulboosz: I tried json.dump without success but indeed json.dumps() works indeed. Implemented, thanks

data/common/export.py
```diff
@@ -81,19 +81,24 @@ def display_changes(key, oldprocesses, processes):
...
-                    * abs(processes[p]["impacts"][impact] - old[p][impact])
-                    / old[p][impact]
-                )
+                # Convert values to float before calculation
+                old_value = float(old[p][impact])
+                new_value = float(processes[p]["impacts"][impact])
+
+                # Handle case where old value is zero
+                if old_value == 0:
+                    percent_change = 100 if new_value != 0 else 0
```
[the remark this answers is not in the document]
paulboosz: done, thanks

data/common/__init__.py
```diff
@@ -212,3 +212,38 @@ def fix_unit(unit):
...
+            - "e+0" becomes "e" (e.g., "1.23e+05" → "1.23e5")
+            - "e-0" becomes "e-" (e.g., "1.23e-05" → "1.23e-5")
+            - "e+" becomes "e" (e.g., "1.23e+5" → "1.23e5")
+    """
+    if isinstance(num, (int, float)):
+        # Convert to scientific notation with 6 significant digits
+        num_py_g = f"{num:.6g}"
+        # Clean up the exponential notation
+        return float(
+            num_py_g.replace("e+0", "e").replace("e-0", "e-").replace("e+", "e")
```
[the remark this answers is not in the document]
paulboosz: Yes that's true. The purpose of the replace was for the test function to rigorously the same tests `1e-4` and not `1e-04`. But that's no use since we can't  produce numbers in python without like `1e-4`. We have to produce `1e-04` and let prettier reformat it to `1e-4`

data/common/__init__.py
```diff
@@ -214,3 +215,25 @@ def fix_unit(unit):
             return "t⋅km"
         case _:
             return unit
+
+
+def format_number(value):
+    """Format a number to a float with 6 significant digits."""
+    if value is None:
+        return "None"
```
vjousse, line 223: What's the point of those lines? I don't see why you need them as your checking the type before calling it in the function below.

data/common/__init__.py
```diff
@@ -214,3 +215,25 @@ def fix_unit(unit):
...
+    return float(f"{value:.6g}")
+
+
+class FormatNumberJsonEncoder(json.JSONEncoder):
+    def encode(self, obj):
+        def recursive_format_number(obj):
+            if isinstance(obj, (int, float)):
+                return format_number(obj)
```
vjousse, line 231: Do you really need a function call here as you don't use it anywhere else?
```suggestion
                return float(f"{obj:.6g}")
```

data/common/__init__.py
vjousse: And I think you could remove the function entirely, see my other comment.

tests/test_json_format_encoder.py
```diff
@@ -0,0 +1,39 @@
...
+    ],
+)
+def test_format_number_json_encoder(input_data, expected, test_id):
+    encoder = FormatNumberJsonEncoder()
+    result = encoder.encode(input_data)
+    # Convert expected to JSON string for comparison
+
+    assert (
+        result == expected
+    ), f"{test_id}: Expected {expected}, but got {result} for input {input_data}"
```
vjousse, line 39: As these tests are tied to the `data/` directory, I would prefer having them inside the `data/` directory directly.

vjousse (commented):
Great work, excited to see this change coming into `master`! I tested the export on my machine and it works as expected.
I've added some little remarks that I think should be addressed before merging 😊

data/common/__init__.py
paulboosz: done

tests/test_json_format_encoder.py
paulboosz: done

vjousse (approved):
Merge all the things! 🎉

ccomb (approved):

PR #808, opened 2024-10-17
feat: Textile export

ccomb:
## :wrench: Problem

The processes and materials for textile are a manual json file with copy-pasted impacts. This poses a risk of errors and a lack of traceability.

## :cake: Solution

This PR implements an automated `textile` export similar to the `food` one, that start from a `activities.json` file that splits into the final `processes.json`, `processes_impacts.json` and `material.json` and retrieves the impacts from simapro after getting the process names from brightway through a simple search string.

This implies some refactoring to share code between `food` and `textile` (and soon `object`) and separate pure functions in the `common` module.
We should also discuss about the resulting impacts changes.

```
cd data
pipenv run make clean
pipenv run make all
```
This should start from scratch by deleting the docker volumes, reimport all the databases into brightway, then do the export of food and textile.
At the end, after around 2h of hard work for the machine, you should get an empty `git diff` and passing tests.
Lasts ~210min on a 2017 notebook...

ecobalyse-private: textile_export


data/textile/export.py
```diff
@@ -1,151 +1,208 @@
...
+        "materialProcessUuid": activity["uuid"],
+        "recycledProcessUuid": activity.get("recycledProcessUuid"),
+        "recycledFrom": activity.get("recycledFrom"),
+        "name": activity["shortName"],
+        "shortName": activity["shortName"],
+        "origin": activity["origin"],
+        "primary": activity.get("primary"),
+        "geographicOrigin": activity["geographicOrigin"],
+        "defaultCountry": activity["defaultCountry"],
+        "priority": activity.get("priority"),
```
n1k0, line 75: This field is never used by anything in the entire codebase it seems, and should be therefore removed entirely.

n1k0 (commented):
Just adding a simple isolated comment here as I've not been requested a review.

data/textile/export.py
ccomb: Thx! You mean the `priority` field?
Btw I'll do another PR to do such changes

n1k0: yes, the priority field

paulboosz (approved):
Significant differences are for : 
- cotton 
- electric mixes

They've been investigated by @ccomb and confirmed to be desirable changes. It looks there was errors in the past values due to copy/pasting.

[Gsheet with the differences.](https://docs.google.com/spreadsheets/d/1RD2EyMF9MgUWH9_7zyhcrspijhnh63iv8-jSBErU8cg/edit?usp=sharing)

Overview of the differences (at most ~5% diff in total impact)
Screenshot 2024-11-04 at 15 40 34

PR #809, opened 2024-10-17
feat: add impacts per 100g for example objects.

n1k0:
## :wrench: Problem

Impacts per 100g, as we already have for Food and Textile, may be wanted for Objects in a near future.

## :cake: Solution

This patch implements just that.

[Discussion](https://mattermost.incubateur.net/fabnum-mte/pl/w6rpeucd67bemgee7jkfedofbe) is currently ongoing across POs to figure out if this is actually wanted or not.

Check that the Object examples explorer page features impacts per 100g of product.


n1k0:
It appears this is actually unwanted by POs. Closing, but we can always reopen if wanted again.

PR #810, opened 2024-10-17
chore: add tolerance to tests comparison

paulboosz:
## :wrench: Problem

We have differences in results between local computer and server-ci. Maybe because of processor architecture (arm64 vs x86).
As a results tests are failing either in local or on server-ci
image

## :cake: Solution

As a temporary fix, allow more tolerance in tests floating points comparison. Going from 1e-16 toleranc to 1e-12 seems reasonable and fix my tests.


n1k0 (approved):
LGTM 👌

PR #813, opened 2024-10-24
feat: Distinguish Objects from Veli.

n1k0:
## :wrench: Problem

Ecobalyse don't feature a simulator for Veli ([Notion Card](https://www.notion.so/VeLI_-S-lection-de-la-cat-gorie-VAE-v0-phase-de-production-ac7d63640c7444b9a314667ef7dcb001))

## :cake: Solution

Add a Veli menu entry and add scope information to Veli examples to help distinguish them from other Objects. 

Check that both menu entries point to expected pages and features (including bookmarks, url/permalinks and so on).


n1k0:
@vjousse so after validating with POs, we actually want to separate Objects and Veli pages and features while using a few transverse components, which is what this patch implements. It also introduces `ENABLE_OBJECT_SECTION` and `ENABLE_VELI_SECTION` env vars so we can define which sections should be enabled and served on a given instance.

index.js
```diff
@@ -37,8 +37,12 @@ const storeKey = "store";
 const app = Elm.Main.init({
   flags: {
     clientUrl: location.origin + location.pathname,
-    enableFoodSection: process.env.ENABLE_FOOD_SECTION === "True",
-    enableObjectSection: process.env.ENABLE_OBJECT_SECTION === "True",
+    enabledSections: {
+      food: process.env.ENABLE_FOOD_SECTION === "True",
+      objects: process.env.ENABLE_OBJECT_SECTION === "True",
```
vjousse, line 42: Maybe we should use `object` or `objects` everywhere but not mixing and matching both?

src/Data/Bookmark.elm
```diff
@@ -30,21 +32,33 @@ type alias Bookmark =
     { created : Posix
     , name : String
     , query : Query
+    , subScope : Maybe Scope
```
vjousse, line 35: Why not `scope` directly?

src/Data/Scope.elm
```diff
@@ -70,6 +74,9 @@ toLabel scope =
         Textile ->
             "Textile"
 
+        Veli ->
+            "Véhicule intermédiaire"
```
vjousse, line 78: What about "Véhicules intermédiaires" to be consistent with "Objets"?

src/Route.elm
```diff
@@ -278,3 +294,12 @@ toString route =
                     [ "textile", "simulator" ]
     in
     "#/" ++ String.join "/" pieces
+
+
+objectPrefixPath : Scope -> String
```
vjousse, line 299: Maybe use the `toString` method of `Scope` here?

src/Views/Page.elm
```diff
@@ -133,20 +133,25 @@ newVersionAlert { session, reloadPage } =
...
-          else
-            Nothing
+        , addRouteIf enabledSections.textile <|
+            Internal "Textile" Route.TextileSimulatorHome TextileSimulator
+        , addRouteIf enabledSections.food <|
+            Internal "Alimentaire" Route.FoodBuilderHome FoodBuilder
+        , addRouteIf enabledSections.objects <|
+            Internal "Objets" (Route.ObjectSimulatorHome Scope.Object) (Object Scope.Object)
+        , addRouteIf enabledSections.veli <|
+            Internal "Véhicule intermédiaire" (Route.ObjectSimulatorHome Scope.Veli) (Object Scope.Veli)
```
vjousse, line 154: Same here, maybe use the plural form?

tests/Views/FormatTest.elm
```diff
@@ -19,8 +19,8 @@ suite =
                 |> asTest "should format zero"
             , 5
                 |> Format.formatFloat 2
-                |> Expect.equal "5,00"
-                |> asTest "should not format an int rendering a specific number of 0 decimals"
```
vjousse, line 23: Why did you remove this test?

vjousse (approved):
LGTM, just some nits :+1:

src/Data/Bookmark.elm
n1k0: Because it helps disambiguating the scope of the page and the scope of the bookmark, which are different things. A page scoped with "objects" might list veli bookmarks.

src/Views/Page.elm
n1k0: Ah, now I get it.

src/Data/Scope.elm
n1k0: Got it.

PR #817, opened 2024-11-05
refactor: run data scripts without using docker

vjousse:
## :wrench: Problem

The data scripts can only easily be run using `docker`. As we may want to host he data apps on scalingo, we should have an easy way to run the data scripts and apps without using docker.

## :cake: Solution

Find a way to run the data scripts locally without requiring docker (so find a way to handle the python packaging mess).

_WIP_

_WIP_


vjousse:
Replaced by https://github.com/MTES-MCT/ecobalyse-data

PR #822, opened 2024-11-06
refactor: Switch to brightway

ccomb:
## :wrench: Problem

We're getting the impacts from simapro, on a dedicated server. There is a dual retrieval that depends on a dedicated server with simapro installed

## :cake: Solution

Switch to 100% brightway

Compare the new impacts with the old ones. You can draw the difference between brightway and simapro by running `make compare_food compare_textile`.

ecobalyse-private: switch_to_brightway


n1k0 (commented):
I don't have enough context to fully approve the patch, but this looks ok from an Elm and javascript codebase standpoint.

ccomb:
work moved to ecobalyse-data

PR #833, opened 2024-11-19
chore(data): Merge PastoEco in a single file to speedup imports and fixed linking to AGB

ccomb:
## :wrench: Problem

Pasto eco had come in several files which are imported one by one. This takes too much time as the `write_database` takes too long.
We also have the problem that pastoeco processes are unlinked to any source DB so we can't compute their impact and cannot implement them in ecobalyse.
This last point leads to wrong impacts for the grass-fed beef which is under-impacted.

## :cake: Solution

Merge the pastoeco csv files into one to speedup the import.
Link the processes to AGB 3.1.1.

If you already have a brightway loaded with data, you can just `make delete_database DB=PastoEco` then `make import_food` and check that you get a PastoEco database with PastoEco processes. Other wise, juste `make clean && make`
Also check that these processes have impacts (for ex using the brightway explorer)

ecobalyse-private: merge_pastoeco


paulboosz:
I did 
- ✅ make delete_database DB=PastoEco 
- ✅ make import_food 
- ✅ check that you get a PastoEco database with PastoEco processes
- ❌ check that these processes have impacts
 All impacts from Pastoeco 
Screenshot 2024-11-20 at 18 06 56
are 0 for me

ccomb:
I've discivered some processes were not linked to AGB due to a name missing the 'Copied from Ecoinvent' suffix. After changing that in a migration, we now have correct linking and impacts except on two processes:

- Newborn dairy calf, Conventional, Alsace, at farm gate, FR - MEANS#16615 U
- MAT 40_#16450 U

We need the first one which is in the dependencies of the ground-beef-grass-fed.  I'm trying to get the process upstream.
We don't need the second one which is unused.

camcoq-start:
Bonjour, j'ai validé ceci : 
image
qui correspond à ce que nous attendons en changement de score d'impact.

PR #834, opened 2024-11-19
refactor: unzip in a temporary folder instead

ccomb:
## :wrench: Problem

We cannot run two imports in parallel because the csv files are unzipped in the same directory.
 
## :cake: Solution

Unzip in a temporary folder instead

Extract two clones of ecobalyse in two different folders, and launch the make process at the same time. Check both work.

### DIFF
--- a/data/common/import_.py
+++ b/data/common/import_.py
@@ -1,9 +1,9 @@
 import functools
 import json
-import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


n1k0 (approved):
I didn't test the patch as indicated but the patch looks reasonable to me.

PR #840, opened 2024-11-21
refactor: add encrypted detailed files

vjousse:
## :wrench: Problem

In order not to make public the Ecoinvent data containing detailed processes impacts, this data is stored in a private repository on Github https://github.com/MTES-MCT/ecobalyse-private, so it is not included with the source code of the Ecobalyse application, even though it is necessary for its proper functioning. This frequently leads to technical problems when putting features into production, as it is very complicated to keep data synchronized between the two repositories.

## :cake: Solution

Re-store detailed processes in the main repository, but encrypt them using git encryption with the help of https://github.com/elasticdog/transcrypt

The warning `*** WARNING : deprecated key derivation used.` on the ci is ok for now, see https://github.com/elasticdog/transcrypt/issues/169

As Scalingo doesn't give access to the git repo I need to git clone it when deploying to Scalingo. Then I can run `transcrypt` to decrypt the files and copy them over.

We don't synchronize with `ecobalyse-private` anymore. To add new detailed files they will just need to be added to a commit in the main repo. You can see the diffs locally but not on Github anymore as the files are encrypted.

Depending on your OS, install `trancrypt` like explained here https://github.com/elasticdog/transcrypt/blob/main/INSTALL.md. Get the `trancrypt` key in https://vaultwarden.incubateur.net/ and add it to your `.env` file under the `TRANSCRYPT_KEY` entry. If you don't have access to the Vault, you should ask for one. 
You can check that files are encrypted by running:

    cat public/data/textile/processes_impacts.json

It should give you some cryptic content.

Then init your repo with `transcrypt` using the following command (you will need to do it only once for all):

    ./bin/run-transcrypt.sh

The processes should then be decrypted and you should be able to read them directly with:

    cat public/data/textile/processes_impacts.json

Try to change some detailed files and check that you can commit the changes in this branch without any problem (you can change the objects one).

Before commiting changes, be sure that your `.git/hooks/pre-commit` file contains the following (it should have been added by the `transcrypt` command):

```bash
# Transcrypt pre-commit hook: fail if secret file in staging lacks the magic prefix "Salted" in B64
 RELATIVE_GIT_DIR=$(git rev-parse --git-dir 2>/dev/null || printf '')
 CRYPT_DIR=$(git config transcrypt.crypt-dir 2>/dev/null || printf '%s/crypt' "${RELATIVE_GIT_DIR}")
 "${CRYPT_DIR}/transcrypt" pre_commit
```


bin/transcrypt
```diff
@@ -0,0 +1,1589 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+#
+# transcrypt - https://github.com/elasticdog/transcrypt
+#
+# A script to configure transparent encryption of sensitive files stored in
+# a Git repository. It utilizes OpenSSL's symmetric cipher routines and follows
+# the gitattributes(5) man page regarding the use of filters.
+#
```
n1k0, line 10: This script is way to large and complex to review for me, and I suppose this is just a local versioning of the transcrypt library in this repo, because Scalingo does not allow installing it as a dependency for some reason. So as we have a local version of it in the repo, why do we still need to install `transcrypt` from homebrew? Can't we always use this one?

n1k0 (approved):
Tested the patch and the decryption/encryption/git workflow, it works great 😄 A few question, nits and remarks and we're good to go 👍

README.md
```diff
@@ -73,6 +72,14 @@ Pour initialiser la base de données (attention, toutes les données présentes,
 
 ## Développement
 
+### Déchiffrage des fichiers
+
+Certains fichiers d’impacts détaillés nécessitent de configurer [`transcrypt`](https://github.com/elasticdog/transcrypt) pour les lire en local. Assurez-vous d’installer les dépendances listées sur [la page dédiée](https://github.com/elasticdog/transcrypt/blob/main/INSTALL.md) puis lancez la commande suivante :
+
+    ./bin/transcrypt -c aes-256-cbc -p 'your_password'
+
+Remplacez `your_password` par le mot de passe `transcrypt` stocké dans https://vaultwarden.incubateur.net/.
```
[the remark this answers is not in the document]
n1k0: Just seen mention of it in the description, scratch this comment.

.buildpacks
```diff
@@ -1,3 +1,4 @@
+https://github.com/Scalingo/apt-buildpack.git
 https://github.com/Scalingo/ssh-private-key-buildpack.git
 https://github.com/Scalingo/python-buildpack
 https://github.com/MTES-MCT/ecobalyse-buildpack-run.git
```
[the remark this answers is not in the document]
vjousse: Unfortunately comments are not supported. And you're right, the ssh-private builpack is not needed anymore :+1:

bin/transcrypt
vjousse: We need to discuss this part. `transcrypt` is just a bash script but it has dependencies:

> column and hexdump commands (on Ubuntu/Debian install bsdmainutils)
> if using OpenSSL 3+ one of: xxd (on Ubuntu/Debian is included with vim) or printf command (with %b directive) or perl

That's why we need to install it (to be sure the dependencies are installed). But I suppose you're right, I may remove it from the repo, download it on Scalingo when deploying and rely on the user to install it locally.

README.md
```diff
@@ -73,6 +72,14 @@ Pour initialiser la base de données (attention, toutes les données présentes,
 
 ## Développement
 
+### Déchiffrage des fichiers
+
+Certains fichiers d’impacts détaillés nécessitent de configurer [`transcrypt`](https://github.com/elasticdog/transcrypt) pour les lire en local. Assurez-vous d’installer les dépendances listées sur [la page dédiée](https://github.com/elasticdog/transcrypt/blob/main/INSTALL.md) puis lancez la commande suivante :
+
+    ./bin/transcrypt -c aes-256-cbc -p 'your_password'
```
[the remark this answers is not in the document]
vjousse: Thanks for the idea, it's a lot better this way :+1:

public/data/object/processes_impacts.json
```diff
@@ -0,0 +1,43 @@
+U2FsdGVkX1//F0iscfNH9mDXUnWSWY+xSNGaGIYLBn8m3W5izFdRJYY/LGWbadrY
+8UcLLenA5hLa6X5f8nEJ4/JMoXuqRFBgcyUjkxT0z/oV7sQEWNa+lIMzDstWaqN0
+oTXnyJBDgvqFfaA7v6kl+VIAdcr4t6M8vedGX84p32zAjI+YjLypW4ARx6KfMzsp
+66z/gLzXB8JTTnEptmt5wS3YJG05Kj7Fh1e/wqQMoLSQlLhsbyxUpumQwl10y5u1
+GDVjEyrgBFldA4gMHDibYBdYxZtr51B9TbOdG/JalSzUzufLOxA0JmEp3tcXgYZE
```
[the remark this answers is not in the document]
vjousse: What about just using `git diff master HEAD public/data/textile/processes_impacts.json` locally for example?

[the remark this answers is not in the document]
n1k0: but the versioned file is encrypted, right?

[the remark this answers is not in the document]
vjousse: Yes but not locally, it is encrypted before sending it to the remote. So local git commands should work as usual 😊

bin/build-specific-app-version.sh
```diff
@@ -197,11 +205,16 @@ mkdir -p $VERSION_DIR
 
 npm run encrypt $PUBLIC_GIT_CLONE_DIR/public/data/textile/processes_impacts.json $PUBLIC_GIT_CLONE_DIR/dist/processes_impacts_textile.json.enc
 npm run encrypt $PUBLIC_GIT_CLONE_DIR/public/data/food/processes_impacts.json $PUBLIC_GIT_CLONE_DIR/dist/processes_impacts_food.json.enc
-npm run encrypt $PUBLIC_GIT_CLONE_DIR/public/data/object/processes_impacts.json $PUBLIC_GIT_CLONE_DIR/dist/processes_impacts_object.json.enc
+
+# Objects are not present in old versions
+if [[ -f "$PUBLIC_GIT_CLONE_DIR/public/data/object/processes_impacts.json" ]]; then
+  npm run encrypt $PUBLIC_GIT_CLONE_DIR/public/data/object/processes_impacts.json $PUBLIC_GIT_CLONE_DIR/dist/processes_impacts_object.json.enc
+fi
```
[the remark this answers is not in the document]
n1k0: 🤯

public/data/object/processes_impacts.json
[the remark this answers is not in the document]
n1k0: 🤯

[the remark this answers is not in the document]
n1k0: Maybe add a line about that in the README? I think it will be a common question

.github/workflows/create_release.yml
```diff
@@ -27,6 +24,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Install transcrypt
+        run: |
+          mkdir -p $HOME/.local/bin
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          wget https://raw.githubusercontent.com/elasticdog/transcrypt/refs/heads/main/transcrypt -O "$HOME/.local/bin/transcrypt"
```
n1k0, line 31: How about linking to a stable tag? What if transcrypt the `main`branch evolves and breaks its API?

.github/workflows/node.js.yml
```diff
@@ -24,6 +24,13 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Install transcrypt
+      run: |
+        mkdir -p $HOME/.local/bin
+        echo "$HOME/.local/bin" >> $GITHUB_PATH
+        wget https://raw.githubusercontent.com/elasticdog/transcrypt/refs/heads/main/transcrypt -O "$HOME/.local/bin/transcrypt"
```
n1k0, line 31: Same remark as to linking to a stable tag

bin/run-transcrypt.sh
```diff
@@ -0,0 +1,7 @@
+# In $SOURCE_VERSION is set, we are deploying on Scalingo
```
n1k0, line 1: Nit: s/In/If

bin/run-transcrypt.sh
```diff
@@ -0,0 +1,7 @@
...
+# On scalingo, we don't have a git repo, so decrypting is handled
+# differently in `buildpack-run.sh`
+if [ -z "$SOURCE_VERSION" ]
+then
+  transcrypt -y -c aes-256-cbc -p "$TRANSCRYPT_KEY"
```
n1k0, line 6: Shouldn't we call `bin/run-transcrypt.sh` here?

buildpack-run.sh
```diff
@@ -1,9 +1,33 @@
 #!/usr/bin/env bash
 
-git clone git@github.com:MTES-MCT/ecobalyse-private.git
+echo "-> Install 'transcrypt' to /usr/local/bin"
 
-# Scalingo should set the $SOURCE_VERSION variable
-./bin/checkout-ecobalyse-private-branch.sh $SOURCE_VERSION
+mkdir -p "$PWD/.local/bin"
+export PATH="$PWD/.local/bin":$PATH
+wget https://raw.githubusercontent.com/elasticdog/transcrypt/refs/heads/main/transcrypt -O "$PWD/.local/bin/transcrypt"
```
n1k0, line 7: Same remark with linking to a stable release

buildpack-run.sh
```diff
@@ -1,9 +1,33 @@
...
+# Trick to avoid cloning the whole repo
+# Instead, only fetch the current commit
+git init
+git remote add origin https://github.com/MTES-MCT/ecobalyse.git
+# depth=1 as we don't need the history
+git fetch --depth 1 origin $SOURCE_VERSION
+git checkout FETCH_HEAD
+
+echo "-> Decrypt detailed impacts"
+transcrypt -y -c aes-256-cbc -p $TRANSCRYPT_KEY
```
n1k0, line 23: Maybe call `bin/run-transcrypt.sh` here?

n1k0 (approved):
A few remarks and comments following recent branch updates

bin/run-transcrypt.sh
vjousse: It is the content of `bin/run-transcrypt.sh` itself 😁

n1k0: Yeah the idea is to have one and only one single place to update the command if needed

vjousse: hmmm sure, but you are asking me to modify the content of `bin/run-transcrypt.sh` itself here, or did I miss something?

n1k0: Bah, I commented that on the wrong file ^^ the idea is to have a single place where we define the caommd line call ^^

bin/build-specific-app-version.sh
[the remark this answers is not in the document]
vjousse: Card created here https://www.notion.so/D-finir-une-politique-d-historique-de-versions-0b9ca63ce5884f09a24bad0ff91a711f?pvs=4

paulboosz (approved):
encrypting/decrypting works and I was able to commit changes to `processes.json`

buildpack-run.sh
```diff
@@ -1,9 +1,38 @@
...
+echo "-> Cloning 'ecobalyse' public repo at $SOURCE_VERSION"
+mkdir ecobalyse
+cd ecobalyse
+# Trick to avoid cloning the whole repo
+# Instead, only fetch the current commit
+git init
+git remote add origin https://github.com/MTES-MCT/ecobalyse.git
+# depth=1 as we don't need the history
+git fetch --depth 1 origin $SOURCE_VERSION
+git checkout FETCH_HEAD
```
n1k0, line 20: Now I think about this, how about getting the raw files directly? They should be available from `https://raw.githubusercontent.com/MTES-MCT/ecobalyse/$SOURCE_VERSION/public/data/textile/processes_impacts.json` and save us a little build time here

n1k0: Meh I just reread about the missing .git env in scalingo, scratch that I guess.

vjousse: Indeed, `transcrypt` needs a "real" git repo to work properly, so I don't really have the choice here.

PR #841, opened 2024-11-26
feat: introduce components db

n1k0:
## :wrench: Problem

[Notion card](https://www.notion.so/Migrer-les-donn-es-de-composants-Objets-dans-un-fichier-JSON-d-di-et-proposer-une-premi-re-version--d91e4f0aeebb41a8b2ad618389de02b3)

We don't have a components database and explorer for objects and veli just yet. Also, a component db and handling api would establish good foundations for further work with textile trims (though this patch doesn't make components totally generic just yet, but that'd be the next step).

## :cake: Solution

Introduce a components JSON database with reusable default components for objects and veli. Add a component explorer as well.

Ensure the explorer works well and that the objects and veli simulators still work as expected.

### DIFF
--- a/bin/build-db
+++ b/bin/build-db
@@ -63,6 +63,7 @@ const elmWithFixtures = elmTemplate
   )
   .replace("%foodProductExamplesJson%", parseAndValidate("public/data/food/examples.json", "id"))
   // Object JSON data
+  .replace("%objectComponentsJson%", parseAndValidate("public/data/object/components.json", "id"))
   .replace("%objectExamplesJson%", parseAndValidate("public/data/object/examples.json", "id"))
   .replace(
     "%objectProcessesJson%",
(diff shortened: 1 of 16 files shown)


vjousse (approved):
Tested locally, LGTM :+1: 
Nit: there is something strange in the CSV export. There is an extra column that doesn't have a name and has an `N/A` value.
```
"Nom","Famille","Catégorie","Coût Environnemental",""
"Chaise","Objets","","422.2806216805331","N/A"
"Table","Objets","","3979.434802943282","N/A"
"VAE","Véhicules intermédiaires","","8540.251035873613","N/A"
```

n1k0:
@vjousse yes, that's indeed because we're handling action buttons as a table column… please file a card about this, we should be able to exclude this sort of columns from the export.

n1k0:
@vjousse scratch it, fixed in 140a35b5e535db33fb9438b965198a182a99763a

PR #843, opened 2024-12-03
refactor: move weaving elec_pppm to textile wellknown.

n1k0:
## :wrench: Problem

This patch is part of [this notion card](https://www.notion.so/Refonte-fichiers-proc-d-s-2ac499dada8f4565baa6adf28c43cbde)

The `elec_pppm` field in textile `processes.json` file is used by only one process: Weaving. There's no point keeping it in the process datasource. Also and most importantly, this will help having a single format to express processes for all domains eventually.

## :cake: Solution

Move the weaving pick-per-meter consumption value to some static value in `Data.Textile.WellKnown`.

Ensure everything keeps working as previously, but api and unit tests should get us pretty much covered.


vjousse (approved):
LGTM :+1: Maybe we should remove it from the export too? https://github.com/MTES-MCT/ecobalyse/blob/master/data/textile/export.py#L113

PR #844, opened 2024-12-04
fix(food)!: food processes identifiers are now UUIDs

n1k0:
## :wrench: Problem

The food processes JSON files are not using UUIDs while they should. This work is part of a much broader work on [processes file format uniformization](https://www.notion.so/Refonte-fichiers-proc-d-s-2ac499dada8f4565baa6adf28c43cbde).

## :cake: Solution

This patch makes the food processes JSON source files use UUIDs and migrates all the necessary data.

## 🚨 Warnings, points to watch

- There might be some work remaining on the current data backend part, but I'm not sure (eg. impact on jupyter notebooks and so on). I'd like reviewers to ensure everything is correctly handled there.
- This patch only focuses on **food processes**, not ingredients. Migrating ingredient unique identifier format should be part of another task unrelated to the ongoing work on processes.
- This patch **breaks the API** because it changes process identifiers so we'll probably have to bump a **major version**.

The app and API should both keep working flawlessly.

ecobalyse-private: fix/update-food-process-to-use-uuids

<details>
<summary>Temporary js code that allowed me to process the uuids through the different food files here for further reference if required</summary>

```js
const crypto = require("crypto");
const fs = require("fs");

const activities = require("./data/food/activities.json");
const activities_to_create = require("./data/food/activities_to_create.json");
const processes = require("./public/data/food/processes.json");
const processesWithImpacts = require("../ecobalyse-private/data/food/processes_impacts.json");

const map = activities.reduce((acc, { id, alias }) => {
  acc[alias] = id;
  return acc;
}, {});

activities_to_create.forEach((activityToCreate, index) => {
  const uuid = map[activityToCreate.id] || crypto.randomUUID();
  activities_to_create[index] = { ...activityToCreate, id: uuid };
});

const newActivitiesToCreate = JSON.stringify(activities_to_create, null, 2);

function handleProcesses(list) {
  list.forEach((process, index) => {
    const uuid = map[process.id];
    if (!uuid) {
      throw new Error(`inspect ${process.name}`);
    }
    list[index] = { alias: process.id, ...process, id: uuid };
  });
  return list;
}

const newProcesses = JSON.stringify(handleProcesses(processes), null, 2);

const newProcessesWithImpacts = JSON.stringify(handleProcesses(processesWithImpacts), null, 2);

fs.writeFileSync("./data/food/activities_to_create.json", newActivitiesToCreate);
fs.writeFileSync("./public/data/food/processes.json", newProcesses);
fs.writeFileSync("../ecobalyse-private/data/food/processes_impacts.json", newProcessesWithImpacts);
```

</details>


vjousse (approved):
Tested locally, the app still works as expected. Can't tell if it's ok regarding the code in `data/`.

PR #846, opened 2024-12-05
refactor: Prevent the activity editor to use an already existing name

ccomb:
## :wrench: Problem

Though the id is guaranteed to be unique, two activities can have the same name.

## :cake: Solution

Prevent the duplicate during edition.

This branch is already live in the ingredients editor, try to add an activity with a duplicate name and check you get a warning.


JFraichou (commented):
I tested it and it worked. I cannot add an ingredient if its name is already used in Ecobalyse.

JFraichou (approved):
After verification, an ingredient can now be edited and saved without being stopped by the warning.

vjousse (approved):

PR #847, opened 2024-12-05
refactor: patch only AGB, not CTCPA which also contains 'AGB' in the filename

ccomb:
## :wrench: Problem

the Agribalyse source file is patched with sed at import, but CTCPA is also patched and shoud not.

## :cake: Solution

Check we target "AGB3..." and not all files containing "AGB"

`make clean import` and check you don't have `sed` errors like that:
`sed: can't read /tmp/tmpepq615gb/Export: No such file or directory`


vjousse (approved):
LGTM :+1:

PR #849, opened 2024-12-05
Ingredients

ccomb:
## :wrench: Problem

## :cake: Solution


JFraichou (commented):
the ecosystemic services (crop/animal group) are removed from every transformed ingredients

ccomb:
ingredients moved to ecobalyse-data

PR #852, opened 2024-12-09
refactor(data): move textile process "correctif" to comment

n1k0:
## :wrench: Problem

The poorly named `correctif` field actually wants to be a `comment` string. We also want all the processes files to benefit from it.

## :cake: Solution

Rename `correctif` to `comment`, so the three processes file formats for food, object and textile has this field.

Check the explorer, ui and API which should all keep working just fine, no impact changes. Also, the test suite should be green.

ecobalyse-private: refactor/move-correctif-to-correctives


vjousse (approved):
LGTM :+1: Thread on Mattermost https://mattermost.incubateur.net/fabnum-mte/pl/wyttiiup4idkx8omp1erf1c8dc

PR #855, opened 2024-12-10
feat: Generalize density, electricity, heat and waste process fields

n1k0:
## :wrench: Problem

JSON file format for processes are different for food, object and textile, we need to homogenize and standardize them so we define processes in a single fashion, most notably for defining density, elec and heat consumption and waste. This work is part of [a greater data format homogenization plan](https://www.notion.so/Refonte-fichiers-proc-d-s-2ac499dada8f4565baa6adf28c43cbde).

## :cake: Solution

Implement the `elec`, `heat`, `density` and `waste` field in the three current JSON files defining food, object and textile processes.

I didn't spend much time examining implications on data generation in the current python backend.

Detailed processes have been updated accordingly in https://github.com/MTE-extended/ecobalyse-private/pull/33, which is to review, approve and land prior to this PR.

Ensure that ui, explorer and API don't break. Scores shouldn't budge.

ecobalyse-private: feat/generalize-density-elec-heat-waste


n1k0:
I was too excited, forgot to update detailed processes files on ecobalyse-private. Hold on.

n1k0:
@vjousse @paulboosz @ccomb patch is now actually ready for review.

ccomb (approved):
Approving but I was really wondering why we keep these fields in the first place since they are originating from the Base Impact. I would have rather tried to migrate to an ACV-only process (with complements). I'm still unconfortable to add heat and electricity to an ecoinvent or agb process which are supposed to already contain them upstream.

n1k0:
> Approving but I was really wondering why we keep these fields in the first place since they are originating from the Base Impact. I would have rather tried to migrate to an ACV-only process (with complements). I'm still unconfortable to add heat and electricity to an ecoinvent or agb process which are supposed to already contain them upstream.

This has been discussed with and validated by @paulboosz: sometimes we want to use other datasources than EI which require this way of representing the process. I understand your concerns but if you don't have another proposal to allow BI and EI processes in a same data format, I think we'll have to compose with the situation and land the patch anyway.

n1k0:
I'd really like to have @paulboosz's approval for landing this patch and possibly address @ccomb's concerns 🙏

paulboosz:
> Approving but I was really wondering why we keep these fields in the first place since they are originating from the Base Impact. I would have rather tried to migrate to an ACV-only process (with complements). I'm still unconfortable to add heat and electricity to an ecoinvent or agb process which are supposed to already contain them upstream.

Good point, indeed processes from ACV-source should already contain the heat and the electricity in the impact.

However there are practical reasons for maintaining theses fields:
-  Many processes require country-specific energy adjustments (e.g., textile dyeing)
- Future use cases like battery manufacturing will need country-specific energy data
- Using elec_MJ and heat_MJ provides a simple parameterization mechanism

The alternative is creating a battery process for each country, it's possible but that would be impractical.

One thing we need to consider is how we get an automatic export of  `elec_MJ` and `heat_MJ` with brightway ? Right now we can't and it would be complex, but that would be great to avoid copy/pasting in the future

paulboosz (approved):

PR #862, opened 2024-12-16
feat(data): ensure consistent nullable alias field in all processes files.

n1k0:
## :wrench: Problem

The three `processes.json` files have an inconsistent use of an `alias` field:

- the food processes feature a non-nullable field
- the object processes don't use the field at all
- the textile processes feature a nullable alias field

## :cake: Solution

Make food and object processes using a nullable `alias` field, consistently with textile.

Note that the `alias` field in object processes is set to `null` as we don't have any use of it just yet. Remember that this patch is part of a broader [processes file format homogeneization initiative](https://www.notion.so/Refonte-fichiers-proc-d-s-2ac499dada8f4565baa6adf28c43cbde).

UI, API and tests shouldn't break.

ecobalyse-data: feat/add-processes-alias-field


vjousse (approved):
LGTM :+1:

PR #863, opened 2024-12-16
fix: broken homepage after upgrading highcharts

n1k0:
## :wrench: Problem

The UI is broken since we upgraded Highcharts.js to v12 (white page of death).

## :cake: Solution

Revert Highcharts to v11. We may upgrade to v12 later, but no need to rush.

UI shouldn't be broken anymore (no white page of death).


vjousse (approved):
Tested locally, works as expected after `npm ci` :+1:

PR #866, opened 2024-12-17
feat(data): unified, cross-domain processes file format.

n1k0:
## :wrench: Problem

This work is part of the [processes file format homogeneization work](https://www.notion.so/Refonte-fichiers-proc-d-s-2ac499dada8f4565baa6adf28c43cbde). We want a single Elm type, module and JSON file format to deal with a list of processes for all domains (food, object, and textile).

image

## :cake: Solution

This patch finalizes most of the work:

- [x] it adds the `categories` field  to object processes
- [x] it renames the food processes `identifier` field to `sourceId` (note we're switching to camelCase as discussed over Mattermost)
  - [x] it names that column *identifiant dans la source* in the process explorer
  - [x] it also adds this field to other processes files (object and textile)
- [x] all processes files now sharing the same data structure, it adds a single `Data.Process` module to deal with food, object and textile processes (and removes the old specific ones transparently)

- Special attention should be paid when reviewing the data backend part of the pacth, which I'm highly unsure of.
- Because this patch updates the data format for processes, a PR has been crafted on the private repo to track these changes: https://github.com/MTE-extended/ecobalyse-private/pull/39

Thanks to the Elm compiler and available unit tests, refactored code should be validated by the test suite. Just pay high attention to the possible impacts on the data backend as I'm not familiar at all with its shenanigans.


vjousse:
@n1k0 it looks like you forgot to rename `source_id` to `sourceId` in the textile processes?

n1k0:
@vjousse you're right, thank you buddy!

vjousse (approved):
Tested the app locally and it works as expected. Tested the data export too and it generates the same processes files on my machine. So everything looks good to me. Maybe having a second look from @ccomb or @paulboosz regarding the changes on the data pipeline could be great.

n1k0:
We need to move forward, I'm landing this 🎉

PR #867, opened 2024-12-18
fix: openapi food examples

vjousse:
## :wrench: Problem

With the change to uuids for the ingredients ids, openapi examples don't work anymore.

## :cake: Solution

Change ingredients ids to their corresponding uuids in examples.

Go to http://localhost:1234/#/api and try the food examples, they should work as expected.


n1k0 (approved):
Awesome, thank you!

PR #869, opened 2024-12-19
feat(data): validate processes files against a JSON schema.

n1k0:
## :wrench: Problem

We want to be sure that the processes files generated by the data pipeline are valid.

image

## :cake: Solution

Add a [JSON schema](https://json-schema.org/) and validate our files against it.

Check that the CI's green with the new CI action, or locally using:

```
$ npm run db:validate
```

**Note: I initially added validation for positive numbers for impacts but now I understand this might be expected for some processes and the validation must therefore be relaxed.**


vjousse (approved):
Love it 😍 I think this is something that should, later on, be added to the "output" of the data pipeline instead of here (so in the brand new `ecobalyse-data` repo), what do you think?
For the time being, it looks great here!

n1k0:
> Love it 😍 I think this is something that should, later on, be added to the "output" of the data pipeline instead of here (so in the brand new `ecobalyse-data` repo), what do you think?

Indeed, that's what I had in mind too. The idea is to add the simplest validation step to CI to improve automated QA and iterate later 👍

PR #870, opened 2024-12-19
fix(dev): npm ci error with `transcrypt`

vjousse:
## :wrench: Problem

When `transcrypt` is already configured for your repo locally and you run `npm install` you get an error saying that the current repository is already configured because the npm `postinstall` step is executed no matter what.

## :cake: Solution

Check if the repo was already configured with `transcrypt` before trying to configure it.

`npm ci` should not report any error and should display this message:

    ℹ️ `transcrypt` was already configured for this repo, skipping.
    
The ci should be green, validating the case that everything still works fine when setting up a new repo from scratch.

### DIFF
--- a/bin/run-transcrypt.sh
+++ b/bin/run-transcrypt.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # If $SOURCE_VERSION is set, we are building on Scalingo with npm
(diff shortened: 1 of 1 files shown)


n1k0 (approved):
Works perfectly as intended.

PR #872, opened 2024-12-23
refactor: abstract components.

n1k0:
## :wrench: Problem

[User story](https://www.notion.so/Abstraire-les-composants-dans-le-code-ecb8ebdfccff4175a2994e0613792466)

Today we have object-specific components, but we need them to be domain-agnostic so we can use the same code and data source format to handle food, object and textile ones.

## :cake: Solution

Move component impacts computation into `Data.Component` module including `Result` to make the calculation domain-agnostic.

As always with code refactoring, ensure the exposed/updated api makes sense and is as much future-proof as possible.

Ensure that nothing breaks:
- tests should be green, including added ones
- Object components explorer should work as previously
- Object and véli simulators should work as previously

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -74,6 +74,8 @@ type Quantity
     = Quantity Int
 
 
+{-| A data structure carrying the impacts and mass resulting from a computation
+-}
 type Results
     = Results
         { impacts : Impacts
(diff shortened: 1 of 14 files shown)


src/Data/Component.elm
```diff
@@ -0,0 +1,328 @@
...
+
+amountToFloat : Amount -> Float
+amountToFloat (Amount float) =
+    float
+
+
+{-| List components which ids are not part of the provided list of ids
+-}
+available : List Id -> List Component -> List Component
+available ids =
```
vjousse, line 116: Even if the comment helps, I would have named the parameter `usedIds` or `excludedIds`, something more explicit than just `ids`.

vjousse (approved):
LGTM, tested locally and it works as expected. Really cool to see more and more abstracted code! 👍

PR #873, opened 2024-12-26
feat(textile): implement trims.

n1k0:
## :wrench: Problem

Note: As often, there should have been more preliminary specification work for such a large feature. I had to improvise and make decisions where it made sense to me.

Notion cards: [1](https://www.notion.so/Sc-nario-par-d-faut-pour-les-trims-1d3abd5235cc46b4bf746d090f094fdd?pvs=23), [2](https://www.notion.so/Ajouter-le-bloc-accessoire-dans-le-simulateur-textile-243d30d52cd14cf6a5afc74bb270acab?pvs=23), [3](https://www.notion.so/Cr-er-les-accessoires-comme-composants-1eb3e86b167943828c8696aa0c1c1dd1?pvs=23) and [4](https://www.notion.so/trim-cr-er-une-page-dans-l-explorateur-pour-les-trims-2def3fb357fa4d5b8c1c57861d84d3f3?pvs=23)

We want to allow users to add trims to their textile simulation and update scoring accordingly. For example, most blue jeans usually come with a zip and a button, which fabrication processes also have impacts.

tired dog

## :cake: Solution

Reuse the Object UI component, logic and code api to add and quantify them in the Textile UI as well. Introduce new API endpoint and parameter to allow defining garment trims.

- there was a duplicate uuid in the textile components file and it's been updated with a new one
- don't be too scared with the diff size, most of it is because the detailed processes file has been regenerated because of this duplicate uuid issue
- the Jean example has been updated to feature one button and one zip, and scoring has been updated accordingly.
- a few minor UI and styling issues were fixed by this patch, check the commit log for details
- no specs were emitted regarding the comparator and the impacts details tabs in the sidebar… I had to improvise and went with a new Accessories entry in the lists

[Review app is available for testing here](https://ecobalyse-pr873.osc-fr1.scalingo.io/)

Reviewers should ensure:

- that the new Trims UI component in the textile simulator page works as expected (add, update quantities, remove trims)
- that the Object components UI still works flawlessly
- that the updated scoring for the example jean matches initial expectations
- that the impacts details tabs features the right data in the sidebar
- that the comparator still works as intended with no regression
- that the new `trims` parameter added to the textile API works as intended (e2e tests have been added)
- that the new `/api/textile/trims` endpoint added to the textile API works as expected


public/data/textile/components.json
```diff
@@ -20,7 +20,7 @@
     ]
   },
   {
-    "id": "0e8ea799-9b06-490c-a925-37564746c454",
+    "id": "d56bb0d5-7999-4b8b-b076-94d79099b56a",
```
n1k0, line 23: Note: this was a duplicate in the file so I created a new unique id.

src/Data/Food/Recipe.elm
```diff
@@ -649,6 +649,7 @@ toStepsImpacts trigram results =
     , packaging = getImpact results.packaging
     , transform = getImpact results.recipe.transform
     , transports = getImpact results.transports.impacts
+    , trims = Just Unit.noImpacts
```
vjousse, line 652: What is the reason for using `Just Unit.noImpacts` instead of `Nothing` here?

src/Server/Query.elm
```diff
@@ -552,6 +554,56 @@ parseMaterialId_ materials string =
...
+parseComponentItem : List Component -> String -> Result String ComponentItem
+parseComponentItem components string =
+    case String.split ";" string of
+        [ quantity, id ] ->
+            Ok ComponentItem
+                |> RE.andMap (parseComponentId components id)
+                |> RE.andMap
+                    (quantity
+                        |> String.toInt
+                        |> Result.fromMaybe ("Quantité invalide: " ++ quantity)
```
vjousse, line 575:
```suggestion
                        |> Result.fromMaybe ("Quantité invalide : " ++ quantity)
```

src/Server/Query.elm
```diff
@@ -552,6 +554,56 @@ parseMaterialId_ materials string =
...
+
+        _ ->
+            Err <| "Format d'accessoire invalide : " ++ string ++ "."
+
+
+parseComponentId : List Component -> String -> Result String Component.Id
+parseComponentId components string =
+    string
+        |> Component.idFromString
+        |> Result.fromMaybe ("Identifiant de composant invalide: " ++ string)
```
vjousse, line 598:
```suggestion
        |> Result.fromMaybe ("Identifiant de composant invalide : " ++ string)
```

vjousse (approved):
Tested locally (API and UI) and it works as expected, great work! :+1: 
Regarding the API, shouldn't we switch the amount and the id ordering? Everywhere (materials, ingredients, and so on) it's always id first and then the quantity, but not for the trims.

src/Data/Food/Recipe.elm
n1k0: Good catch!

n1k0:
> Should we add a check somewhere to be sure that ids are always unique?

@vjousse (sorry I can't reply to this comment idk why): it's specifically *because* we have this check when the data are integrated in the db that I've caught this bug.

PR #874, opened 2025-01-02
chore: remove obsolete pre-commit command.

n1k0:
This isn't useful anymore. I even wonder if it has ever worked 🤔


vjousse:
Indeed, I came across the same command in `ecobalyse-data` and I was wondering the same…

vjousse (approved):

PR #876, opened 2025-01-02
feat(data,ui): add trims to more textile examples, render them in explorer

n1k0:
## :wrench: Problem

More example garments should feature trims ([Notion card](https://www.notion.so/Sc-nario-par-d-faut-pour-les-trims-1d3abd5235cc46b4bf746d090f094fdd)). We should also render them in the textile examples explorer.

## :cake: Solution

Update textile examples with new trims data. Render them in the textile examples exporer.

Check that the updated textile examples (see notion card) feature the expected trims. Check the textile examples explorer.

### DIFF
--- a/public/data/textile/components.json
+++ b/public/data/textile/components.json
@@ -1,7 +1,7 @@
 [
   {
     "id": "86b877ff-0d59-482f-bb34-3ff306b07496",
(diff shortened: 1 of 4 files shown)


vjousse (approved):

PR #877, opened 2025-01-06
feat: allow expanding trim details.

n1k0:
## :wrench: Problem

Users want trim impact and amount details, as we have in the Object simulator.

## :cake: Solution

Allow expanding textile trim details as we allow it for object components.

Check that trim details are expanded and collapsed when clicking the black arrow next to a trim row.


vjousse (approved):

PR #878, opened 2025-01-06
feat: allow staff to access detailed impacts from explorer.

n1k0:
## :wrench: Problem

[Notion  card](https://www.notion.so/Permettre-l-quipe-ecobalyse-d-acc-der-aux-impacts-d-taill-s-de-proc-d-s-depuis-l-explorateur-5fe4870a76e4401fa3daa4df23806c7c)

Now that detailed processes impacts are encrypted in the repo, there's no UI to browse them. Ecobalyse staff (team members) should be able to browser and navigate them easily from the web UI and CSV exports.

capture

## :cake: Solution

- Check if the authenticated user is a staff member, and if so render detailed processes impacts in explorers
- Allow sorting impact columns
- Add detailed impacts columns to CSV exports for staff members

- Process explorer pages for food, objects and textile have been merged into a single Elm module, so they now all share the same UI and features, hence the LOC reduction in the diff.
- It might be a good idea to check how hackable is the check for a `staff` boolean by faking the auth endpoint response, but I suspect we could live with that anyway.

You need two accounts on Ecobalyse, one with the `is_staff` flag enabled, and the other one without it (note: your own personal existing team account probably has this flag enabled already). The [account page](https://ecobalyse-pr878.osc-fr1.scalingo.io/#/auth/) features a note about this bit being enabled when connected on the platform:

image

Then, check that:

- [ ] a logged-in staff member sees detailed impacts in explorers
- [ ] a logged-in non-staff member can't access detailed impacts in explorers
- [ ] unauthenticated users cant's access detailed impacts in explorers
- [ ] detailed impacts are also listed in the detailed process modal view (to open one, click a table row)
- [ ] detailed impacts are exported in the CSV export when authenticated as a staff member
- [ ] detailed impacts are NOT exported in the CSV export when authenticated as a regular member


vjousse (commented):
I’m trying to test it locally, but I have a strange behavior: once logged in, and the « Mon compte » menu displayed in the top right, if I refresh the page it seems that I am disconnected, the « Mon compte » link goes back to « Connexion ou inscription ». Do you see the same problem?

vjousse:
> > Do you see the same problem?
> 
> Yes and it also happens on production, so we should probably file a card about it :(

It works fine on production for me, but not on this review app, that's pretty strange 🤔

n1k0:
> the problem is with the local storage decoding. The staff field is missing in my local storage (so I suppose that the Elm json decode is failing), if I add it by hand it works again as expected.

Damn, good catch. I'll fix the bug in this one but we should probably warn in some way that decoding has failed, for another PR. I'll file a card about that.

vjousse (approved):
Awesome, LGTM!

n1k0:
@vjousse thanks! I'll wait for Camille's approval to land this patch. Meanwhile I filed an issue about the deserialization bug mentioned earlier: https://www.notion.so/Bug-d-codage-session-utilisateur-depuis-localStorage-8e135bd915a0423c89e67756f40074a1

n1k0:
@vjousse wait, actually scratch that, I'm gonna land the patch onto staging because I have a demo planned in 15 minutes 🤣

PR #879, opened 2025-01-07
fix: correct data on trims

paulboosz:
## :wrench: Problem

Some data in trims are wrong : 

- amount for component "Bouton plastique" is 0,002, should be 0,001
- process for "zip long" should be `brass`

EDIT 8/1/2024
- process for "zip court" should be `brass`
## :cake: Solution

Make the corrections !

For component "zip long", process demanded was `Brass, Nickel free {GLO}|market for| Alloc Def, S` but I can't find it so in agreement with camille coquelet we fallback to  `brass//[RoW] market for brass`

ask for camille coquelet validation


n1k0 (approved):
The changes look good from a technical standpoint. We should wait for Camille's approval before landing the patch though.

paulboosz:
image
Approved (displayName added)

PR #884, opened 2025-01-08
fix: Warn on session data decoding error.

n1k0:
## 🔧 **Problem**

Context: comment from @vjousse (https://github.com/MTES-MCT/ecobalyse/pull/878#issuecomment-2577039188):

> the problem is with the local storage decoding. The `staff` field is missing in my local storage (so I suppose that the Elm json decode is failing), if I add it by hand it works again as expected.

No error message is rendered to users when decoding their serialized session data fails.

## 🍰 Solution

Warn the user that decoding the session has failed, with useful information on how to report the problem:

image

## 🏝️ **How to test**

Tamper with serialized user session data in localstorage and check that an error message is rendered to the user.

### DIFF
--- a/src/Data/Session.elm
+++ b/src/Data/Session.elm
@@ -296,9 +296,7 @@ decodeRawStore : String -> Session -> Session
 decodeRawStore rawStore session =
     case Decode.decodeString decodeStore rawStore of
         Err error ->
-            session
-                |> notifyStoreDecodingError error
-                |> updateStore (always defaultStore)
+            session |> notifyStoreDecodingError error
 
         Ok store ->
             { session | store = store }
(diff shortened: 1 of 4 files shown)


vjousse (commented):
The error message works great!



But I think that the session is not properly reset. When I refresh the page I still have the same error and the local storage is left unchanged. FYI, I removed the field "terms_of_use" of the user object to get the error.

```json
{"comparedSimulations":[],"bookmarks":[],"auth":{"user":{"email":"foo@bar.baz","first_name":"","last_name":"","organization":"","staff":true,"token":"*****"}}}
```

n1k0:
@vjousse indeed, that's because we actually need to update localstorage with a Cmd sent to the runtime, which https://github.com/MTES-MCT/ecobalyse/pull/884/commits/130a156e79aafb389c92db8a9baba6508023140f implements :)

vjousse (approved):
Works like a charm! I may just have changed the text of the action button from « D’accord » to « D’accord, réinitialiser la session » or « Ok, réinitialiser la session » to be sure that we understand what we are agreeing to do.

PR #891, opened 2025-01-13
chore: readd score_history

vjousse:
## :wrench: Problem

With the deletion of the `data/` directory, the `score_history` was removed too, making the CI build fail on `master`.

## :cake: Solution

Add the script back as it was in the `bin/` directory.

We should question the actual pertinence of this script and how we could better maintain it, but it's not in the scope of this PR (the scope being: make `master` green again).

Go manually in Github actions https://github.com/MTES-MCT/ecobalyse/actions, choose "Score History" on the left and then run the workflow on this branch `fix/score-history`: the workflow should be green.


bin/score_history/create_table_score_history.sql
```diff
@@ -0,0 +1,16 @@
...
+id uuid NOT NULL,
+query text NOT NULL,
+mass numeric NOT NULL,
+elements text NOT NULL,
+lifecycle_step text NOT NULL,
+lifecycle_step_country text NOT NULL,
+impact text NOT NULL,
+value numeric NOT NULL,
+norm_value_ecs numeric NOT NULL
+);
```
n1k0, line 16: This deserves a bit of indentation methinks

```suggestion
CREATE TABLE score_history (
  datetime timestamp NOT NULL,
  branch text NOT NULL,
  commit text NOT NULL,
  domain text NOT NULL,
  product_name text NOT NULL,
  id uuid NOT NULL,
  query text NOT NULL,
  mass numeric NOT NULL,
  elements text NOT NULL,
  lifecycle_step text NOT NULL,
  lifecycle_step_country text NOT NULL,
  impact text NOT NULL,
  value numeric NOT NULL,
  norm_value_ecs numeric NOT NULL
);
```

n1k0 (approved):
LGTM

PR #897, opened 2025-01-13
feat: handle component process transforms

n1k0:
## :wrench: Problem

[Notion card](https://www.notion.so/Mettre-jour-la-vue-du-composant-pour-int-grer-2-proc-d-s-5dabac1fbe534c328962bc098ac3533c)

We should be able to apply transformation processes to component materials.

## :cake: Solution

- [x] Update components model to turn the current list of processes to a list of **elements**; an element is a material process and 0, 1 or more transformation processes of it.
- [x] handle sequential waste computations of transformations
- [x] handle sequential impacts computations of transformations
- [x] add electricity and heat impacts using generic processes (~RoW) for now

- Scores don't budge for now, as no components nor exemples feature transformation processes applied.
- Component model has changed. Theoretical implementation only as I don't have actual production data just yet.
- Processes from different domains are merged in memory in Static.Db: this is a temporary situation before we merge the different `processes.json` files into a single one in some future epic PR, but not this one which is large enough already.
- The large diff size is mainly due to tests added and JSON data updated.
- I've had to rename some types and functions in `Data.Component` to improve code logic and improve DX. Sorry if that adds some oise to the diff but I do think it serves the cause.
- UI has not been updated yet because the patch would be way too large to be reviewed efficiently; I offer UI + actual data to be added, used and rendered in the UI in a next PR (when data are ready for instance)

There's no impacts on the UI nor the API and it should purposely remain so. Unit tests have been added to the suite which should be green.


src/Data/Textile/WellKnown.elm
```diff
@@ -96,6 +97,7 @@ load processes =
         |> fromAlias "dyeing-fabric"
         |> fromAlias "dyeing-synthetic-fiber"
         |> fromAlias "dyeing-yarn"
+        |> fromAlias "elec-medium-region-asia"
```
vjousse, line 100: It looks like we are using aliases as primary keys here, are we ok with that? Shouldn't we use ids instead?

vjousse (commented):
From a code point of view, I don't see anything problematic here. Implementation looks great and, as far as I understand the problematic, the tests seem to cover the main cases, great job :+1: 
Just one comment regarding the well know processes, it looks strange to me to retrieve them using their alias instead of their unique ids.

src/Data/Textile/WellKnown.elm
n1k0: That's exactly the purpose of aliases: identifying well known processes for internal use with explicit names.

vjousse: Well, sure, but if we rely on them in the code, we should ensure that they are unique and not empty from the data pipeline perspective (this is something that is not enforced at the moment).
And for me, using aliases as PKs seem to defeat the purpose of the whole move from slug to uuids for the `id` field, but I may be missing something here.

n1k0: Ok, please file a dedicated issue about that, this is definitely not the scope of this card

n1k0: Or is it ? 😩

vjousse (approved):
Great work, LGTM :+1:

src/Data/Textile/WellKnown.elm
n1k0: after discussing it over mm, it's been decided to switch to relying on uuids in the code but in a next dedicated PR.

PR #900, opened 2025-01-16
fix(security): upgrade django to 5.1.5.

n1k0:
see https://github.com/MTES-MCT/ecobalyse/security/dependabot/56


n1k0:
@vjousse ping?

vjousse (approved):

PR #901, opened 2025-01-16
feat: show heat, elec, waste, density in process explorer.

n1k0:
## :wrench: Problem

Process explorers don't currently show waste, density, elec and heat fields.

## :cake: Solution

Show them.

Check that these fields are added as columns in process explorer tables and as rows in process explorer details modals.

### DIFF
--- a/src/Page/Explore/Processes.elm
+++ b/src/Page/Explore/Processes.elm
@@ -75,7 +75,7 @@ baseColumns detailed scope =
       , toValue = Table.StringValue <| .unit
       , toCell = .unit >> text
       }
-    , { label = "Electricité"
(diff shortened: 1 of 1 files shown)


src/Page/Explore/Processes.elm
```diff
@@ -73,6 +75,22 @@ baseColumns detailed scope =
       , toValue = Table.StringValue <| .unit
       , toCell = .unit >> text
       }
+    , { label = "Electricité"
```
vjousse, line 78:
```suggestion
    , { label = "Électricité"
```
🤓

vjousse (approved):

PR #903, opened 2025-01-20
Refactor/merge processes

n1k0:
## :wrench: Problem

Context: there are 3 process and component JSON db files sharing the same format mapped to three distinct in-memory dbs in Elm. We should have only one for each.

## :cake: Solution

Merge the files in-memory, add scopes to ease filtering them.

- #897 introduced a "hack" to duplicate and merge components and processes at the root of the db object but that wasn't sustainable and had to be fixed: this patch does that
- I didn't have the guts to merge the actual JSON files, because this would have way too much implication on the codebase and the patch would be too large to review efficiently and accurately. This patch is already large enough and provide a good amount of added sanity to make it useful I believe.

Ensure tests are green. Ensure process and component explorers behave as expected, most notably filtering processes according to related scope.


vjousse (approved):
Tested locally, API included, everything looks good. The changes in the static DB generation are awesome! ❤️

PR #907, opened 2025-01-22
feat: render component transforms

n1k0:
## :wrench: Problem

[Notion card](https://www.notion.so/Composants-de-n-proc-d-s-mati-res-et-n-proc-d-s-transformation-b30599ba654c46758f962eabe6b8595c)

We need to render the details of applied transform processes to each component element material.

## :cake: Solution

Render them.

image
image

A bug was fixed where energy was applied as kWh instead of MJ for heat consumption in transform process impacts computation (https://github.com/MTES-MCT/ecobalyse/pull/907/commits/8ce0548ffebdbc29f83c0787c80b32e8ae89cc34).

Check that the UI conforms to users expectations (see card)

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -169,7 +169,13 @@ applyTransforms allProcesses transforms (Results materialResults) =
                             in
                             Results
                                 { impacts = Impact.sumImpacts [ transformImpacts, impacts ]
-                                , items = Results { impacts = transformImpacts, items = [], mass = Quantity.negate wastedMass } :: items
+                                , items =
+                                    Results
+                                        { impacts = transformImpacts
+                                        , items = []
+                                        , mass = Quantity.negate wastedMass
+                                        }
+                                        :: items
                                 , mass = outputMass
                                 }
                         )
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -254,15 +254,13 @@ computeMaterialResults amount process =
                 |> Impact.multiplyBy (amountToFloat amount)
 
         mass =
-            (if process.unit == "kg" then
-                amountToFloat amount
+            Mass.kilograms <|
+                if process.unit == "kg" then
+                    amountToFloat amount
 
-             else
-                -- apply density
-                amountToFloat amount * process.density
-            )
-                |> Mass.kilograms
-                |> Quantity.multiplyBy (1 - Split.toFloat process.waste)
+                else
+                    -- apply density
+                    amountToFloat amount * process.density
     in
     -- global result
     Results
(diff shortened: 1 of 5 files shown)


src/Data/Component.elm
```diff
@@ -157,23 +160,31 @@ applyTransforms allProcesses transforms (Results materialResults) =
...
-                                            |> Impact.multiplyBy (Mass.inKilograms mass)
-                                        , heat.impacts
-                                            |> Impact.multiplyBy (Energy.inKilowattHours transform.heat)
-                                            |> Impact.multiplyBy (Mass.inKilograms mass)
-                                        ]
+                                    [ transform.impacts
+                                    , elec.impacts
+                                        |> Impact.multiplyBy (Energy.inKilowattHours transform.elec)
+                                    , heat.impacts
+                                        |> Impact.multiplyBy (Energy.inMegajoules transform.heat)
```
n1k0, line 167: Heat should never be expressed in kWh but in MJ: this is fixed by this patch

n1k0:
Hmm I realize there's waste to take into account at the material step as well. Will update patch.

n1k0:
@vjousse patch is now ready to review again after fixing the last issue in https://github.com/MTES-MCT/ecobalyse/pull/907/commits/20153bb9bb0f2daae29c51d34a4fbf1f7b24d8d3

Sorry for the multiple notifications :/

src/Data/Component.elm
vjousse: So this was a bug in the actual code? Is it a data problem that we could avoid/check at the pipeline level?

src/Views/Component.elm
```diff
@@ -66,80 +73,84 @@ componentView :
...
+    if debug then
+        pre [ class "p-2 bg-light" ]
+            [ results
+                |> Component.encodeResults (Just Definition.Ecs)
+                |> Encode.encode 2
+                |> text
+            ]
+
+    else
+        text ""
```
vjousse, line 151: Is it something that you want to keep around for the time being?

vjousse (approved):
Tested locally and it looks good to me. Code wise it looks good too!
Do you want to keep the debug function around?

src/Data/Component.elm
n1k0: Basically I converted the properly decoded energy expressed in MJ, but expressed it in kWh while the formula expect to keep the same unit. I don't see an easy solution from the data modelization standpoint.

src/Views/Component.elm
n1k0: Frankly, yes. It helped me tremendously for inspecting and debugging the code. I don't think there's a use case for general use hence why it's kept local. But I'd hate having to find this code again to reapply it the next time I need to investigate a computation rendering issue.

n1k0:
For the debug function, a thing we might want eventually is a user setting available to staff or admin users only, so one could toggle these debug views depending on the context. I created a card about that

PR #910, opened 2025-01-27
feat(textile,ui): apply default trims on product category change

n1k0:
See [Notion card](https://www.notion.so/Les-cat-gories-en-textile-doivent-avoir-les-trims-par-d-fauts-eb163196faa841da941283c831e1c42f) which exhaustively describes the use case and expectations.

### DIFF
--- a/public/data/textile/products.json
+++ b/public/data/textile/products.json
@@ -21,6 +21,7 @@
     },
     "name": "Chemise",
     "surfaceMass": 200,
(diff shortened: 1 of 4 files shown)


vjousse (approved):
Tested locally works as expected. I tested bookmarked products not having the `trim` property by tweaking the local storage and it works fine too (doesn't display any trim) :+1:

PR #915, opened 2025-01-30
chore: check ecobalyse-data sync for PR

paulboosz:
## :wrench: Problem

We have 2 repos `ecobalyse-data` and `ecobalyse` so we have out of sync issues.

I experienced them and it's quite annoying.

We agreed upon a workflow with the following rules : 
- `ecobalyse-data` PR should be merged first
- Only then can a corresponding syncing`ecobalyse` PR can be merged

With this workflow, `ecobalyse-data/main` should always be sync with `ecobalyse/branch_a` when doing a PR on `ecobalyse`:

## :cake: Solution

In every `ecobalyse` PR add a check that verifies if `ecobalyse-data` and `ecobalyse` are in sync for the generated `ecobalyse-data` files : 
- "public/data/food/ingredients.json"
- "public/data/food/processes.json"
- "public/data/textile/materials.json"
- "public/data/textile/processes.json"
- "public/data/object/processes.json"

- As `ecobalyse/master` might be temporarily behind `ecobalyse-data/main`, this check should not apply in the CI on `ecobalyse/master`. Only on PR to `ecobalyse/master`

- If someone merge a json-changing PR in `ecobalyse-data` and he doesn't merge the corresponding PR on `ecobalyse`, then this check is going to block all `ecobalyse` PR because they will all be out of sync. But that's kind of the point, it forces us to always sync `ecobalyse-data` and `ecobalyse`

- I didn't add `public/data/textile/processes_impacts.json` to the check as it was more complicated with the encryption. Any modification to `processes_impacts.json` should normally modify `processes.json` so it's not that important

- Unit testing check-ecobalyse-data-sync.sh would be nice

### Success
  - running `./check-ecobalyse-data-sync.sh ` should succeed (if this is sync to `ecobalyse-data`)

### Failure
  - add a difference in one of the generated json (listed above)
  - running `./check-ecobalyse-data-sync.sh ` should fail and display the diff

### DIFF
--- a/.github/workflows/ecobalyse-data-sync.yml
+++ b/.github/workflows/ecobalyse-data-sync.yml
@@ -11,10 +11,5 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install wget
-      run: |
-        sudo apt-get update
-        sudo apt-get install -y wget
-
     - name: Check synchronization with ecobalyse-data for generated JSON
       run: ./check-ecobalyse-data-sync.sh
--- a/.github/workflows/ecobalyse-data-sync.yml
+++ b/.github/workflows/ecobalyse-data-sync.yml
@@ -12,4 +12,4 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Check synchronization with ecobalyse-data for generated JSON
-      run: ./check-ecobalyse-data-sync.sh
+      run: ./bin/check-ecobalyse-data-sync.sh
--- a/.github/workflows/ecobalyse-data-sync.yml
+++ b/.github/workflows/ecobalyse-data-sync.yml
@@ -1,0 +1,20 @@
+name: ecobalyse-data sync
+
+on:
+  pull_request:
+    branches: [ master, staging ]
+  workflow_dispatch:
+
+jobs:
+  check-ecobalyse-data-sync:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install wget
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wget
+
+    - name: Check synchronization with ecobalyse-data for generated JSON
+      run: ./check-ecobalyse-data-sync.sh
--- a/check-ecobalyse-data-sync.sh
+++ b/check-ecobalyse-data-sync.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+set -euo pipefail
+
+# Directory where ecobalyse-data will be cloned
+TEMP_DIR=$(mktemp -d)
+ECOBALYSE_DATA_REPO="https://github.com/MTES-MCT/ecobalyse-data.git"
+ECOBALYSE_DATA_BRANCH="${ECOBALYSE_DATA_BRANCH:-main}"
+
+# Cleanup function
+cleanup() {
+    rm -rf "$TEMP_DIR"
+}
+trap cleanup EXIT
+
+# Clone ecobalyse-data repository
+echo "Cloning ecobalyse-data repository (branch: ${ECOBALYSE_DATA_BRANCH})..."
+git clone -q -b "${ECOBALYSE_DATA_BRANCH}" "$ECOBALYSE_DATA_REPO" "$TEMP_DIR"
+
+# Files to check
+FILES=(
+    "public/data/food/ingredients.json"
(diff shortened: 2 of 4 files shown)


.github/workflows/ecobalyse-data-sync.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  check-ecobalyse-data-sync:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install wget
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wget
```
vjousse, line 17: I’m pretty sure this is not needed as you can do the same with `curl -O` and curl should be available, no?

check-ecobalyse-data-sync.sh
vjousse, line 1: This file should be under `bin/`

check-ecobalyse-data-sync.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+set -euo pipefail
+
+# Directory where files will be downloaded
+TEMP_DIR=$(mktemp -d)
+ECOBALYSE_DATA_BRANCH="${ECOBALYSE_DATA_BRANCH:-main}"
+RAW_GITHUB_URL="https://raw.githubusercontent.com/MTES-MCT/ecobalyse-data/${ECOBALYSE_DATA_BRANCH}"
```
vjousse, line 7: Nit: you may want to make the URL configurable with an env variable as you did with the branch name here

check-ecobalyse-data-sync.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "Downloading files from ecobalyse-data repository (branch: ${ECOBALYSE_DATA_BRANCH})..."
+
+# Create necessary directories in TEMP_DIR
+for file in "${FILES[@]}"; do
+    mkdir -p "$TEMP_DIR/$(dirname "$file")"
+done
+
+# Download each file
+for file in "${FILES[@]}"; do
+    wget -q "$RAW_GITHUB_URL/$file" -O "$TEMP_DIR/$file" || {
```
vjousse, line 33: See my comment above, switching to `curl` if it is available by default in the ubuntu image would be preferred.

check-ecobalyse-data-sync.sh
```diff
@@ -0,0 +1,73 @@
...
+        echo "--⚠️  Failed to download $file from ecobalyse-data repository"
+        DIFFERENCES_FOUND=1
```
vjousse, line 35: I don’t think setting `DIFFERENCES_FOUND` to 1 here makes any sense? (you’re resetting it to 0 below anyway)

vjousse (commented):
Great initiative! I’ve added some comments that I would like to be addressed before merging. Don't hesitate if you want to discuss about it.

check-ecobalyse-data-sync.sh
```diff
@@ -0,0 +1,73 @@
...
+        continue
+    }
+done
+
+echo "Comparing JSON files between ecobalyse-data and ecobalyse repositories..."
+DIFFERENCES_FOUND=0
```
n1k0, line 41: As already mentioned, this should be defined earlier in the file

n1k0 (commented):
Great idea. I've mostly thumbed up @vjousse review comments here.

.github/workflows/ecobalyse-data-sync.yml
paulboosz: done

check-ecobalyse-data-sync.sh
paulboosz: done

check-ecobalyse-data-sync.sh
paulboosz: done

check-ecobalyse-data-sync.sh
paulboosz: indeed ! done

check-ecobalyse-data-sync.sh
paulboosz: done

paulboosz:
Thanks for the review @vjousse  @n1k0. 
I implemented the suggested improvements

vjousse (approved):
LGTM! :+1:

PR #917, opened 2025-02-04
fix: re-allow overriding spinning country.

n1k0:
## :wrench: Problem

[Notion card](https://www.notion.so/Pays-inconnu-textile-filature-1b9b056852fc804c98b1c5e664220415)

#713 introduced a possible regression where the spinning country couldn't be overridden by the user anymore when selecting the "Pays inconnu (par défaut)" value, because the main material country was always used instead any user provided alternative.

Though we need to decide what behavior we do actually want when no user value is provided here:

- Always use the main material country as a fallback
- Always use the unknown country as a fallback

## :cake: Solution

The patch currently implements the first second option so we can decide what we want. (Edit: decision has been made to go with the second option, see Notion card linked above)

Scoring changes.

Check that selecting the "Pays inconnu (par défaut)" at the Spinning step actually applies the default Electricity mix from India as per documented in the methodology docs.


src/Data/Textile/Inputs.elm
```diff
@@ -181,7 +181,10 @@ fromQuery { countries, textile } query =
         |> RE.andMap mainMaterialCountry
         -- Spinning country is either provided by query or fallbacks to material's default
         -- country, making the parameter optional
-        |> RE.andMap (getCountryResult mainMaterialCountry query.countrySpinning)
+        -- FIXME: choose between unknownCountryResult and mainMaterialCountry:
+        --        - mainMaterialCountry: always use material country whatever user chooses (weird)
+        --        - unknownCountryResult: fallback to unknown country (IN) when no value's provided
+        |> RE.andMap (getCountryResult unknownCountryResult query.countrySpinning)
```
n1k0, line 187: For now I'm leaving this comment here so we can decide what behavior we want.

n1k0: It's been decided to always unknown country value (IN)

n1k0:
@vjousse ping

vjousse (approved):

PR #939, opened 2025-02-12
chore: use new deployment stack `scalingo-22`

vjousse:
## :wrench: Problem

The Scalingo stack that we are using will end its support in April 2025.

## :cake: Solution

Migrate to scalingo 22 https://doc.scalingo.com/platform/internals/stacks/stacks#migrating-to-a-new-stack

The change should be transparent.

Deployment should be green and the review app should still work as before.


n1k0 (approved):
Everything works as intended 👌

PR #951, opened 2025-02-27
chore(food,textile,api)!: Remove deprecated API endpoints.

n1k0:
## :wrench: Problems

- `GET` + query string endpoints must be removed completely following deprecation started in 4.0.0 (#954).
- `POST` parameters are not fully validated (eg. some JSON properties skip through range validation)

## :cake: Solution

1. Remove these API endpoints which already have a POST+JSON equivalent:
    - `GET /food`
    - `GET /textile/simulator`
    - `GET /textile/simulator/{impact}`
    - `GET /textile/simulator/detailed`
1. Update tests so they use their POST counterparts
1. Ensure POST parameters are properly validated

## ⏰ Watch out

This patch clears a bunch of tech debt by introducing a dedicated, isolated layer of validation, independent from decoding. Special attention should especially be paid to the newly introduced `Data.Validation` API, don't hesitate to challenge its design.

- API docs should not mention the removed endpoints anymore
- The removed endpoints should obviously not respond anymore
- Test suite should be green


n1k0:
Updated patch, title and description to remove the query-string based endpoints completely because we don't need a deprecation step as we have API versioning to circumvent that.

n1k0:
Switching back to Draft because this depends on #954 to land first.

vjousse (approved):
This is an awesome work, impressive 😍 I’ve added some remarks and nits, but it’s such an awesome improvement that it should land as soon as possible. I’m approving it so that you’re not blocked, but open to discuss it further if you want.

PR #952, opened 2025-02-27
fix: explicit strategies and htc/htn fixes

ccomb:
Output from https://github.com/MTES-MCT/ecobalyse-data/pull/44 and https://github.com/MTES-MCT/ecobalyse-data/pull/59


paulboosz (approved):
I have the same impacts. 

Let's merge this and always stay in sync from now on

PR #959, opened 2025-03-10
feat: allow customize component element mass

n1k0:
This is WiP

### DIFF
--- a/src/Data/Object/Query.elm
+++ b/src/Data/Object/Query.elm
@@ -65,7 +65,7 @@ addComponentItem id query =
     { query
         | components =
             query.components
-                ++ [ { id = id, quantity = Component.quantityFromInt 1 } ]
+                ++ [ { custom = Nothing, id = id, quantity = Component.quantityFromInt 1 } ]
     }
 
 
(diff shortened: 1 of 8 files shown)


src/Data/Object/Query.elm
```diff
@@ -78,18 +79,65 @@ removeComponent id ({ components } as query) =
...
+                                        else
+                                            Nothing
+
+                                    Nothing ->
+                                        Just
+                                            { elements = updateElements component.elements
+                                            , name = Nothing
+                                            }
+                        }
                     )
```
vjousse, line 141: Maybe replacing it with a local function like the `updateElements` one could enhance the readability?
And it looks like the `case` could be replaced by a `Maybe.withDefault`?

vjousse (approved):
I tested the changes locally and, as far as I can tell, everything works as expected!
I added a little nit as a review, everything else looks good to me. The test you added was the tests I expected to see, good job :+1:

src/Data/Object/Query.elm
n1k0: I swear I tested many refactors but this version is the more legible one I could find. I'm all ears for your take on this, but otherwise I'm just gonna land this as is.

n1k0: slightly improved legibility in 0dfb7d12b08c4e320a622398cb239227fa4f9195

PR #966, opened 2025-03-18
fix(security): upgrade django to 5.1.7.

n1k0:
> An issue was discovered in Django 5.1 before 5.1.7, 5.0 before 5.0.13, and 4.2 before 4.2.20. The django.utils.text.wrap() method and wordwrap template filter are subject to a potential denial-of-service attack when used with very long strings.

https://github.com/MTES-MCT/ecobalyse/security/dependabot/59

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ name = "pypi"
 python_version = "3.12"
 
 [packages]
-django = ">=5.1.5,<6"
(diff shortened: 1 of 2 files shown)


n1k0:
@vjousse poke

vjousse (approved):

PR #968, opened 2025-03-19
chore: data JSON export refactor

vjousse:
## :wrench: Problem

See https://github.com/MTES-MCT/ecobalyse-data/pull/62

Tests should be green, and the app should load as usual (especially regarding the ecosystemic services)


n1k0 (commented):
I see `"scenario": null` a bunch of times and it's always `null`. Is it actually necessary? 

Also, I see a lot of `sourceId` being `null` while the source is `Ecoinvent X.X.X`; the purpose of these two fields combined together is to be able to retrieve the original lci from its original source (the unique source id could be a technical name if we don't have anything else)… Are we okay losing this ability?

vjousse:
The `scenario` field is used to compute the ecosystemic services. Where do you see that it’s null?
To be honest, I don’t see the point of having the `sourceId` field at all as we don't have reliable primary keys in the LCA databases…

vjousse:
Indeed, the "primary key" in the database is the `name` field, so that’s why I don’t see the point of this `sourceId` field…

n1k0 (approved):
I'm gonna approve this patch but I'd really like we kept track of a value for the `sourceId` to uniquely identify a source lci, whatever format it'd be (but I understand it's impossible, which worries me a lot with the "database technologies" we currently rely on.

Edit: missed your comment in between, so I'm amending here:

> Indeed, the "primary key" in the database is the name field, so that’s why I don’t see the point of this sourceId field…

If we ensure the `name` field never changes and is never updated nor updatable (eg. from a back office page), then indeed. But in my humble personal experience, a `name` field, even along with a `displayName` one, is so tempting to be updatable that I usually find that a more technical &and explicitely read-only name like `sourceId` is safer.

Well, your call.

PR #969, opened 2025-03-20
feat(object,ui): allow updating a component element material.

n1k0:
I'm embarrassed as I didn't really find a matching card in the backlog, but I really really wanted to tackle this.

This patch allows updating a component element material.

Edit: created a Notion card https://www.notion.so/Permettre-de-changer-la-mati-re-d-un-l-ment-de-composant-1c0b056852fc804d8b03dc317fe0595c


public/data/object/processes.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "categories": [],
+    "categories": ["material"],
```
vjousse, line 3: Shouldn’t we add the `material` category to the `Composant en plastique (PP)` too?

src/Views/Component.elm
```diff
@@ -360,7 +379,7 @@ elementTransformsView config component index transformsResults transforms =
 
                     -- Note: allows truncated ellipsis in table cells https://stackoverflow.com/a/11877033/330911
                     , style "max-width" "0"
-                    , title <| Process.getDisplayName transform
+                    , title transform.name
```
vjousse, line 382: Oh nice! If I get it right, it allows you to display the `displayName` in the cell and at the same time having the process name on hover?

vjousse (approved):
As far as I could test locally, everything works as expected! Good job 😊

src/Views/Component.elm
n1k0: yes! Best of both worlds :)

n1k0:
Fixed #971

PR #981, opened 2025-03-27
feat(object,ui): allow set a custom component name (fixes #980)

### LINKED ISSUE #980: Allow editing a custom component name
## :wrench: Problem
As a user, when I modify a component to customize it to suit my needs, I want to be able to change its name, so I can share it efficiently to other paople.

## :cake: Solution
Allow updating a component name.

This will also be able for integration in a future backend.

- [ ] ensure a component name can be edited
- [ ] ensure a textile trim name can't

n1k0:
See #980


vjousse (approved):
Works fine! 😍 What about always trimming the name before saving it?

PR #989, opened 2025-03-31
feat: store and render app version along bookmarks

### LINKED ISSUE #982: Store versioned queries in localStorage
## :wrench: Problem
The bookmark queries are currently stored ion localStorage without the app version information attached. This should be required to compute scoring against an expected version, and/or discard those using previous versions.

## :cake: Solution
Attach the version information to stored serialized queries, render it when bookmarks are listed.

- [ ] The version used when creating a new bookmark should be rendered:
  - [ ] in the bookmarks sidebar
  - [ ] in the compared bookmarks list in the comparator modal
- [ ] If version isn't a tag BUT we have a commit hash (eg. review app), render it using a specific color (eg. warning)
- [ ] If we don't have version information at all because a bookmark has been saved prior of this feature being implemented, render a specific information about that (specific color, tooltip text, etc)

n1k0:
Fixes #982

### DIFF
--- a/src/Request/Version.elm
+++ b/src/Request/Version.elm
@@ -1,16 +1,21 @@
 module Request.Version exposing
     ( Version(..)
     , VersionData
+    , decodeVersionData
+    , encodeVersionData
     , getTag
     , is
     , loadVersion
     , pollVersion
+    , toMaybe
     , updateVersion
     )
 
+import Data.Common.DecodeUtils as DU
 import Data.Github as Github
 import Json.Decode as Decode
 import Json.Decode.Pipeline as Pipe
+import Json.Encode as Encode
 import RemoteData exposing (WebData)
 import RemoteData.Http as Http
 import Time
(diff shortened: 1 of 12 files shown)


src/Request/Version.elm
```diff
@@ -23,38 +28,46 @@ type alias VersionData =
 
 
 type Version
-    = NewerVersion
+    = NewerVersion VersionData VersionData
```
vjousse, line 31: Maybe naming the types with old/new may help the readability here?

vjousse (approved):
Hard to fully test without having a "real" version released but as far as I could test locally, everything works as expected! Really great improvement 😊

src/Request/Version.elm
n1k0: https://github.com/MTES-MCT/ecobalyse/pull/989/commits/5cfa3a0ecffeb0799557d4ce44c8fdb4af175041

PR #1001, opened 2025-04-01
feat(object,ui): allow multiple component instances

### LINKED ISSUE #994: Allow using a component multiple times
## :wrench: Problem
Currently, we can't use a component multiple times in the object simulator. 

## :cake: Solution
Allow using a component multiple times in the simulator.

This is because we identify each component in the list by its identifier, so it must be unique. We should switch to indexing so we don't have to rely on their identifier anymore.

Ensure that users can leverage a given component multiple times in the simulator and that there's no functional regression otherwise.

n1k0:
Fixes #994

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -124,16 +124,20 @@ type alias ExpandedElement =
     }
 
 
+type alias Index =
+    Int
+
+
 {-| Index of an item element and associated source component
 -}
 type alias TargetElement =
-    ( Component, Int, Int )
+    ( TargetItem, Index )
 
 
 {-| Index of an item and associated source component
 -}
 type alias TargetItem =
-    ( Component, Int )
+    ( Component, Index )
 
 
 {-| An amount of some element
@@ -666,7 +670,7 @@ extractMass (Results { mass }) =
 {-| Remove an element from an item
 -}
 removeElement : TargetElement -> List Item -> Result String (List Item)
-removeElement ( component, itemIndex, elementIndex ) =
+removeElement ( ( component, itemIndex ), elementIndex ) =
     updateItemCustom ( component, itemIndex )
         (\custom ->
             { custom
@@ -724,7 +728,7 @@ updateCustomElement ( component, elementIndex ) update =
 
 
 updateElement : TargetElement -> (Element -> Element) -> List Item -> List Item
-updateElement ( component, itemIndex, elementIndex ) update =
+updateElement ( ( component, itemIndex ), elementIndex ) update =
     updateItem itemIndex <|
         \item ->
             { item
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -670,8 +670,8 @@ extractMass (Results { mass }) =
 {-| Remove an element from an item
 -}
 removeElement : TargetElement -> List Item -> Result String (List Item)
-removeElement ( ( component, itemIndex ), elementIndex ) =
-    updateItemCustom ( component, itemIndex )
+removeElement ( targetItem, elementIndex ) =
+    updateItemCustom targetItem
         (\custom ->
             { custom
                 | elements =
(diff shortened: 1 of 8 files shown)


src/Data/Component.elm
```diff
@@ -123,6 +124,18 @@ type alias ExpandedElement =
     }
 
 
+{-| Index of an item element and associated source component
+-}
+type alias TargetElement =
+    ( Component, Int, Int )
```
vjousse, line 130: Do we agree that it would be the same than `( TargetItem, Int )`? And what about renaming the `Int` type to a custom `Index` type for example (same for the `Int` in `TargetItem` below)?
`(TargetItem, Index)` would make more sense to me than `(Component, Int, Int)` and `(Component, Index)` more sense than `(Component, Int)`, what do you think?

src/Data/Component.elm
```diff
@@ -181,6 +202,25 @@ addItem id items =
...
+            items |> setElementMaterial ( targetItem, elementIndex ) process
+
+        ( Category.Material, Nothing ) ->
+            items |> addElement targetItem process
+
+        ( Category.Transform, Just elementIndex ) ->
+            items |> addElementTransform ( targetItem, elementIndex ) process
+
+        ( Category.Transform, Nothing ) ->
+            Err "Un procédé de transformation de peut être ajouté qu'à un élément existant"
```
vjousse, line 218:
```suggestion
            Err "Un procédé de transformation ne peut être ajouté qu'à un élément existant"
```

src/Page/Object.elm
```diff
@@ -49,7 +49,7 @@ type alias Model =
     , bookmarkName : String
     , bookmarkTab : BookmarkView.ActiveTab
     , comparisonType : ComparatorView.ComparisonType
-    , detailedComponents : List Component.Id
+    , detailedComponents : List Int
```
vjousse, line 52: Should’nt it be `Index` instead of `Int`?

src/Page/Object.elm
```diff
@@ -64,38 +64,38 @@ type Modal
...
-    | RemoveElementTransform Component Int Int
+    | RemoveComponentItem Int
+    | RemoveElement TargetElement
+    | RemoveElementTransform TargetElement Int
     | SaveBookmark
     | SaveBookmarkWithTime String Bookmark.Query Posix
     | SelectAllBookmarks
     | SelectNoBookmarks
-    | SetDetailedComponents (List Component.Id)
+    | SetDetailedComponents (List Int)
```
vjousse, line 88: Same, `Index`?

src/Page/Object.elm
```diff
@@ -64,38 +64,38 @@ type Modal
...
     | SwitchComparisonType ComparatorView.ComparisonType
     | SwitchImpact (Result String Definition.Trigram)
     | SwitchImpactsTab ImpactTabs.Tab
     | ToggleComparedSimulation Bookmark Bool
     | UpdateBookmarkName String
-    | UpdateComponentItemName Component String
-    | UpdateComponentItemQuantity Component.Id Component.Quantity
-    | UpdateElementAmount Component Int (Maybe Component.Amount)
+    | UpdateComponentItemName TargetItem String
+    | UpdateComponentItemQuantity Int Component.Quantity
```
vjousse, line 97: Same, `Index`?

src/Page/Textile.elm
```diff
@@ -119,13 +118,12 @@ type Msg
     | OnStepClick String
     | OpenComparator
     | RemoveMaterial Material.Id
-    | RemoveTrim Trim.Id
+    | RemoveTrim Int
```
vjousse, line 121: Same, `Index`?

src/Page/Textile.elm
```diff
@@ -153,7 +151,7 @@ type Msg
     | UpdateStepCountry Label Country.Code
     | UpdateSurfaceMass (Maybe Unit.SurfaceMass)
     | UpdateTraceability Bool
-    | UpdateTrimQuantity Trim.Id Trim.Quantity
+    | UpdateTrimQuantity Int Component.Quantity
```
vjousse, line 154: Same, `Index`?

vjousse (approved):
Some nits but otherwise LGTM!

PR #1022, opened 2025-04-14
refactor: decouple ingredient/material id from process id

paulboosz:
## :wrench: Problem

[Validated ecobalyse-data PR](https://github.com/MTES-MCT/ecobalyse-data/pull/73)

Currently, an activity generates both a process and potentially an ingredient that share the same ID. An ingredient is essentially an extension of a process with additional metadata.

This design has several issues:
1. Ingredients and processes are different entities/concepts but share the same ID.
2. It's impossible for multiple ingredients to use the same process

 As a result, we're calculating the same process impacts multiple times unnecessarily (e.g., multiple vegetables using the same "Cauliflower, winter, conventional, at farm gate {FR} U" process).

## :cake: Solution

Decouple ingredient and process Ids by :
- Keeping the same ids for ingredient/material
- Assigning each process a deterministic uuid based on the key : `database_name:activity_name`
- Deduplicating activities based on the previous key
- Adding `processId` field in ingredient/material to reference their associated process

- This changes the ids of processes (food/textile/object)
- This does not change the ids of ingredients/materials
- This makes breaking changes for the API for process Ids (eg: cooking, packaging,...)

## TODO :

- [x] Update harcoded process ids in `countries.json, WellKnown.elm materials.json components.json products.json` 
- [x] fix ingredient - process relationship, currently it uses `default` as a foreign key. It should use `processId`
- [x] fix tests

- everything should be green


n1k0 (approved):

paulboosz:
approved and ecobalyse-data is synced -> merged
image

PR #1036, opened 2025-04-16
chore: Replace "cotton Row ginning" with "market for cotton GLO"

ccomb:
new export from https://github.com/MTES-MCT/ecobalyse-data/pull/82


ccomb:
cotton RoW is already a sort of GLO, as a weighted mean of CN, IN, US

PR #1039, opened 2025-04-16
chore: correctly set sourceId

paulboosz:
## :wrench: Problem
Currently :
-  `sourceId` is optional in `processes.json`, making it unreliable for finding corresponding activities in LCA databases
-  The `name` field is actually being used as the real `sourceId` since LCA databases don't use hash-like IDs

## :cake: Solution

- Make `sourceId` a required field that contains the exact LCA activity name from the source database
- delete obsolete field `default`
- Remove `name` field to avoid confusion

- [x] fix the elm frontend


n1k0:
@paulboosz updated the Elm code counterpart. I think it's safe @vjousse gives a proper look at the patch :)

vjousse:
Looks good to me but before approving, we should discuss the fact that `source_id` is still optional here this https://github.com/MTES-MCT/ecobalyse-data/pull/67#discussion_r2055922324 in the `ecobalyse-data` repo

vjousse (approved):

PR #1045, opened 2025-04-23
chore: create local test component data

### LINKED ISSUE #1044: Use controlled data for components unit tests
## :wrench: Problem

refs #1014

Unit tests must rely on controlled data fixtures instead of production data which may evolve, needing constant test updates to match them

## :cake: Solution

Use local test fixtures in components unit tests

Tests should be green

n1k0:
fixes #1044


vjousse (approved):
Awesome, great move! 👍

PR #1059, opened 2025-04-29
fix(food,ui): render ingredient technical process name tooltip

### LINKED ISSUE #1057: The technical name of processes has disappeared
### Problem

Since v5.0, there is a bug:

ICV seen rolling hover the ingredients is not the same anymore because the technical name of processes has disappeared and been replaced by the display name

In v4.0.1
Image

Image

Image

Now, in v5.0

Image

Image

Image

### Expected fix

We expect same icv than in 4.0.1

n1k0:
fixes #1057


paulboosz (approved):
LGTM

vjousse (approved):

PR #1072, opened 2025-05-06
refactor: merge processes in a single file for cross-domain reusability

paulboosz:
## :wrench: Problem

[Issue ](https://github.com/orgs/MTES-MCT/projects/53?pane=issue&itemId=104257020&issue=MTES-MCT%7Cecobalyse%7C984)
- Currently we have 3 processes file, 1 for each domain (food, textile, object).
- We can't reuse processes between domains (for example electricity, transport, plastic, metal,...)

## :cake: Solution

- have only 1 processes file with a scope field, for example : 
`"scopes": ["food", "textile"]`
- delete all the domain-specific processes files

- I deleted the `category` field of textile processes. This field was only present in domain textile :( and was used to filter textile material. I replaced it with a `textile_material` tag in `categories`
- the processes explorer will need to be updated
- processes.json is now sorted by id to avoid irrelevant diff

- [ ] Tests should pass
- [ ] Explorer view of processes needs to be updated
- [x] Release and version creation should be modified


n1k0 (commented):
After a quick first look, I couldn't find any process scoped for `veli`, is this expected?

paulboosz:
> After a quick first look, I couldn't find any process scoped for `veli`, is this expected?

Before they weren't any `processes.json` for veli either nor any explorer for veli processes. My guess is that `veli` and `object` share the same processes. I'll add a `veli` scope to `object` processes to make this more explicit

n1k0:
@vjousse I migrated the frontend o use the single processes file, though I struggled updating the version handling stuff, most notably the `build-specific-app-version` script and express server code. Would you mind giving it a look and possibly update that part?

n1k0 (approved):
LGTM

PR #1083, opened 2025-05-12
fix: properly decode json processes on login

### LINKED ISSUE #1084: JSON db decoding error on login
## :wrench: Problem
JSON decoding error occurs when loadin the page with an auth token set in localStorage:

Image

Same when unauthenticated and login in:

Image

n1k0:
Fixes #1084 

How to test on the review app:
- [ ] Log in and you shouldn't get a JSON decode error
- [ ] Refresh the page in the browser : you shouldn't get a JSON decode error
- [ ] When authenticated, you should have access to detailed impacts


paulboosz (approved):
I tested and it worked :)

PR #1086, opened 2025-05-12
chore: update ecobalyse data sync

paulboosz:
## :wrench: Problem

Github action `ecobalyse-data sync`is not up to date with the merging of processes.json in a single file
## :cake: Solution

Update Github action `ecobalyse-data sync` so that it check the sync of the single file `processes.json`


n1k0 (approved):
LGTM

PR #1103, opened 2025-05-23
chore: use agb32 instead of agb31

ccomb:
Use Agribalyse 3.2
See https://github.com/MTES-MCT/ecobalyse-data/pull/33


ccomb:
Replaced with #1201

PR #1119, opened 2025-06-02
fix(textile): compute printing impacts from surface

### LINKED ISSUE #1095: investigate print feature (delete unused process)
### Problem

We have 4 printing processes in textile explorer. 
2 with m2 as unity and 2 with kg. 

Which one is used in Textile LCA calculator ? both are used. 
We need to clarify this.

### Solution

For the 2 processes custom : 97c209ec-7782-5a29-8c47-af7f17c82d11 and cfdc5e31-25fc-56ff-9a54-04670ecad301
- delete elec and heat value 
- change unity for m2 (no kg)

For the formula, only take in account the surface and not the weight. #1119 

### Acceptance criteria

- [ ] update [documentation](https://fabrique-numerique.gitbook.io/ecobalyse/textile/cycle-de-vie-des-produits-textiles/ennoblissement-1)

### Test Scenario

Test 1: Teeshirt made in China (100% printed)

- [ ] "Ennoblissement" step will have 494pts impact due to "inventaires enrichis"

Test 2: Teeshirt majorant (india) 100% printed
- [ ] "Ennoblissement" step will have 962pts impact due to "inventaires enrichis"

Test 3: Teeshirt made in China (20% printed)
- [ ] "Ennoblissement" step will have 174pts impact due to "inventaires enrichis"

n1k0:
Fixed #1095 (work in progress)


vjousse (approved):
Can’t tell about the feature per see without more context, but looks fine technically

camcoq-start:
To give more details : 
- we will ignore the product mass to calculate printing impact and just use the surface calculated and the printed surface ratio 

We wait a new check of maurine about data and then i will create the issue corresponding to the need

PR #1130, opened 2025-06-10
ci: use uv instead of pipenv for `score_history`

vjousse:
## :wrench: Problem

The `score_history` script is [failing on `master`](https://github.com/MTES-MCT/ecobalyse/actions/runs/15559004198/job/43806436083) due to the move from `pip` to `uv`.

## :cake: Solution

Switch to `uv` and remove the need for the backend server in the CI as we don’t use it in this script.

Manually dispatch a new action `score_history` workflow on this branch, it should be green like this one: https://github.com/MTES-MCT/ecobalyse/actions/runs/15559884369/job/43809282316

### DIFF
--- a/.github/workflows/score_history.yml
+++ b/.github/workflows/score_history.yml
@@ -99,10 +99,8 @@ jobs:
           npm run server:build
           scalingo db-tunnel DATABASE_URL &
           npm run server:start &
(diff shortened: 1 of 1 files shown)


n1k0 (approved):
LGTM

Issue #1097, opened 2025-05-21
Change exemple values in textile calculator for tshirt coton - remanufacturé


camcoq-start:
### Problem

Tshirt coton - remanufacturé  has same price, references and company type than fast fashion product

### Solution

Change values for this exemple: 
- price: 30€
- references: 200
- company type: PME/TPE
- tracability: true

### Test scenario

Select Tshirt coton - remanufacturé
Check values
Expected durability coefficient: 1,45
Expected environmental cost: 281 points

n1k0:
@camcoq-start PR is ready for testing at https://github.com/MTES-MCT/ecobalyse/pull/1136

PR #1136, opened 2025-06-11
chore(textile): update upcycled tshirt example.

n1k0:
refs #1097


paulboosz (approved):
LGTM

I'll let @camcoq-start confirm cause Ecoscore of upcycled tshirt is 298 Pts, not 281 Pts like expected in #1097

Issue #1101, opened 2025-05-22
Mass of trims is exclude from total mass product until "confection" stage


camcoq-start:
### Problem

Trims had been added to textile LCA calculator but the calculator logic is based on product's final mass. (ex: t-shirt= 0,15kg). 
To obtain a 0,15kg final product, LCA calculator take account of each raw material loss in every stage.
This is a problem with trims because trims's mass is included in each life cycle stage (from raw material to end of life), but the mass is only added at the confection stage. 
So, we sub-estimate the amount of raw material at the first stage.

### Solution

- exclude trims's mass in the following steps: "Matière première", "transformation - Confection", "transformation - Ennoblissement", "transformation - tissage", "transformation - fillage"
- Add trim's mass after the confection stage. 
- trim's mass is included in the mass transported between confection stage and distribution stage.

### Acceptance criteria
- raw material is not based on all the losses and exclude trim's mass
- from the confection stage to the end of life stage, trim's mass is included. The product mass is the final product mass that user firstly enter as value.

### Test scenario (in progress)

Test 1: jean coton (450g) - Majorant par défaut:
- product's final mass: 0,45kg
- trims : 0,013kg
- loss 

- [ ]

PR #1139, opened 2025-06-11
fix(textile): exclude trims weight before the Making step

n1k0:
refs #1101


vjousse (approved):
LGTM from a tech POV 👍

PR #1141, opened 2025-06-12
fix: handle reusing outdated magic links.

n1k0:
## :wrench: Problem

Users may get a disturbing "Unauthorized" error notification when clicking a previously used magic link, this is because the backend API responds with a generic 401 when a magic link token has already been consumed. 

## :cake: Solution

In this specific situation and if the user is already signed in, don't render any error message.

The patch also automatically discards auth session data when the server is desynchronized (eg. the user doesn't exist anymore in the db, which happens when a new review app is deployed or locally during testing)

Also, I didn't see a point notifying the user that the link has already been used, as he's already signed in anyway.

- [ ] While already signed in, try to reuse a previsouly used magic link: no error message should show up
- [ ] While already signed in, try deleting the user associated to the signed-in account from the database, and load the account page: an error should notice that the session has been discarded, and the user should be logged out

### DIFF
--- a/src/Page/Auth.elm
+++ b/src/Page/Auth.elm
@@ -244,10 +244,20 @@ updateAccountTab session currentAuth profileForm _ msg model =
             )
 
         ProfileResponse _ _ (RemoteData.Failure error) ->
-            ( model
-            , session |> Session.notifyBackendError error
-            , Cmd.none
-            )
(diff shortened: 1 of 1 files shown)


vjousse (approved):
Tested locally and acceptance tests are ok. LGTM!

PR #1148, opened 2025-06-16
feat: add journaling of actions

### LINKED ISSUE #1146: Add journalisation of admin actions (for components)
## :wrench: Problem

We need to know who has created/updated/deleted a component in the admin to be able to follow the history of changes for a component.

## :cake: Solution

Add a journal of actions performed on the components.

> TODO

vjousse:
fixes #1146


n1k0 (commented):
Looks neat from a non-Python expert pov, though I'd like to implement a new view in the backoffice so we can browse these events. BTW, is there a way to paginate the results?

Issue #1120, opened 2025-06-03
Allow create a textile component in Object


camcoq-start:
### Problem

There is many specificities in Textile LCA simulator. 
Those specificities cannot be handled in component approach but we need to create textile component for chair or sofa or curtains.

### Solution

Get Textile LCA simulator compatible with component and so allow Alban create textile component for object.

Create the following processes (these are existing process but in front code): 
- [x] 3 spinning processes : 

- Nom= filature conventionnelle (40 Nm), Nom technique = filature conventionnelle (40 Nm), Source = Custom, unit= kg Electricité = 3,64Kwh  , chaleur= 0MJ , perte= 12% , densité= n/a, categories= manufacturing, material type = fibres organique, 
- Nom= filature non conventionnelle (40 Nm), Nom technique = filature non conventionnelle (40 Nm), Source = Custom, Kwh= 1,82 , MJ=0, loss= 12%, categories= manufacturing, material type = fibres organique 
- Nom= filage (40 Nm) , Nom technique = filage(40 Nm), Source = Custom, Kwh= 1,36 , MJ= 0 , loss= 3%, categories= manufacturing, material type = fibres synthétiques 

- [x] 1 weaving process:
- Nom = tissage ( 40Nm et 250 g/m2), Nom technique= tissage ( 40Nm et 250 g/m2), Source = Custom, unit = kg, chaleur = 0MJ , Electricité= 5,82Kwh* , material type = fibres organiques, fibres synthétiques, loss = 6,25

5,82*= 250(g/m2)*40 (Nm)/1,08/2*1000(g)/250(g/m2)*0,0003145 --> this is the function used in Textile LCA to calculate elec quantity. 

Note1 : for now, just one process is too limited to fully integrate Textile LCA simulator in component approach. It is enough for Object.

- [x] Add to categories for pre-treatment processes, the material type:
- bleaching : material type = fibres organiques
- degreasing : material type = fibres organiques
- washing : material type = fibres synthétiques

- [x] Add to the following raw materials textile processes, the material type:
- Production de PET, granulés : fibres synthétiques
- Production de fibre de coton: fibres organiques
- Production de fibre de coton bio: fibres organiques

Note3: For now we don't add the microfibre complement. This is mostly due to the using stage and this parameter lives in the material json file. (Material json file is excluded in component approach)
 
### Acceptance criteria

- [x] Create a component with raw and manufacturing textile processes in object simulator
[...]

PR #1157, opened 2025-06-18
feat: textile component in object

paulboosz:
## :wrench: Problem

#1120


camcoq-start:
test feedback: 

- when i am authenticated, i haven't any component in the list. So i am not able to test textile component creation
- when i am not authenticated, i have a list of 3 component.

n1k0:
Why am I assigned this? Can I have a bit more context? I don't understand the feedback in https://github.com/MTES-MCT/ecobalyse/pull/1157#issuecomment-2987960724

PR #1164, opened 2025-06-21
feat(ui): revamp app notification system

n1k0:
WiP


n1k0:
> From a pure functional point of view, notifications seems to work as they should. Would it be hard to add a close button to the notifications even if they disappear after X seconds? From an UI point of view, I love to be able to close the notification before the timeout, but it’s just a nice to have to me.

Good idea, I was wondering the same. Will implement 👍

PR #1180, opened 2025-06-26
feat(component): filter transforms by material constraints

### LINKED ISSUE #997: List transformation process according raw material process properties
### Problem

When i create an element in a component, i select raw material process. Then i have to select a transformation process. List of transformation process should depend on raw material process (material type and unity). 

Notice that this feature will be used in all sector. For instance: 
- in "origin" field in materials.json file, we use the value organic fibre to select pre-treatment like bleaching
- in "food categories" in ingredients.json file, we use for exemple "meat, fish ..." to chose the right transport, or adapt electricity value for the cooking process

### Solution

Filter processes according to unity and according to tag in "categories". 

### Acceptance criteria

- [x] in object, processes have a material type (metal, wood)
- [x] in an element, raw material process define the manufacturing process listed
- [x] i can't add a transformation process with same material type if unity is different
- [ ] i can only add transformation process with same material type and unity than raw material process

### Test scenario

- [x] Scenario 1: in object,

- i add "plastique granulée (PP)" in an element of a component
- then i add a transformation process
- the list is restricted to plastic transformation process "moulage par injection", "extrusion plastique"

- [ ] Scenario 2: in object,

- i add "Bois feuillu chêne" in an element of a component
- then i click on "add a transformation process"
- transformation process listed is only "sciage-séchage" because it is material type = wood and unity = m3
- i can't see "revêtements" because it is a transformation process

- [x] Scenar 3 : change raw material element

- delete the tranformation process (back to empty element)

- [x] Scenario 4 : 

- add "acier non allié"
- transformation are "transformation metallique"

n1k0:
fixes #997

### DIFF
--- a/src/Data/Process.elm
+++ b/src/Data/Process.elm
@@ -11,10 +11,13 @@ module Data.Process exposing
     , findById
     , getDisplayName
     , getImpact
+    , getMaterialType
     , getTechnicalName
     , idFromString
     , idToString
     , listByCategory
+    , listByMaterialType
+    , listByUnit
     )
 
 import Data.Common.DecodeUtils as DU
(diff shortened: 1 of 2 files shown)


vjousse (approved):
LGTM from a code point of view. It seems that you’ve decided to not throw an error if there are multiple material types for the same process, but instead, take the first one. Looks good to me as it something that will be managed in the backend 👍

n1k0:
@vjousse this is actually a very good point, I can totally check for multiple material types and filter for any of them. Thanks for the idea!

PR #1189, opened 2025-06-29
refactor: handle empty responses from the backend api

n1k0:
## :wrench: Problem

The Elm backend http client doesn't handle an empty http response by (eg. for 204 responses), let's fix that.

## :cake: Solution

Default empty body string to `"null"` so it's always valid JSON.

Backend API now responds with 204 for deletions.

Tests have been updated, so… not much. Maybe manually ensure that deleting a component from the admin doesn't explode your computer.


vjousse (approved):
A lot cleaner this way, thanks for the improvement 👍

PR #1208, opened 2025-07-02
feat(ui): improve alert on existing user account

n1k0:
image


vjousse (approved):
A lot better! Maybe displaying a warning or an error Alert would be better?

camcoq-start (commented):
cool, thank you

PR #1209, opened 2025-07-02
fix: backend url rewriting

vjousse:
## :wrench: Problem

- When browsing the `v6.0.0` version in the selector, a `/` is added to the `clientUrl` causing http errors because of `//` in urls
- Versions under the `/versions/v6.0.0/` url for example are doing HTTP calls to `/versions/v6.0.0/backend` instead of `/backend`

## :cake: Solution

- Use a regex to remove all trailing `/` in clientUrl
- Patch the v6.0.0
- Use a nginx rewrite rule for backend calls

I was unable to find how to manage the `.proxyrc.json` of Parcel to handle URLs rewriting in a clean way, so I’ve hardcoded the version number for now. Happy to receive some help here.

The best solution for me would be to use nginx locally like in production to avoid those kind of problems and possible errors.

The `v6.0.0` release have been patched and deployed in production.

In the review app, browse to the `v6.0.0` version in the selector and try to login/register/see detailed impacts: it should work as expected.

### DIFF
--- a/.proxyrc.json
+++ b/.proxyrc.json
@@ -4,6 +4,13 @@
     "target": "http://127.0.0.1:8002/",
     "timeout": 30000
   },
+  "/versions/v6.0.0/backend": {
+    "target": "http://127.0.0.1:8002/",
+    "pathRewrite": {
+      "^/versions/v6.0.0/backend": ""
+    },
+    "timeout": 30000
+  },
   "/backend": {
     "target": "http://127.0.0.1:8002/",
     "pathRewrite": {
--- a/backend/src/app/cli/commands.py
+++ b/backend/src/app/cli/commands.py
@@ -69,8 +69,10 @@ async def _create_user(
 
     async with alchemy.get_session() as db_session:
         users_service = await anext(provide_users_service(db_session))
-        user = await users_service.upsert(data=obj_in.to_dict(), auto_commit=True)
-        console.print(f"User created: {user.email}")
+        user = await users_service.upsert(
+            data=obj_in.to_dict(), auto_commit=True, match_fields=["email"]
+        )
+        console.print(f"User upserted: {user.email}")
 
 
 @user_management_group.command(
--- a/packages/python/ecobalyse/ecobalyse/patch_files.py
+++ b/packages/python/ecobalyse/ecobalyse/patch_files.py
@@ -64,6 +64,24 @@ def patch_index_js_file(index_js_file: pathlib.Path, version: str):
     write_patched_data(nb_patched, data, index_js_file)
 
 
+def patch_double_slash(
+    index_js_file: pathlib.Path, patch_file: pathlib.Path, git_dir: pathlib.Path
+):
+    with (
+        open(index_js_file, "r") as index_file,
+    ):
+        data = index_file.read()
+        if data.count("replace") > 0:
+            logger.info(
+                f"Patch content already present in `{index_js_file}`, skipping."
+            )
+        else:
+            logger.info(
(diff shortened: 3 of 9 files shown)


.proxyrc.json
```diff
@@ -4,6 +4,13 @@
     "target": "http://127.0.0.1:8002/",
     "timeout": 30000
   },
+  "/versions/v6.0.0/backend": {
+    "target": "http://127.0.0.1:8002/",
+    "pathRewrite": {
+      "^/versions/v6.0.0/backend": ""
+    },
+    "timeout": 30000
+  },
```
n1k0, line 13: wow :( I suppose we don't have many more options for a quick new release

backend/src/app/cli/commands.py
```diff
@@ -69,8 +69,10 @@ async def _create_user(
 
     async with alchemy.get_session() as db_session:
         users_service = await anext(provide_users_service(db_session))
-        user = await users_service.upsert(data=obj_in.to_dict(), auto_commit=True)
-        console.print(f"User created: {user.email}")
+        user = await users_service.upsert(
+            data=obj_in.to_dict(), auto_commit=True, match_fields=["email"]
+        )
```
n1k0, line 74: 😅👌

packages/python/ecobalyse/ecobalyse/patch_files.py
```diff
@@ -64,6 +64,24 @@ def patch_index_js_file(index_js_file: pathlib.Path, version: str):
...
+        data = index_file.read()
+        if data.count("replace") > 0:
+            logger.info(
+                f"Patch content already present in `{index_js_file}`, skipping."
+            )
+        else:
+            logger.info(
+                f"Applying patch file `{patch_file}` to `{git_dir}` using `git apply`."
+            )
+            subprocess.run(["git", "apply", patch_file], check=True, cwd=git_dir)
```
n1k0, line 82: 🙈🙉🙊

servers.conf.erb
```diff
@@ -22,6 +22,15 @@ server {
...
   }
+  #
+  # Reroute versions to backend
+  location ~ ^/versions/[^/]+/backend/(.*)$ {
+    rewrite ^/versions/[^/]+/backend/(.*)$ /$1 break;
+    proxy_pass http://localhost:8002;
+    proxy_set_header Host $host;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+  }
```
n1k0, line 33: I have a hunch we should come with a more general approach with double slash handling… btw, why isn't the nginx `merge_slashes` directive, why is enabled by default, work in our case? 🤔

vjousse: Here the problem I’m solving is not the double slash ending. It’s the fact that http calls to `/versions/*/backend/*` should be sent to the Litestar backend (localhost:8002) and the path rewritten to keep only the part after the `/backend/` string (Litestar doesn’t know anything before the `/api`).

.proxyrc.json
n1k0: Yeah but that means docker-compose, and… well, it does mean docker-compose 🫠

n1k0: (we should investigate further but let's land this patch and release a new version first)

n1k0 (approved):

.proxyrc.json
vjousse: Sure, but we already use docker compose for PG and it works pretty well for now 😊 But yes, let’s land it as it is for now!

PR #1218, opened 2025-07-04
fix(ui): don't scrolltop on explorer modal closed

### LINKED ISSUE #1192: Explorer modal window scrolls back to the top of list
## :wrench: Problem

When I'm scrolling through a list in the explorer  (processes, ingredients...) and opening a modal window, then closing it, the window scrolls back to the top. 

I would like the view to stay where I'm at

https://github.com/user-attachments/assets/0c4a56c3-a87e-498f-9a9f-4a7234acf310

## :cake: Solution

Don't scroll back to the top

n1k0:
fix #1192 cc @paulboosz

### DIFF
--- a/src/Page/Explore.elm
+++ b/src/Page/Explore.elm
@@ -75,57 +75,53 @@ type Msg
 init : Scope -> Dataset -> Session -> PageUpdate Model Msg
 init scope dataset session =
     let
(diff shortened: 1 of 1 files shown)


paulboosz (approved):
Works great ! Using the explorer is now much more convenient 😊🙌

PR #1220, opened 2025-07-04
fix(ui): fix bogus versions listed in old versions selector

### LINKED ISSUE #1216: Bogus prereleases are still listed in version selector
## :wrench: Problem

When switching to a version older than 6.1.1, old bogus prerelease versions are still listed and available in the version selector:

Image

n1k0:
trying to fix #1216

@vjousse this is probably not done "the nicest way" but I couldn't resist trying 🤣


n1k0:
Ok, let's wait for @vjousse to be back from holidays to address this properly 😅

Issue #1075, opened 2025-05-07
Improve Ecobalyse KPI


camcoq-start:
### Problem

- Lack of traffic stats
- ecoinvent requirements (invoice based on users + CGU change) (last connexion stored in back-end)
- Lack of knowledge about users

These problem are related to kpi.
 
### Solution

Matomo is not a reliable tool. Look for another tool (ex: plausible) --> Posthog

KPI we want to follow for vanity metrics : 
- daily single visitor by page (textile, alimentaire, object, food, explorateur textile, explorateur alimentaire, explorateur object, explorateur food, api, api faq, accueil)
- daily single visitor in ecobalyse (front and api)
- successful API requests (in total, per endpoints, per version)
- successful/ended front simulation

For our statistic page, exclude week-end from stat. 

Better understand our users by following:  
- quantity of user account
- authenticated visit (total front and api, per page, per organization type)
- funnels
- track number of saved simulation, highcharts extract, clic on compare

vjousse:
I did some investigations on Plausible, everything that is mentioned above can be done except:
- user journey (first page, seconde page, third page). See https://github.com/plausible/analytics/discussions/149 for more details. Using funnels may do the trick depending on what we want to achieve here

n1k0:
There's a [Posthog guide on ADEME's Notion instance](https://notion.incubateur.ademe.fr/Bonnes-pratiques-PostHog-1976523d57d7809483bec3686c169d15). It seems there's no official/shared ADEME instance right now, so the question is how do we handle the costs?

PR #1222, opened 2025-07-07
Feat/1075 improve kpi

n1k0:
refs #1075 

- [x] Create Posthog account for Ecobalyse
- [x] Invite @vjousse and @camcoq-start 
- [x] Install Posthog deps
- [x] Allow configuring Posthog host and key through env vars
- [x] Configure Posthog to track client side events
- [ ] Configure Posthog to track API events


n1k0:
Merging with r=@camcoq-start

PR #1266, opened 2025-07-18
feat(bo): add user accounts admin

n1k0:
## :wrench: Problem

We currently have no easy way to list existing user accounts, making it complicated and inefficient to investigate user authentication issues and provide minimal support.

## :cake: Solution

Add a backend API endpoint and a UI to list user accounts in the back-office.

Click on the new `Utilisateurs` tab in the admin, and check that:

- [x] user informations are displayed ok
- [x] basic filters work as expected
- [x] sortable columns work as intended

image

### DIFF
--- a/e2e/auth.spec.js
+++ b/e2e/auth.spec.js
@@ -174,9 +174,7 @@ test.describe("auth", () => {
 
       await page.getByLabel("Menu principal").getByRole("link", { name: "Admin" }).click();
 
-      await expect(
-        page.getByRole("heading", { name: "Administration des Composants" }),
-      ).toBeVisible();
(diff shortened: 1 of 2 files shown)


camcoq-start (approved):
Thanks nico :) 

What is the meaning of "active" for users please ? How do you define an inactive user?

n1k0:
@camcoq-start after investigating a little, the `is_active` field in the db is created at account registration time, and can be changed later if needed (think account deactivation). The flag is also required for accessing several backend endpoints, meaning a non-active user can't access them (eg. profile update, api tokens, etc). Does that address your question?

vjousse (approved):

PR #1282, opened 2025-07-23
fix(textile)!: apply bleaching pretreatments energy mix impacts

### LINKED ISSUE #1240: Fix viscose pretreatment impact
### Problem

For clothes made with viscose, at the ennoblishment step, we only take in account the bleaching toxicity impact. But there is also electricy and heat consumption that participate to environmental impact. 

### Solution

Fix our way to calculate viscose ennoblishment impact by adding electricity and heat consumption by bleaching. 

### Acceptance criteria

- [ ] Viscose has pre-treatment now : "bleaching"
- [ ] can see it in front end while using exploratory mode
- [ ] score is ok

### Test scenario
skirt fully made with viscose : 
- ennoblishment = 571 pts

n1k0:
fixes #1240 (work in progress)

Notes to reviewers and/or people taking over the patch while I'm on holiday:

- the `ArtificialFromInorganic` origin is permanently removed from the codebase as there aren't any material featuring it in the current database (nor will we ever want to support it in the future); methodology documentation will be updated accordingly by Maurine (validated with @camcoq-start)
- the codebase has been refactored to compute pretreatments impacts in a single place, but we now have score diffs we can't explain, and should be kept investigated before landing
- I've added "inventaires enrichis" details in the UI so we can view/debug the pre-treatments toxicity impacts from there:
image

### DIFF
--- a/src/Views/Textile/Step.elm
+++ b/src/Views/Textile/Step.elm
@@ -864,21 +864,38 @@ ennoblingToxicityView db ({ selectedImpact, inputs } as config) current =
                     Nothing ->
                         Impact.empty
 
-            toxicity =
-                Impact.sumImpacts
-                    [ current.preTreatments.toxicity
-                    , dyeingToxicity
-                    , printingToxicity
-                    ]
+            toxicityDetails =
+                [ ( "Pré-traitements", current.preTreatments.toxicity )
+                , ( "Teinture", dyeingToxicity )
+                , ( "Impression", printingToxicity )
+                ]
         in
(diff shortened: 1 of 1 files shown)


camcoq-start:
@n1k0 : Retours de test: 
- le score d''inventaire enrichis a augmenté (de 380 à 403pt) alors que je me serais attendu à ce qu'il soit égale. Ce qu'on a ajouté est de l'impact lié à l'elec/chaleur. 
- je ne vois pas 'blanchiement' dans le champs prétraitement

PR #1285, opened 2025-07-23
chore: clean material.json

paulboosz:
#1280


public/data/textile/materials.json
```diff
@@ -1,235 +1,219 @@
...
     "origin": "NaturalFromAnimal",
     "primary": true,
     "processId": "40ec17ee-6b44-5ee2-8fbc-559aed52764f",
     "recycledFrom": null,
-    "recycledProcessUuid": null,
-    "shortName": "Laine nouvelle filière"
+    "recycledProcessId": null
   },
   {
+    "alias": "ei-coton",
```
n1k0, line 136: I don't get why we introduce this semantic alias, especially when the patch aims at avoiding having semantic identifiers; as you say in the description:

> Cette id sémantique peut poser problème si l'on change la matière sous-jacente et que la string ne représente plus le procédé sous jacent.

Shouldn't we only rely on uuids?

src/Data/Textile/Material.elm
```diff
@@ -105,24 +136,14 @@ encode : Material -> Encode.Value
...
-        , ( "recycledProcessUuid"
+        , ( "processId", Process.encodeId v.process.id )
+        , ( "recycledProcessId"
           , v.recycledProcess |> Maybe.map (.id >> Process.encodeId) |> Maybe.withDefault Encode.null
           )
-        , ( "recycledFrom", v.recycledFrom |> Maybe.map encodeId |> Maybe.withDefault Encode.null )
+        , ( "recycledFrom", v.recycledFrom |> Maybe.map Encode.string |> Maybe.withDefault Encode.null )
         , ( "geographicOrigin", Encode.string v.geographicOrigin )
         , ( "defaultCountry", v.defaultCountry |> Country.codeToString |> Encode.string )
         ]
```
n1k0, line 149: Nit: please order keys alphabetically 🙏

src/Data/Textile/Query.elm
```diff
@@ -56,6 +57,15 @@ type alias MaterialQuery =
     }
 
 
+materialWithId : Material.Id -> Split -> Maybe Spinning -> Maybe Country.Code -> MaterialQuery
+materialWithId id share spinning country =
+    {  country = country
```
n1k0, line 62: Nit: formatting; please ensure running `elm-format` before committing (note: we may want to automate this in a pre-commit hook but that should be another PR)

src/Data/Textile/Query.elm
```diff
@@ -448,12 +458,12 @@ default =
...
     , materials =
-        [ { country = Nothing
-          , id = Material.Id "ei-coton"
-          , share = Split.full
-          , spinning = Nothing
-          }
-        ]
+        case Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84" of
+            Err _ ->
+                []
```
n1k0, line 463: Please avoid silencing errors. This function should return a `Result` instead of an empty query, with an error when the uuid isn't found (which may happen at some point if we remove the matching material)

src/Page/Explore/TextileMaterials.elm
```diff
@@ -20,7 +20,7 @@ import Views.Icon as Icon
 import Views.Link as Link
 
 
-recycledToString : Maybe Material.Id -> String
+recycledToString : Maybe String -> String
 recycledToString maybeMaterialID =
     maybeMaterialID
         |> Maybe.map (always "oui")
```
n1k0, line 26: This should rather take an actual Uuid

src/Page/Textile.elm
```diff
@@ -1162,8 +1162,7 @@ view session model =
                                 , onAutocompleteSelect = OnAutocompleteSelect
                                 , placeholderText = "tapez ici le nom de la matière première pour la rechercher"
                                 , title = "Sélectionnez une matière première"
-                                , toLabel = .shortName
-                                , toCategory = .origin >> Origin.toLabel
+                                , toLabel = .name                                , toCategory = .origin >> Origin.toLabel
```
n1k0, line 1165: formatting; see previous remark about that

tests/Data/Textile/InputsTest.elm
```diff
@@ -69,9 +78,18 @@ suite =
...
                 , { tShirtCotonFrance
                     | materials =
-                        [ { id = Material.Id "ei-coton", share = Split.half, spinning = Nothing, country = Nothing }
-                        , { id = Material.Id "ei-pp", share = Split.half, spinning = Nothing, country = Nothing }
-                        ]
+                        case
+                            ( Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84"
+                            , Material.idFromString "73ef624d-250e-4a9a-af5d-43505b21b527"
+                            )
+                        of
```
n1k0, line 85: As we now have results, we should use the TestUtils suiteFromResult* test helpers

tests/Data/Textile/QueryTest.elm
```diff
@@ -16,12 +16,12 @@ sampleQuery : Query
...
-              , spinning = Nothing
-              , country = Just (Country.Code "CN")
-              }
-            ]
+            case Material.idFromString "9dba0e95-0c35-4f8b-9267-62ddf47d4984" of
+                Ok id ->
+                    [ materialWithId id Split.full Nothing (Just (Country.Code "CN")) ]
+
+                Err _ ->
+                    []
```
n1k0, line 24: We should return a `Result` instead of relying on a possibly invalid query

tests/Data/Textile/QueryTest.elm
```diff
@@ -72,46 +72,39 @@ suite =
...
-                  ]
+                , (case ( Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84", Material.idFromString "73ef624d-250e-4a9a-af5d-43505b21b527" ) of
+                    ( Ok cottonId, Ok syntheticId ) ->
+                        [ materialWithId cottonId Split.half Nothing Nothing
+                        , materialWithId syntheticId Split.half Nothing Nothing
+                        ]
+
+                    _ ->
+                        []
+                  )
```
n1k0, line 93: Same, use `TestUtils.suiteFromResult2`

tests/Data/Textile/QueryTest.elm
```diff
@@ -72,46 +72,39 @@ suite =
...
+                  (case ( Material.idFromString "9dba0e95-0c35-4f8b-9267-62ddf47d4984", Material.idFromString "73ef624d-250e-4a9a-af5d-43505b21b527", Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84" ) of
+                    ( Ok polyesterId, Ok polypropyleneId, Ok cottonId ) ->
+                        [ materialWithId polyesterId Split.sixty Nothing Nothing
+                        , materialWithId polypropyleneId Split.thirty Nothing Nothing
+                        , materialWithId cottonId Split.tenth Nothing Nothing
+                        ]
+
+                    _ ->
+                        []
+                  )
```
n1k0, line 107: Ditto

tests/Server/RouteTest.elm
```diff
@@ -153,12 +153,12 @@ textileEndpoints db =
...
-                      , share = Split.full
-                      , spinning = Nothing
-                      }
-                    ]
+                    case Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84" of
+                        Ok decodedId ->
+                            [ materialWithId decodedId Split.full Nothing (Just (Country.Code "invalid")) ]
+
+                        Err _ ->
+                            []
```
n1k0, line 161: Ditto

tests/Server/RouteTest.elm
```diff
@@ -175,25 +175,27 @@ textileEndpoints db =
...
-                      }
-                    ]
+                    case Material.idFromString
+
+                    "1c686e00-6db8-469e-8d7f-3864bd3238bd" of
+                        Ok decodedId ->
+                            [ materialWithId decodedId Split.full Nothing Nothing ]
+
+                        Err _ ->
+                            []
```
n1k0, line 185: Ditto, plus formatting issues

tests/Server/RouteTest.elm
```diff
@@ -175,25 +175,27 @@ textileEndpoints db =
...
-                      , share = Split.full
-                      , spinning = Nothing
-                      }
-                    ]
+                    case Material.idFromString "62a4d6fb-3276-4ba5-93a3-889ecd3bff84" of
+                        Ok decodedId ->
+                            [ materialWithId decodedId Split.full Nothing (Just (Country.Code "NotACountryCode")) ]
+
+                        Err _ ->
+                            []
```
n1k0, line 198: Same same

n1k0 (changes requested):
The change makes sense though implementation must be improved and strengthened

tests/Server/RouteTest.elm
paulboosz: done ✅

tests/Data/Textile/QueryTest.elm
paulboosz: done ✅

tests/Data/Textile/QueryTest.elm
paulboosz: done ✅

tests/Data/Textile/InputsTest.elm
paulboosz: done ✅

paulboosz (commented):
first batch of modifs following review

public/data/textile/materials.json
paulboosz: A refactoring I didn't want to do right now 😆

But even if we do this refactoring, I find there is still an use case for `alias` :
- uuid are obviously needed
- alias are also useful : when we communicate between humans we would intuitively use alias and say 
"Modify the cff manufacturer allocation to 0.7. on `coton-rdp`"  rather than 
"Modify the cff manufacturer allocation to 0.7. on `d3d20384-88fb-4c75-ac12-c647365ac6c5`"

Also as we merge `ingredients` and `materials` in the same `metadata` object, ingredients have an `alias` that is really useful to communicate. 

We use "bellpepper-unheated-greenhouse-fr" instead of "361139ac-1028-4cc8-9c0e-4955c5fa39d3"

n1k0:
All the changes look good, I think we can safely land this patch!

paulboosz:
> All the changes look good, I think we can safely land this patch!

Great thanks !
Just waiting on validation from @vjousse  or @cedricr on corresponding ecobalyse-data patch : https://github.com/MTES-MCT/ecobalyse-data/pull/121

PR #1289, opened 2025-07-25
refactor: let invalid cooking process be detected as a compilation error

ccomb:
## :wrench: Problem

During migration to AGB 3.2 a test failure revealed a cooking process hidden as a uuid in the code. That uuid needed to be changed but that did not trigger a failure but a minor change in test results that could have gone unseen and was not obvious.

## :cake: Solution

- Move that cooking process to the wellKnown processes so that it's a needed process of the food sector.
- Remove the wildcard case (`_ ->`) so that a changed or missing cooking triggers a compilation error.

Just run the tests

### DIFF
--- a/src/Data/Food/Recipe.elm
+++ b/src/Data/Food/Recipe.elm
@@ -122,7 +122,7 @@ availablePackagings usedProcesses =
 
 
 compute : Db -> Query -> Result String ( Recipe, Results )
-compute db =
+compute ({ food } as db) =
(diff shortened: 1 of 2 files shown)


vjousse (approved):
Looks fine to me 👍
Just a little nit: I would have replaced `db.food.wellKnown` with `food.wellKnown` everywhere by pattern matching it in the function definition for example:
```elm
stepListView ({ food } as db) session { impact, initialQuery } recipe results =
```

PR #1293, opened 2025-07-31
feat: list and parse scalingo logs

vjousse:
## :wrench: Problem

It’s hard to get information about logs on scalingo as they are archived in a non friendly interface.

## :cake: Solution

Create a script that can download the logs and try to parse some information out of it.

I’m using python mutliprocessing to speed up the log parsing as it takes quite some time without it.

```python
uv run bin/scalingo_api.py log-archives
```

Should display the slowest log lines, and logs should have been downloaded in `./archives_logs` directory.


paulboosz (approved):
Works fine ! I had to add a SCALINGO_API_TOKEN in .env
Maybe add it to .env.sample ?

PR #1306, opened 2025-08-19
feat(bo): add processes admin ui

n1k0:
## :wrench: Problem

We can't list processes in the back-office.

## :cake: Solution

Add a processes admin page listing them and allow searching and filtering them by scope(s)

Browse the new process admin page, and play around with the search field and the scope filter form.

Note that the components admin is now searchable as well.


vjousse (approved):
The implementation looks good to me.
I have performance issues (search box is laggy) using FF and chrome but it’s still usable and acceptable.
Let’s optimize further later on if needed, good job!

PR #1341, opened 2025-09-02
feat: add elements model

### LINKED ISSUE #1338: Contrôler que les id des composants utilisent des id de procédé existant
### Probleme

Première tache avant que Nico regarde ce que ça change que de consommer la bdd plutôt que des fichiers statiques

vjousse:
## :wrench: Problem

fixes #1338 

## :cake: Solution

Add a new `Element` model instead of the json field.

1. Setup database from `master` and test the API

```
git checkout master
./bin/reset-docker-db.sh
npm run start:dev
http http://localhost:1234/backend/api/components
```

You should have components with elements in them. Stop the dev server.

2. Switch to the current branch, update the DB and test the api

```
git checkout chore/add-integrity-constraints
uv run backend database upgrade --no-prompt
npm run start:dev
http http://localhost:1234/backend/api/components
```

You should have components without elements (as the new tables are empty). Stop the dev server.

3. Migrate the data

```
uv run backend data migrate-elements
```

4. Test the api

```
npm run start:dev
http http://localhost:1234/backend/api/components
```

You should have components with elements in them. Stop the dev server.

Then, all the changes should be transparent from an API consumer perspective. So, just test that everything works as before for the components.

- [x] Elements w/o transforms are not updated
- [x] Scopes are not updated
- [x] Data migration of the `element` values on staging

### DIFF
--- a/backend/src/app/cli/commands.py
+++ b/backend/src/app/cli/commands.py
@@ -412,4 +412,5 @@ async def _migrate_elements(db_session) -> None:
                         "owner_id": user.id,
                         "elements": component.elements_json,
                     },
+                    auto_commit=True,
                 )


cedricr:
@vjousse I’m still looking into it, but while testing the changes I have to following problems :

1) starting from an new, empty database, loading the fixtures will fail

```shell
uv run backend database upgrade --no-prompt
uv run backend users create-default-user
uv run backend fixtures load-components public/data/object/components.json

(...)
  File "/home/cedricr/eb/ecobalyse/backend/src/app/domain/components/services.py", line 47, in to_model_on_create
    return await self._populate_with_journaling(data, "create")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cedricr/eb/ecobalyse/backend/src/app/domain/components/services.py", line 200, in _populate_with_journaling
    element.to_dict(), operation="create"
    ^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'to_dict'
```
2) starting from a new, empty database, and loading the fixtures, *then* switching to this branch and running the migration with `uv run backend database upgrade --no-prompt`, the migration will run without error, the new table is created as expected, the `elements` jsonb field in the `component` table disappear as expected, but no data is migrated (the `element` table stay empty). This might be as planned, relying on updated fixtures, though ?

3) at this point, going the the admin interface and editing a component by adding an element, I see the green “Composant sauvegardé“ toast message, but if I re-edit the same component, the newly added element is not there, and neither does it appear in the database.

backend/src/app/domain/components/services.py
```diff
@@ -169,160 +169,112 @@ async def delete_many(
...
+        ):
+            with self.repository.session.no_autoflush:
+                return await self._create_component(data, processes_service, owner)
+
+        if (
+            operation == "update"
+            and is_dict(data)
+            or (operation == "upsert" and has_id)
+        ):
+            with self.repository.session.no_autoflush:
```
cedricr, line 277: I don’t remember for sure right now, but I think it’s the no_autoflush is necessary for the create, but useless for the update.

backend/src/app/domain/components/services.py
```diff
@@ -169,160 +169,112 @@ async def delete_many(
...
         )
 
+    async def _create_element(
+        self,
+        element: ModelDictT[ComponentElement],
+        component_id: str,
+        processes_service,
+    ):
+        element_dict = (
+            element.to_dict() if type(element) is ComponentElement else element
```
cedricr, line 179: It would probably be better to use `isinstance`, but I’d rather do that upstream if possible anyway, so that we can be more assertive on the type entering this method.

backend/src/app/domain/components/services.py
```diff
@@ -169,160 +169,112 @@ async def delete_many(
...
+        )
+        tranforms_ids = element_dict.pop("transforms", None)
+
+        element_dict["material_id"] = element_dict.pop("material")
```
cedricr, line 183: we’ll need a default argument in the case there’s not material

backend/src/app/cli/commands.py
```diff
@@ -356,3 +358,43 @@ async def _load_processes_json(components_data) -> None:
...
+)
+@click.pass_context
+def data_management_group(_: dict[str, Any]) -> None:
+    """Manage data migrations."""
+
+
+@data_management_group.command(
+    name="migrate-elements", help="Migrate existing elements."
+)
+def migrate_elements() -> None:
```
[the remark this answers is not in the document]
vjousse: This should be removed later on when the migration script will be executed. I’ve added related tests too!

PR #1352, opened 2025-09-08
fix(object): apply transforms preserving material unit

n1k0:
refs https://github.com/MTES-MCT/ecobalyse/issues/1223


n1k0:
@camcoq-start a fix is being deployed to the branch 👍

PR #1355, opened 2025-09-10
docs: update the installation instructions

cedricr:
## :wrench: Problem

Attempt to update and simplify the setup instructions.

Fixes https://github.com/MTES-MCT/ecobalyse/issues/1499

## :cake: Solution

Adding a dependency to [dotenvx](https://github.com/dotenvx/dotenvx) might simplify the management of the envvars. For example, instead of having to manually `export TRANSCRYPT_KEY=xxx ; ./bin/run-transcrypt.sh`, the postinstall step could work automatically. 

I’ve added some comments in the impacted code to open a discussion. They will have to be removed before squashing the PR.


README.md
```diff
@@ -74,21 +78,70 @@ Note: docker est également une dépendance requise pour lancer la suite de test
...
+uv run backend fixtures load-components public/data/object/components.json
+```
+
+### Créer un super-utilisateur
+
+```bash
+uv run backend users create-user --email foo@bar.org --first-name Foo --last-name Bar --superuser
+```
+
+### Utilser la ligne de commandes `backend`
```
cedricr: Utiliser

cedricr:
Need to add `?ssl=prefer` to the database URLs in the URL; otherwise some scripts won’t work (related to https://github.com/MTES-MCT/ecobalyse/pull/1852).
Switching to draft so that it doesn’t get merged without that.

Issue #1334, opened 2025-08-28
Add a "comment" field to components


nicolaspkandeel:
## :wrench: Problem

It may be important to provide more information about a component than the name and the composition, so that the user knows what component should be used in a calculation.

Examples : zip lenght in Textile : the range of applicable zip could be provided
Exact reference of wheel is usefull for vehicule.
The corresponding size of glazzing is usefull the use the correct glass for vehicles.

## :cake: Solution

Adding  a "comment" field to components

## :desert_island: How to test

- [ ] A field named "comment" is available for every component in the admin interface, in which a sentence can be written.
- [ ] The comment is visible in the component explorer

PR #1371, opened 2025-09-16
feat(object,ui): add support for component comments

n1k0:
## :wrench: Problem

see #1334 

## :cake: Solution

This patch adds support for a new `comment` component field in Elm land and in the back-office.

- [x] Elm can decode an optional comment field (fallbacks to `""` when doesn't previously exists)
- [x] Explorer can list component comments
- [x] Backend exposes a new `comment` JSON property in component api responses
- [x] Back-office renders that field
- [x] Back-office allows updating a component comment


vjousse (approved):
Tested locally 👍

PR #1380, opened 2025-09-17
fix: fixed mass validation message

ccomb:
Fixed the error message found in the API when we set a zero mass


n1k0 (commented):
Failing test labels should be updated as well

Issue #1413, opened 2025-09-29
Integrate new processes for Vehicules


nicolaspkandeel:
## :wrench: Problem

Several processes are needed in Ecobalyse for Vehicules : 
- batteries
- fuels
- steel and aluminium with recycling and market options
- electronics
## :cake: Solution

Integrate new processes : 

Nom | Nom technique | Source | Catégories | Unité | Densité
-- | -- | -- | -- | -- | --
Composants électroniques | electronic component, passive, unspecified//[GLO] market for electronic component, passive, unspecified | Ecoinvent 3.9.1 | Matériau | kg | 0
Batterie Li-ion (NMC811) | Battery cell, Li-ion, NMC811 {CN}\| battery cell production,   Li-ion, NMC811 \| Cut-off, U | Ecoinvent 3.11 | Matériau | kg | 6.7
Batterie   Li-ion (LFP) | Battery, Li-ion, LFP, rechargeable {CN}\| battery production, Li-ion, LFP,   rechargeable \| Cut-off, U | Ecoinvent 3.11 | Matériau | kg | 8.6
Batterie   Li-ion (Autres) | Battery, Li-ion, NCA, rechargeable {CN}\| battery production, Li-ion, NCA,   rechargeable \| Cut-off, U | Ecoinvent 3.11 | Matériau | kg | 6.3
Batterie Plomb | Battery, lead acid, rechargeable, stationary {RoW}\| battery production,   lead acid, rechargeable, stationary \| Cut-off, U | Ecoinvent 3.11 | Matériau | kg | 33.3
Traitement de batterie en fin de vie | Used Li-ion battery {GLO}\| market for used Li-ion battery \| Cut-off, U | Ecoinvent 3.11 | Matériau | kg | 0
Diesel (B7), norme Euro6 | diesel, B7, EURO6 | Ecobalyse | Energie++ | L | 0
Diesel (B7), norme Euro4 | diesel, B7, EURO4 | Ecobalyse | Energie++ | L | 0
Essence (E10), norme Euro6 | petrol, E10, EURO6 | Ecobalyse | Energie++ | L | 0
Superéthanol (E85), norme   Euro6 | petrol, E85, EURO6 | Ecobalyse | Energie++ | L | 0
Aluminium (recyclé) | aluminium, cast alloy//[RER] treatment of aluminium scrap, post-consumer,   prepared for recycling, at refiner | Ecoinvent 3.9.1 | Matériau | kg | 0
[...]

PR #1419, opened 2025-10-08
feat: add vehicle processes

paulboosz:
## :wrench: Problem

#1413 
## :cake: Solution


cedricr (approved):
lgtm !

PR #1445, opened 2025-10-14
chore: add Sentry integration to the backend

### LINKED ISSUE #1429: Setup Sentry in the Python backend
## :wrench: Problem

The Python backend is not reporting errors to Sentry

Objectif : mieux suivre les erreurs/bugs

cedricr:
## :wrench: Problem

Fixes #1429

## :cake: Solution

Integrate Sentry using the [Litestar specific instructions](https://docs.sentry.io/platforms/python/integrations/litestar/)

- I’ve added an “heath check” url to test the integration. Do we want to keep it after testing or not ? My opinion is that it’s always useful to have it when from time to time we might have doubts on Sentry’s heath. That being said, being a backend specific URL I’m not sure it can be called directly in production?
- I’ve added a “subsystem” tag, so that we can differenciate between the errors reported by the front-end / express / backend. I’ve not tested that it works correctly for the javascript part – maybe we should add a specific “health check” there too.

Login locally using an admin account. Then `http://localhost:1234/backend/check-sentry` should return a 500-error, and log a ZeroDivisionError on Sentry.

- [x] The error should be marked as "development"
- [x] It should have the proper release hash
- [x] It should have a `subsystem` tag with a value of `back-end`

In production/review-apps, login using an admin account. Then `https://ecobalyse-staging-pr1445.osc-fr1.scalingo.io/backend/check-sentry` should return a 500-error, and log a ZeroDivisionError on Sentry.
- [x] Errors should be marked as "production", or "review-app"
- [x] They should have the proper release hash
- [x] They should have a `subsystem` tag with a value of `back-end`

### DIFF
--- a/backend/src/app/config/app.py
+++ b/backend/src/app/config/app.py
@@ -49,7 +49,7 @@ settings = get_settings()
 
 sentry_dsn = settings.app.SENTRY_DSN
 
-if sentry_dsn and not settings.is_test_env:
+if sentry_dsn and not settings.is_test_env():
     sentry_sdk.init(
         dsn=sentry_dsn,
         environment=settings.app.SENTRY_ENVIRONMENT,
--- a/backend/src/app/config/base.py
+++ b/backend/src/app/config/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import binascii
 import json
 import os
+import sys
 from dataclasses import dataclass, field
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


n1k0 (commented):
Works flawlessly, good job!

> I’ve added an “heath check” url to test the integration. Do we want to keep it after testing or not ?

If it's public I'm reluctant to allow this url to crash the server… Up to you folks

PR #1505, opened 2025-10-29
fix: resynchronize the processes files with the ecobalyse-data repository

cedricr:
## :wrench: Problem

The processes.json and processes_impact.json files were not identical with the ones generated by running `npm run export:all` on the ecobalyse-data repository.

## :cake: Solution

Given that the differences are only in comment fields, we just overwrite the files with the generated ones.


paulboosz (approved):
I think I'm responsible for this, thank you for the fix !

PR #1506, opened 2025-10-29
feat: allow renaming bookmarks

### LINKED ISSUE #1473: Renommer une simulation sauvegardée
### Problème

Une fois une simulation sauvegardée, on ne peut pas la renommer

### Solution

Permettre à un utilisateur qui a sauvegardé une simulation de pouvoir la renommer.

vjousse:
## :wrench: Problem

fixes #1473 

## :cake: Solution

### DIFF
--- a/src/Data/Session.elm
+++ b/src/Data/Session.elm
@@ -44,6 +44,7 @@ import Data.User as User2
 import Json.Decode as Decode exposing (Decoder)
 import Json.Decode.Pipeline as JDP
 import Json.Encode as Encode
+import List.Extra as LE
 import RemoteData exposing (WebData)
 import Request.BackendHttp.Error as BackendError
 import Request.Version exposing (Version)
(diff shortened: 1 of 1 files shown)


src/Views/Bookmark.elm
```diff
@@ -291,25 +294,68 @@ bookmarkView cfg ({ name, query, version } as bookmark) =
...
-                |> (++) cfg.session.clientUrl
-                |> href
-            ]
-            [ text name
-            ]
+        , case ( beingRenamed, cfg.bookmarkBeingRenamed ) of
+            ( True, Just renamedBookmark ) ->
+                input
+                    [ type_ "text"
+                    , class "form-control"
```
n1k0: I suggest adding the `form-control-sm` to avoid the line to flicker when switching from label to text input

src/Views/Bookmark.elm
```diff
@@ -291,25 +294,68 @@ bookmarkView cfg ({ name, query, version } as bookmark) =
...
+                        |> (++) cfg.session.clientUrl
+                        |> href
+                    ]
+                    [ text name
+                    ]
+        , if beingRenamed then
+            button
+                [ type_ "submit"
+                , class "btn btn-sm btn-success"
+                , title "Sauvegarder la simulation dans le stockage local au navigateur"
```
n1k0: nit: s/local au/local du

PR #1507, opened 2025-10-30
chore: add location and rename sourceId->activityName

paulboosz:
[ecobalyse-data PR](https://github.com/MTES-MCT/ecobalyse-data/pull/146)

## :wrench: Problem

Fix :
- https://github.com/MTES-MCT/ecobalyse/issues/1483
- https://github.com/MTES-MCT/ecobalyse/issues/1475

## :cake: Solution

- Ajouter le champs `location` pour permettre d'identifier une ICV (voir [#1483](https://github.com/MTES-MCT/ecobalyse/issues/1483)
- Afficher ce champs dans l'explorateur
- Renommage : `sourceId` -> `activityName`

- Renommage : `sourceId` -> `activityName`

- tests should pass

### DIFF
--- a/backend/src/app/domain/processes/schemas.py
+++ b/backend/src/app/domain/processes/schemas.py
@@ -64,7 +64,6 @@ class Process(CamelizedBaseStruct):
 
     categories: list[Category]
     comment: str
-    display_name: Optional[str]
     id: UUID
     impacts: Impacts
     source: str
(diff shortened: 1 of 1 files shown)


tests/Data/ComponentTest.elm
```diff
@@ -1100,39 +1102,40 @@ sawing : Result String Process
 sawing =
     decodeJson (Process.decode Impact.decodeImpacts) <|
         """ {
+                "activityName": "Sawing + kiln drying in Europe (wood), constructed by Ecobalyse",
                 "categories": ["transformation", "material_type:wood"],
-                "comment": "added by Ecobalyse",
```
n1k0: why is this removed and how is it related to the purpose àof this patch?

cedricr:
> I'd like a review from @vjousse or @cedricr because I'm surprised there's no db migration for the renamed field here

@paulboosz indeed, the data migration is missing but I’m not sure the process is documented. I could try adding it, but it would be better if it could wait for @vjousse.

Fun fact: the backend tests pass, we should ensure they don’t in that case. On the other hand, resetting the DB with ./bin/reset-docker.sh **do** crash.

vjousse:
It’s expected that the tests do pass: we don’t test the data migration. I think that for the moment the fact that the deploy is failing due to the missing data migration is enough.

vjousse (changes requested):
As far as I can tell, the `location` field is missing from the db model, it should be added and the API endpoints should be modified and tested accordingly. Don’t hesitate if you need help about that.

PR #1511, opened 2025-10-31
feat(object): compute impacts against localized energy mixes

n1k0:
## :wrench: Problem

See #1326:  users must be able to tweak the electricity and heat mixes used depending on the origin of their component

## :cake: Solution

Allow selecting a country for each product component. Expose energy mixes used at each transformation step in a tooltip.

Capture d’écran 2025-11-03 à 11 25 00

### Data model

I'm using the `countries.json` datafile, but don't rely on the `default country` (code `---`). I'm using the label `Monde` in the ui when the country isn't known, which then leverages the default energy mixes as before (`Électricité moyenne tension, Asie` and `Mix chaleur (Monde)`). Otherwise, the mixes attached to the "country" (or more accurately, the *region*) is used. This strategy makes the patch immediately useful, while refactoring the whole data pipeline to have a proper data modeling of countries and regions could take a while.

### ⚠️ Data issue with aggregated transformation processes

Most of the current transformation processes don't have any elec and heat consumption data attached, most probably because all the impacts (transformation plus elec and heat consumption impacts) are aggregated. Please pick the [`Tissu 2kg 50%coton 50%polyester (canapé 3p)` example](https://ecobalyse-staging-pr1511.osc-fr1.scalingo.io#/object/simulator/ecs/eyJjb21wb25lbnRzIjpbeyJpZCI6IjVmNzRjZGEyLTE5YTktNDdjNy05ZmQ3LTc1ODhjMjRhZGYyMSIsInF1YW50aXR5IjoxfV0sImR1cmFiaWxpdHkiOjF9), which is the only example leveraging textile transformation processes having non-zero elec and heat consumptions defined.

Toy around with the new country selector in the object simulator (read the section above for caveats).

### DIFF
--- a/src/Page/Object.elm
+++ b/src/Page/Object.elm
@@ -610,8 +610,6 @@ simulatorView session model =
                 , setDetailed = SetDetailedComponents
                 , title = "Production des composants"
                 , updateElementAmount = UpdateElementAmount
-
-                -- FIXME: implement
                 , updateItemCountry = UpdateComponentItemCountry
                 , updateItemName = UpdateComponentItemName
                 , updateItemQuantity = UpdateComponentItemQuantity


cedricr (approved):
Tests pass, UI works as advertized, and the code looks like it makes sense 😆 
Great job!

PR #1562, opened 2025-11-20
chore: stabilize process ids

paulboosz:
# Problem

#172 

While doing this I found a bug, 3 processes where not exported because they were thought to be duplicate

- Lamb, fresh meat, at slaughterhouse (WFLDB) location:NZ (visible:false)
- Potato, at farm (WFLDB) location:CN
- Cherry, at farm (WFLDB) location:GLO 

# Solution

- ids are in activities.json
- no process ids are generated
- tests are checking that there are no duplicate or missing ids in `activities.json` and `processes.json`

# How to test

npm run export:all should not create diff


paulboosz:
Avant (même impact car même icv sous-jacente) :
avant

Après (différenciation) :
apres

PR #1564, opened 2025-11-20
chore!: renaming `country` to `geoZone` and `zone` to `worldRegion`

### LINKED ISSUE #1532: Renommer le fichier countries.json en 'location' ?
### Problème 

On a des régions et des pays dans la table

### Solution 

- [ ] Renommer le nom du fichier: Geozone
- [ ] Remplacer 'Pays' dans l'explorateur par 'Zone géographique'
- [ ] Coté API, on passe l’ancien nom en deprecated, et on rajoute le nouveau
- [ ] On ne backporte pas à la version stable

cedricr:
## :wrench: Problem

Fixes #1532

See also: https://github.com/MTES-MCT/ecobalyse-data/pull/179


n1k0 (commented):
These are just preliminary remarks with the `geoZone` name, which could just be `geozone` and solves a lot of headaches wrt naming conventions across codebases. `worldRegion` could also be just `region` imho, but I won't die on this other hill. More broadly, even if I understand we can't use a same single concept and type for "any scale geographic location", I fear the two names — *geo zone* and *world region* — might be quite confusing to end-users and documentation readers… though I don't have much better suggestions to offer. What would prevent us to use a single concept and type? (I'm asking the question because I didn't spend the necessary time to investigate the matter yet am offering others to chime in, and I'm definitely part of the ones who should be involved answering that question)

PR #1572, opened 2025-11-24
feat: dump published components

### LINKED ISSUE #1390: Extraire statiquement les données de la base
## :wrench: Problem

Nous avons besoin, lorsque l’on mergera une PR sur `master`, de pouvoir produire un export du contenu de production de la base de données dans des fichiers statique JSON. Le but est de garder le code avec les données pour n’importe quel commit dans `master`.

uniquement le script + export en fichier statique via git 

- [ ] Example expectation
- [ ] Another example expectation

vjousse:
## :wrench: Problem

fixes #1390 

## :cake: Solution

Add a command to export published components, and add a workflow dispatch on GH actions to test the automatic generation of the file.

Currently, the workflow is activated on every push here, I can/will remove it before merging so that it’s only triggered by hand.

Push an empty commit here, or try `uv run backend json dump-components` locally with some components marked as `published` in the admin.

### DIFF
--- a/.github/workflows/export_json.yml
+++ b/.github/workflows/export_json.yml
@@ -2,7 +2,6 @@ name: Export static JSON
 
 on:
   workflow_dispatch:
-  push:
 
 jobs:
   export_json:


.github/workflows/export_json.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+          python-version: ${{ matrix.python-version }}
+
+      - name: Add scalingo to know hosts
+        run: echo "KNOWN_HOSTS=$(ssh-keyscan -H ssh.osc-fr1.scalingo.com)" >> $GITHUB_ENV
+
+      - name: Install SSH key
+        uses: shimataro/ssh-key-action@v2
```
cedricr: Is that too complex to do with a pair of echoes and redirections ? Just shuddering when I image that repo being compromised 😅

cedricr (approved):
I’ve written a couple of completely ignorable suggestions 😆.

Apart from that, it works perfectly from the command-line, and I’ve forced-run the action, which works without errors (I’ve got no idea where to find the artifacts though !)
That’s niiiiice !

vjousse:
> I’ve written a couple of completely ignorable suggestions 😆.
> 
> Apart from that, it works perfectly from the command-line, and I’ve forced-run the action, which works without errors (I’ve got no idea where to find the artifacts though !) That’s niiiiice !

Well, I get all you concerns… but I don’t want to spend more time on this one. The day we want to address those valid concerns, we should not forget to do it for `score_history` too.

Issue #1609, opened 2025-12-03
Refactor Data.Component.Query to prevent an impossible state


n1k0:
### Problem

Right now, we can modelize a Component query in two "impossible" states:

- `{ assemblyCountry = Just Country.Code "PT", items = [] }` No items at all but an assembly country provided (note: we might want to keep it just in order to present the user with an empty product editor ui featuring no items at all)
- `{ assemblyCountry = Just Country.Code "PT", items = [{"id": "xxx"}] }`: this one is much more annoying as it contradicts itself; if we only have one and just one item in the product, it should never be assembled, so why would we accept specifying a country of assembly?

An actually valid state is this one:

- `{ assemblyCountry = Just Country.Code "PT", items = [{"id": "xxx"}, {"id": "yyy"}] }`

### Solution

We may want to refactor the Query so we make this dangerous impossible state actually impossible to represent in code:

```elm
type Query
    = SingleItem Unit.Ratio Component.Id
    | MultipleItems Unit.Ratio (Maybe Country.Code) (List Component.Id)
```

```elm
case query of
    SingleItem durability componentId -> 
        -- implement single item query processing, there **can't** be an assembly country

    MultipleItems durability maybeAssemblyCountry componentIds -> 
        -- implement multiple items query processing, with an optional assembly country
```

PR #1610, opened 2025-12-03
fix(object): do not apply transport to assembly for a single item

n1k0:
## 🤔 Problem

When a product is modeled featuring a *single* component, transport impacts are currently computed for shipping it to an assembly country, then to a distribution one. This makes no sense: if there's a single component, no need to assemble anything, so final transport impacts should only be from production country to distribution.

## 💡 Solution

Avoid computing transport impacts to an assembly step when the product features a single component (the step is already visually hidden, but the impacts were added to total score).

## ⚠️ Remarks & comments

This patch also fixes other bugs I discovered while investigating the initial issue, and refactors a few things along the way:

### Erroneous component item transport rendered data

This patch also fixes a rendering bug initially reported by @albanfournier [on Mattermost](https://mattermost.incubateur.net/fabnum-mte/pl/b9djwrjrdirmum9zjses4j6dor):

> Je ne me concentre pas dessus mais pour info voici des info/impacts qui ne semblent pas cohérents.
>
> Je suis arrivé sur cet onglet en cliquant sur le module "Utiliser dans le simulateur" du composant "Plateau Chene vernis" dans l'itnerface Admin 

image

### Avoid implementing item transports computation logic twice

The patch introduces a new `Data.Component.computeItemTransportToAssembly` fn to compute transports for an item to its assembly country. This function is called from the simulator and the detailed HTML view, so there's now a single place where to maintain it :)

Also fixes the case where transport of a single item was still taking the assembly step as per comment in https://github.com/MTES-MCT/ecobalyse/pull/1610#pullrequestreview-3535864352 

### Component config loading race condition

I also discovered that while we initialize the app with the default component configuration from `Data.Component.Config.default`, the time the real one loaded over HTTP arrives, we have scores already computed, and they're not refreshed. This is a classic race condition issue, which is solved in this patch by ensuring reprocessing the initially requested route when the component config is received, so the computation is performed again using these new data.

### View code simplification

The patch simplifies view code, most notably by decreasing cyclomatic complexity and streamlining the way we handle custom component names.

### Further enhancements

We found with @vjousse that we could refine the types modelizing a Query to avoid impossible state, and filed an issue about that in #1609, which is not addressed by this patch.

### DIFF
--- a/public/data/components/config.json
+++ b/public/data/components/config.json
@@ -1,4 +1,7 @@
 {
+  "distribution": {
+    "country": "FR"
+  },
   "endOfLife": {
     "scopeCollectionRates": {
       "object": 70
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -707,14 +707,14 @@ computeDistributionTransports { config, db } maybeAssemblyCountry items =
                 |> Maybe.map
                     (\{ code } ->
                         db.distances
-                            |> Transport.getTransportBetween Impact.empty code (Country.Code "FR")
+                            |> Transport.getTransportBetween Impact.empty code config.distribution.country.code
                     )
                 |> Maybe.withDefault config.transports.defaultDistance
 
         -- Many items, assembly country specified
         ( _, Just { code } ) ->
             db.distances
-                |> Transport.getTransportBetween Impact.empty code (Country.Code "FR")
+                |> Transport.getTransportBetween Impact.empty code config.distribution.country.code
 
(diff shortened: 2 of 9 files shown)


camcoq-start (changes requested):
It is ok for product initially modeled with a single item.
There is a bug for product which become single items product.

Exemple: 
1. i chose Germany as assembly country for sofa exemple (ex: germany)
2. i delete components to finally have a single item product
3. transport seems to take in account an assembly country

image

PR #1615, opened 2025-12-04
feat: add packaging from ctcpa

paulboosz:
https://github.com/MTES-MCT/ecobalyse/issues/1305
 
- adding a few packaging solutions to begin with
- fixing an unrelated issue regarding how we access the process comment :
 Before we used `activity["comment"]` of the production exchange but `activity["Comment"]` of the activity is more detailed 

- change unit `Item(s)` to `item`

To test db migration : 
- `uv run backend database upgrade --no-prompt`
- `docker exec -it ecobalyse_dev_db psql -U ecobalyse -d ecobalyse_dev -c "SELECT enumlabel FROM pg_enum JOIN pg_type ON pg_enum.enumtypid = pg_type.oid WHERE pg_type.typname = 'unit';"`

should return `item` instead of "Item(s)"

- `uv run backend database downgrade --no-prompt`
- `docker ...`
should show `Item(s)` again


paulboosz:
> The migrations are conflicting

should be ok now

PR #1634, opened 2025-12-10
chore: serialize the JournalEntries values as string

### LINKED ISSUE #1585: Erreur en accédant à l'historique des modifications d'un composant
:wrench: Problème

En essayant d'accéder à l'historique de modification d'un composant, par ex `Assise plastique Chine` (`8209c191-722d-4074-b75a-e26a62cecabe`), on obtient une erreur de décodage JSON qui ne contient pas/plus les scopes :

Image

cedricr:
## :wrench: Problem

Fix #1585

## :cake: Solution

This is the backend fix proposed in #1585. The backend will now serialize the JournalEntries values as formatted string, instead of json objects. 

Obviously, this breaks the front-end in its current state. This should not be merged without the front-end fixes.

### DIFF
--- a/backend/src/app/db/models/journal_entry.py
+++ b/backend/src/app/db/models/journal_entry.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-from sqlalchemy.util.topological import sort
 
 import datetime
 import json


backend/src/app/db/models/journal_entry.py
```diff
@@ -53,3 +55,10 @@ class JournalEntry(
     user: Mapped[User] = relationship(
         back_populates="journal_entries", innerjoin=True, lazy="joined"
     )
+
+    @property
+    def value_str(self) -> str | None:
+        if self.value is not None:
+            # Used for json/msgspec serialization
+            return json.dumps(self.value, ensure_ascii=False, indent=2, sort_keys=True)
+        return self.value
```
vjousse:
```suggestion
```
This should not be needed?

n1k0 (approved):

PR #1670, opened 2025-12-23
feat(data): new ingredients

ccomb:
## :wrench: Problem

Many missing ingredients

## :cake: Solution

Added ingredients from https://github.com/MTES-MCT/ecobalyse-data/pull/196 for the FR variant
Note that the old ingredients are renamed with a `-2025` suffix and will be removed later.

Check the new ingredients


public/data/food/ingredients.json
GLestra15: Baker's yeast Ingrédient supprimé, non agricole brut

PR #1698, opened 2026-01-07
chore: remove unused and obsolete npm dependencies

n1k0:
Investigating our use of some npm dependencies we declare using in `package.json`, I discovered a few obsolete and unused ones, so this patch removes them


vjousse (approved):
❤️

PR #1700, opened 2026-01-08
fix: preserve transformations ordering

### LINKED ISSUE #1478: Changements non souhaité de l'ordre des transformations dans les composants
:wrench: Problème

Lors de la création d'un composant avec un élément acier ou alu subissant deux transformations (lamination + transformation finale), l'ordre des transformations est modifié automatiquement.
Cela ne change pas significativement le résultat, mais affiche un ordre des procédés de transformation qui ne correspond pas à la réalité métier.

:cake: Solution

Ajouter un champs position dans les procédés de transformation.

:rotating_light: Points d'attention/commentaires

:desert_island: Comment tester

[ ] Exemple d'attente
[ ] Autre exemple d'attente

cedricr:
## :wrench: Problem

When creating a component in the BO, the ordering of the transforms is not preserved.

Fix #1478

## :cake: Solution

We implemented the [Association Object](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#association-object) pattern, in order to add a `position` field to the transforms.

We could have gone further and made use of the [Association Proxy](https://docs.sqlalchemy.org/en/20/orm/extensions/associationproxy.html) extension to make it more transparent, and probably of the [ordering list](https://docs.sqlalchemy.org/en/20/orm/extensions/orderinglist.html) extension, but there’s already enough magic involved in SQLAlchemy as is :sweat_smile: 

The existing components should probably be checked to ensure the current order is correct, and updated otherwise.

- [ ] Tests should pass
- [ ] Creating a component in the BO with two transforms in the order A, B should preserve the order on save
- [ ] Creating a component in the BO with the same transforms in the order B, A should preserve the order on save

### DIFF
--- a/backend/src/app/domain/components/services.py
+++ b/backend/src/app/domain/components/services.py
@@ -190,11 +190,13 @@ class ComponentService(SQLAlchemyAsyncRepositoryService[m.Component]):
         transforms: list[m.ProcessElementTransform] = []
 
         if len(transforms_ids):
+            processes = await processes_service.list(m.Process.id.in_(transforms_ids))
             for idx, tid in enumerate(transforms_ids):
                 # See https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#association-object
                 elementTransform = m.ProcessElementTransform(position=idx)
-                process = await processes_service.get(tid)
-                elementTransform.transform = process
+                elementTransform.transform = next(
+                    (p for p in processes if str(p.to_dict()["id"]) == str(tid))
+                )
                 transforms.append(elementTransform)
 
             elt.process_transforms.extend(transforms)


backend/src/app/domain/components/services.py
```diff
@@ -180,17 +180,24 @@ async def _create_element(
...
-            elt.process_transforms.extend(
-                await processes_service.list(m.Process.id.in_(tranforms_ids))
-            )
+        transforms: list[m.ProcessElementTransform] = []
+
+        if len(transforms_ids):
+            for idx, tid in enumerate(transforms_ids):
+                # See https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#association-object
+                elementTransform = m.ProcessElementTransform(position=idx)
+                process = await processes_service.get(tid)
```
vjousse: Shouldn’t we keep the `processes_service.list(m.Process.id.in_(tranforms_ids))` call and search the processes ids in it to avoid doing as much SQL queries than the number of transforms that we have?

vjousse (approved):
LGTM with a litte nit. Good job!

vjousse (approved):
LGTM, great work!

PR #1704, opened 2026-01-09
fix(ui): fix incomplete comparator chart legend when exported

n1k0:
## :wrench: Problem

[thread](https://mattermost.incubateur.net/fabnum-mte/pl/4moh4fw6min4iqe6fgnurgdcce)

problem

## :cake: Solution

solution


cedricr (approved):
LGTM !

PR #1712, opened 2026-01-14
fix(bo,ui): constrain comment cell height in processes admin

n1k0:
## :wrench: Problem

Before:

image

## :cake: Solution

After:

image


paulboosz (approved):
Yhaa

PR #1736, opened 2026-01-20
feat: add elm auto-formatting on git pre-commit

n1k0:
## :wrench: Problem

Elm code should be formatted on pre-commit

## :cake: Solution

Configure precommit accordingly

Tweak some elm, introduce erroneous formatting, commit, enjoy.


cedricr (commented):
Nice !
Two other things I’d like to do in future PRs:
- exclude the `.venv` folder from checks, as fix:all will check everything inside and raise lots of noise
- probably add elm-check on the CI too, for the rare case where the pre-commit is not properly configured.

Apart from that, everything looks good to me, except I think you’re missing a dot somewhere.

PR #1843, opened 2026-02-24
chore: sort processes

paulboosz:
## :wrench: Problem

Currently processes.json is unsorted but activities.json is

See also: https://github.com/MTES-MCT/ecobalyse-data/pull/223

## :cake: Solution

Sort processes.json using the same key as activities.json


cedricr:
I’ve added the link to the data PR – it’s better for tracability, and it helps when reviewing too!

PR #1844, opened 2026-02-24
fix: backport #1825 and #1826

ccomb:
backport #1825 (close the modal when clicking outside) and #1826 (remove css transition on autocomplete lists)


n1k0 (approved):
Thanks!

PR #1906, opened 2026-03-13
chore: new export with a cull cow meat

ccomb:
meat without bone from cull cow


cedricr:
@ccomb where’s the corresponding data PR ?

PR #1928, opened 2026-03-23
fix: database connection pooling

vjousse:
## :wrench: Problem

## :cake: Solution

### DIFF
--- a/backend/src/app/domain/accounts/controllers/access.py
+++ b/backend/src/app/domain/accounts/controllers/access.py
@@ -203,15 +203,18 @@ class AccessController(Controller):
 
         if cache_duration:
             logger.info(f"-> Using cache duration `{cache_duration}`")
+            print(f"-> Using cache duration `{cache_duration}`")
             memory_store = request.app.stores.get("memory")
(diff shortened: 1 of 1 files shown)


vjousse:
Documenting the problem here.

As we have a L Scalingo instance, `WEB_CONCURRENCY` variable is set to 2 by Scalingo and so 2 processes are spawned by `uvicorn` see the code here https://github.com/Kludex/uvicorn/blob/02bed6f8c38e74f684bb0e572977a9bfdc1f6fea/uvicorn/config.py#L335

Each python worker opens its own PG pool, it’s `DATABASE_POOL_SIZE` (10) + `DATABASE_MAX_POOL_OVERFLOW` (15) so 25 connections * 2, so it tries to open 50 connections to PG under heavy load and we have only 30 connections available with our Scalingo plan.

Downscaling `DATABASE_POOL_SIZE` to 8 and `DATABASE_MAX_POOL_OVERFLOW` to 4 allows for a maximum of (8+4)*2=24 connections at the same time, fixing the problem.

PR #1965, opened 2026-03-27
feat(data): modify material types

nicolaspkandeel:
Add material type with percentage, without processes changes

## :wrench: Problem

Fixes https://github.com/MTES-MCT/ecobalyse/issues/1921

See also : https://github.com/MTES-MCT/ecobalyse-data/pull/253


nicolaspkandeel:
Compared with XLS simulation.
The incineration process should be modified, but the target process is not available and must be added in ecobalyse-data. Not sensible => ready to merge.

PR #1978, opened 2026-03-31
chore(security): update dependabot config to handle branches and ecosystems

n1k0:
This is a backport of #1977 for the stable/textile branch


n1k0:
I just checked, only the main branch matters 👍 closing

PR #2027, opened 2026-04-08
feat: check ingredient hierarchy

paulboosz:
## :wrench: Problem

fix https://github.com/MTES-MCT/ecobalyse/issues/2217

## :cake: Solution

Github action : Check ingredients hierarchy
Trigger : 
- on click
- on merge with master
- on PR creation if data files are modified

For each base_product (carrot, artichoke, salt, ...) do : 
- check hierarchy : organic < fr < eu < default
- if hierarchy is not respected -> register a violation in a csv
image

- create 1 plot of with all variant impact for comparison
image

- create bookmark importable in ecobalyse with all variant impacts for interactive comparison
image

WARNING : can look inconsistent with the plot because ecobalyse also include other lifecycleSteps like transport

- The report is downloadable in  `Checks` -> `Ingredient hierarchy check` -> `Upload report and plots` 

## Remarks

- no impacts change
- a `baseIngredient` field is added to compare ingredients within a group

## Future

- There is ~30 violations that are being investigated right now to see if it's abnormal or not
- Once the violation are classified as coherent or not by the method team, we could make the test fail if there a new violation to act as an alert

see also : 
https://github.com/MTES-MCT/ecobalyse-data/pull/294


bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+    Returns:
+        by_base: dict mapping base_ingredient -> list of ingredient dicts (with added fields)
+    """
+    by_base = defaultdict(list)
+
+    for ingr in ingredients:
+        if not ingr.get("visible", False):
+            continue
+        base = ingr.get("baseIngredient")
+        if not base:
```
cedricr: Shouldn’t that be an error, not having a base ingredient?

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+        else:
+            factors[key] = 0
+    return factors
+
+
+def fetch_ingredient_impacts(ingredient_id, api_url):
+    """Call the food API for a single ingredient at 1kg."""
+    url = f"{api_url}/api/food/"
+    payload = {"ingredients": [{"id": ingredient_id, "mass": 1000}]}
+    headers = {"Authorization": "Bearer dummy"}
```
cedricr: How does that work ?

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+        ecs_total: total ECS score (µPt)
+    """
+    results = api_response["results"]
+    ingredients_total = results["recipe"]["ingredientsTotal"]
+    bonus_impacts = results["recipe"]["totalBonusImpact"]
+
+    impacts_norm = {}
+    for key, value in ingredients_total.items():
+        factor = norm_factors.get(key, 0)
+        impacts_norm[key] = 1e6 * value * factor
```
cedricr, line 163: maybe add a comment here

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+    complements_norm = {}
+    for key, value in bonus_impacts.items():
+        complements_norm[key] = value  # already in µPt
+
+    ecs_total = sum(impacts_norm.values()) + sum(complements_norm.values())
+    return impacts_norm, complements_norm, ecs_total
+
+
+def fetch_all_impacts(by_base, api_url, norm_factors):
+    """Fetch impacts for all ingredients. Returns a dict: alias -> result dict."""
```
cedricr: Why not add type metadata, instead of putting that in the comment?

cedricr: (and idem for the other functions)

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+        valid = [v for v in variants if v["alias"] in impact_results]
+        if len(valid) < 2:
+            continue
+
+        # Sort by ECS for consistent ordering
+        valid.sort(key=lambda v: impact_results[v["alias"]]["ecs_total"])
+
+        bookmarks = []
+        for i, ingr in enumerate(valid):
+            bm = make_bookmark(ingr["alias"], ingr["id"], base_ts + count * 100 + i)
```
cedricr, line 457: I guess you should explain why each bookmark need a different timestamp here

cedricr (commented):
Fantastic job 👏👏 

Here’s a first partial review while I do some tests.

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+#     "requests",
+# ]
+# ///
+"""Check ingredient impact hierarchy and generate comparison graphs + bookmarks.
+
+For each base food product, verifies that a hierarchy of impacts is respected
+among its ingredients (organic < fr < eu < non-ue < default), and produces:
+- Stacked bar charts per base product (output/ingredient_plots/)
+- Importable bookmark files for the Ecobalyse comparator (output/bookmarks/)
+- A CSV report of hierarchy violations (output/ingredient_hierarchy_report.csv)
```
cedricr: maybe add the description of the _fr.csv file too

bin/check_ingredient_hierarchy.py
paulboosz: indeed

bin/check_ingredient_hierarchy.py
paulboosz: Not sure but thats how we do it in score_history.py

bin/check_ingredient_hierarchy.py
paulboosz: done and also an issue : https://github.com/MTES-MCT/ecobalyse/issues/2262

bin/check_ingredient_hierarchy.py
cedricr: ok @vjousse tells me that works in the CI because it’s running in test mode, but it doesn’t work locally.

Furthermore, it might stop working soon, as that’s exposing a bug in the API (it shouldn’t return a result with an incorrect token) ; except if, in fixing this, we keeps the exception in test mode.

In any case, I’d really like something that

- can be tested locally (at least, make the auth token an env variable)
- doesn’t abuse an bug in the API :zany_face: 
- clearly explain what it’s doing and why.

BTW, do we really need to go through the API ? What are we missing without, the normalized values ?

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+
+# --- Step 1: Parse ingredients and group by base product ---
+
+
+def load_json(path):
+    with open(path) as f:
+        return json.load(f)
+
+
+def parse_variant_type(alias, base_ingredient):
```
cedricr: we should probably error out when the suffix is not in INGREDIENT_ORDER (which might get a better name), what do you think ?

bin/check_ingredient_hierarchy.py
paulboosz: done

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+import matplotlib.pyplot as plt
+import pandas as pd
+import requests
+
+# Constants
+
+PROJECT_ROOT = pathlib.Path(__file__).parent.parent.resolve()
+INGREDIENTS_PATH = PROJECT_ROOT / "public" / "data" / "food" / "ingredients.json"
+IMPACTS_PATH = PROJECT_ROOT / "public" / "data" / "impacts.json"
+OUTPUT_DIR = PROJECT_ROOT / "output"
```
cedricr, line 36: This should probably be added to .gitignore

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
+#!/usr/bin/env python3
```
cedricr:
```suggestion
#!/usr/bin/env -S uv run --script
```

see https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file.
And the script permissions should be changed with
chmod a+x bin/check_ingredient_hierarchy.py

cedricr (commented):
Second part of the review

.github/workflows/ingredient_hierarchy.yml
```diff
@@ -0,0 +1,91 @@
...
+            fi
+            echo "-> Waiting for the Node server to boot..."
+            sleep 1
+          done
+
+          uv run bin/check_ingredient_hierarchy.py http://localhost:8001
+
+      - name: Upload report and plots
+        uses: actions/upload-artifact@v4
+        if: always()
```
cedricr, line 45: is this `if: always()` really needed here ?

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+    rows = []
+    for alias in aliases:
+        result = impact_results.get(alias)
+        if not result:
+            continue
+        for impact_key, val in result["impacts_norm"].items():
+            if impact_key in EXCLUDED_IMPACTS:
+                continue
+            rows.append(
+                {"product_name": alias, "impact": impact_key, "cout_enviro": val}
```
cedricr:
```suggestion
                {"product_name": alias, "impact": impact_key, "ecs": val}
```

strange to have a french abbrev with other keys in english.

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+            )
+        # Complements (ecosystemic services): leading space sorts them first in the legend.
```
cedricr: maybe `_` instead ? I don’t remember if it’s sorted before, but an explicit char looks less like a typo.

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+    plt.tight_layout()
+    plt.savefig(output_path)
+    plt.close()
+    return True
+
+
+def generate_all_plots(by_base, impact_results, bases_with_violations):
+    """Generate stacked bar charts for all base products with 2+ variants."""
+    count = 0
+    for base, variants in sorted(by_base.items()):
```
cedricr: Except if people said it was nice to have all the plots, I’d only generates the ones showing a violation. And same thing for the bookmarks.

That would also facilitate an obvious next step that would be to include them in the PR comments !

cedricr (commented):
Last part !

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,593 @@
...
+INGREDIENTS_PATH = PROJECT_ROOT / "public" / "data" / "food" / "ingredients.json"
+IMPACTS_PATH = PROJECT_ROOT / "public" / "data" / "impacts.json"
+OUTPUT_DIR = PROJECT_ROOT / "output"
+PLOTS_DIR = OUTPUT_DIR / "ingredient_plots"
+BOOKMARKS_DIR = OUTPUT_DIR / "bookmarks"
+
+# --- Sanity checks ---
+#
+# All expected impact-ordering sanity checks live here.
+# For each sanity check, add a comment to explain the reasons behind the check
```
cedricr: Better reformulate this, I know it’s not, but it reads like a prompt 🙈 😆

bin/check_ingredient_hierarchy.py
paulboosz: Haha yes I used this was working in score_history.py so I used it but I understand it's not clean.

One issue to prefer the api to data files is that complements values in data files are not final, they are multiplied in the front. But my next task https://github.com/MTES-MCT/ecobalyse/issues/2106 was to move this to the data pipeline so i'll do that first and refactor this PR to just use data files.

Sidenote : score_history.py is exploiting so i'll need to update this if we fix this bug. In score_history I'm storing scores of products (Tshirt,...) so it makes more sense to use the api.

paulboosz (commented):
Thanks for the review ! I'll do https://github.com/MTES-MCT/ecobalyse/issues/2106 first to simplify this PR

bin/check_ingredient_hierarchy.py
paulboosz: done

bin/check_ingredient_hierarchy.py
paulboosz: Yes seen with the method team, it's useful to have all the plots

bin/check_ingredient_hierarchy.py
paulboosz: Not sure because there is a lot of alias suffix not in INGREDIENT_HIERARCHY : with the name of the country (-tr, -de, -wi) or with other suffix (-overseas, ...).

.github/workflows/ingredient_hierarchy.yml
paulboosz: right now it's not because check_ingredient_hierarchy.py doesn't fail but in the near future the idea is to have a failure when a new violation arise. In that case report won't be uploaded and job will stop unless we have if: always

bin/check_ingredient_hierarchy.py
paulboosz: done

paulboosz: done

bin/check_ingredient_hierarchy.py
paulboosz: actually matplotlib hides any legend beginning with `_`. I used `*`

paulboosz:
@cedricr ready for a new review

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,572 @@
...
+# Sanity checks between different base_ingredient : (lower_alias, higher_alias)
+EXPLICIT_PAIR_SANITY_CHECKS = [
+    # The most impactful chicken should have lower impact than conventional beef
+    ("chicken-breast-br-max", "beef-without-bone"),
+    # Organic beef should have an impact lower than the most impactful chicken
+    ("beef-without-bone-organic", "chicken-breast-br-max"),
+]
+
+
+# These impacts are = 0 so we exclude them to remove noise from the graph
```
ccomb: `etf` is not really zero, but we still want to ignore it

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,572 @@
...
+
+def compute_normalization_factors(impacts_data: dict[str, Any]) -> dict[str, float]:
+    factors: dict[str, float] = {}
+    for key, val in impacts_data.items():
+        if val.get("ecoscore"):
+            factors[key] = (
+                val["ecoscore"]["weighting"] / val["ecoscore"]["normalization"]
+            )
+        else:
+            factors[key] = 0
```
ccomb: Default a missing factor to 0, silently dropping any impact key absent from impacts.json.
Worth a warning?

ccomb (approved):
Excellent, need to land fast!

Possible nice follow-up : render a self-contained web page inside the artefact

```
  import base64

  def img_tag(path):
      b64 = base64.b64encode(path.read_bytes()).decode()
      return f'<img src="data:image/png;base64,{b64}" style="max-width:100%">'

  html = [
      "<h1>Ingredient hierarchy report</h1>",
      df.to_html(index=False),
      *(f"<h2>{p.stem}</h2>{img_tag(p)}" for p in sorted(PLOTS_DIR.glob('*.png'))),
  ]
  (OUTPUT_DIR / "report.html").write_text("<html><body>" + "\n".join(html) + "</body></html>")
```

Or even better : publish output/ to GitHub Pages

bin/check_ingredient_hierarchy.py
```diff
@@ -0,0 +1,572 @@
...
+        ecs_with_complements += value
+    if "complements" in ingredient["metadata"]:
+        for key, value in ingredient["metadata"]["complements"].items():
+            if value:
+                ecs_with_complements += value
+
+    return ecs_with_complements
+
+
+# --- Step 4: Check hierarchy ---
```
cedricr: Step 2 and 3 are missing ;-)

cedricr (approved):
Niiice !

bin/check_ingredient_hierarchy.py
paulboosz: indeed, explicited this in comment

PR #2028, opened 2026-04-08
feat(experimental): contribute examples PRs from the ui

n1k0:
This patch adds a new contribution section in the sidebar of generic simulators (object, veli, food2), so authenticated users can submit their simulations to be added to the official ecobalyse exemples static JSON database (`public/data/{scope}/examples.json` files), through a github PR. The created PR is automatically assigned for review to a new reviewing team especially created.

Sample PR created: #2079 (notice the reviewing team assigned)

## ⚠️ Points to watch, caveats

- There's an issue with prettier getting in the way when checking examples JSON formatting, so I've added the examples JSON files to .prettierignore
- I had to resort to creating a [github team](https://github.com/orgs/MTES-MCT/teams/ecobalyse-contrib-reviewers) to assign as PR reviewer, as otherwise we'd have needed to make ecobalyse-bot an admin, which sounded overkill and unecessarry risky. We'll have to put methodology people in this team, though they'll need to become members of the MTES-MCT org first as a prerequisite

## Before the patch lands

On the scalingo production app:

- [x] set the `GITHUB_TOKEN` env var to @ecobalyse-bot's auth token (filed in vaultwarden)
- [x] set the `GITHUB_REVIEWING_TEAM` env var to "ecobalyse-contrib-reviewers" (team members are ediatable [here](https://github.com/orgs/MTES-MCT/teams/ecobalyse-contrib-reviewers))


n1k0:
I'd like preliminary feedback with the patch and basic features :)

cedricr:
> I'd like preliminary feedback with the patch and basic features :)

I’ll look at that first thing on monday morning!

camcoq-start:
Awesome ! Works very well. I will test it more with more exemple :)

Just one feedback: 
- i can't contribute from textile simulator. It could be interesting to add a message here as: "Contribution is not possible in textile simulator. If you want to contribute, use object simulator."

n1k0:
> i can't contribute from textile simulator.

@camcoq-start indeed, neither from food1, as these two scopes are not generic ones :) I'll add a message (or better, hide the tab section in these two scopes)

camcoq-start:
Thanks for hiding contribution feature in textile and food1 :) looks less disturbing.
I approve this new feat ! :)

Last feedback to challenge together: do you think we should add more context on the PR created ? 
For example: 
- assign directly to alban and nicolas who will review these contributions
- add a little intro for users who open the PR and are not used to git 
ie: "La contribution est soumise à validation par l'équipe méthode d'Ecobalyse. Vous pouvez suivre sur ce lien l'état d'avancement de l'intégration de votre exemple de produit."

camcoq-start (approved):
works perfectly :) let's communicate to our users about the feature !!

n1k0:
> - assign directly to alban and nicolas who will review these contributions
add a little intro for users who open the PR and are not used to git
> - ie: "La contribution est soumise à validation par l'équipe méthode d'Ecobalyse. Vous pouvez suivre sur ce lien l'état d'avancement de l'intégration de votre exemple de produit."

I'll make the list of reviewers configurable and add a bit more context in the success message :)

n1k0:
@camcoq-start @cedricr @vjousse implementing reviewers assignment is much trickier than expected so I'm gonna have to revert that part. Explanation is the github API apparently requires the review assignment to be performed by another user than the one creating the PR. Which would invlove creating a second machine user just for that, and I don't think we want to. Sorry!

n1k0:
Ok, I had to resort to creating a [github team](https://github.com/orgs/MTES-MCT/teams/ecobalyse-contrib-reviewers) to assign as PR reviewer, as otherwise we'd have needed to make ecobalyse-bot an admin, which sounded overkill and unecessarry risky. We'll have to put methodology people in this team, though they'll need to become members of the MTES-MCT org first as a prerequisite

backend/src/app/domain/contrib/controllers/contrib.py
```diff
@@ -0,0 +1,30 @@
...
+        operation_id="CreateExampleContrib",
+        path=urls.CONTRIB_EXAMPLES,
+        guards=[requires_active_user],
+    )
+    async def create_example_contrib(
+        self,
+        current_user: m.User,
+        data: ExampleContribCreate,
+    ) -> ExampleContribResponse:
+        settings = get_settings()
```
vjousse: Nit: usually we put the `get_settings` after the imports so that it can fail at boot when loading the file, instead of failing at runtime.

backend/src/app/domain/contrib/schemas.py
```diff
@@ -0,0 +1,24 @@
...
+from enum import StrEnum
+from typing import Any
+
+from app.lib.schema import CamelizedBaseStruct
+
+
+class GenericScope(StrEnum):
+    FOOD2 = "food2"
+    OBJECT = "object"
+    VELI = "veli"
```
vjousse: There is an already existing `Scope` enum in `app.domain.components.schemas.Scope` we should at least put the declarations together (maybe in the `components` module for now, and later on by creating a separate `scope` module).

The `Scope` or `AllScope` enum could inherit from the `GenericScope` to avoid duplication.

backend/src/app/domain/contrib/services.py
```diff
@@ -0,0 +1,211 @@
...
+    GenericScope,
+)
+from httpx import AsyncClient
+from litestar.exceptions import ValidationException
+
+logger = structlog.get_logger()
+
+
+def get_examples_path(scope: GenericScope) -> str:
+    return f"public/data/{scope.value}/examples.json"
```
vjousse: I think that putting this string in the settings could make sense.

backend/src/app/domain/contrib/services.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def get_github_api_url(repo: str, path: str) -> str:
+    return f"https://api.github.com/repos/{repo}/{path}"
```
vjousse: Same as above, I would like to have it in the settings

backend/src/app/domain/contrib/services.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def clean_str(s: str, fallback: str = "") -> str:
+    return (s or fallback).strip()
+
+
+def get_user_full_name(profile: m.UserProfile) -> str:
+    first_name = clean_str(profile.first_name)
+    last_name = clean_str(profile.last_name)
+    return clean_str(" ".join([first_name, last_name]), "Anonyme")
```
vjousse: This `clean_str` method looks strange, why do you need it? (if profile strings are sometimes invalid because they contain trailing spaces, we should remove the spaces when creating the profile I think)

backend/src/app/domain/contrib/services.py
```diff
@@ -0,0 +1,211 @@
...
+def format_json_string(json_string: str) -> str:
+    # Note:  these files are ignored by prettier, so we're safe from formatting conflicts
+    # FIXME: this is extracted from the ecobalyse-data repository, we should use a single
+    #        source for formatting JSON
+    return (
+        json.dumps(
+            json.loads(json_string), ensure_ascii=False, sort_keys=True, indent=2
+        )
+        + "\n"
+    )
```
vjousse: I think that in should be moved into a new file in `lib/json.py` for example

backend/src/app/domain/contrib/services.py
```diff
@@ -0,0 +1,211 @@
...
+            "### Contributeur",
+            "",
+            f"- Nom : {user_full_name}",
+            f"- Organisation : {org_info}",
+            "",
+            "### Paramètres de l’exemple",
+            "```json",
+            query_as_string,
+            "```",
+        ]
```
vjousse: I think that it could be simplified with a `f"""…"""` python string

src/Views/Bookmark.elm
```diff
@@ -408,6 +435,76 @@ bookmarkView cfg ({ name, query } as bookmark) =
...
+        disabledForm =
+            config.contribRequestPending || not isAuthenticated || simulationIsEmpty || simulationExists
+
+        error icon msg =
+            div [ class "text-danger fs-7" ] [ icon, text <| "\u{00A0}" ++ msg ]
+    in
+    div [ class "card-body d-flex flex-column gap-2" ]
+        [ div [ class "fs-7 text-muted" ]
+            [ text <|
+                "Proposez l'ajout de votre simulation à la base de données d’exemples "
```
vjousse:
```suggestion
                "Proposez l’ajout de votre simulation à la base de données d’exemples "
```
🤗

vjousse (approved):
Some nits with the python code but great work, it could be landed as it is for me!

backend/src/app/domain/contrib/services.py
n1k0: I did that but reverted it, I see no benefit putting the string template in the settings, as I see no use case of it. We can always update the string in the function directly, if we ever update the place we store these example files.

PR #2085, opened 2026-04-15
feat: add « Europe et Maghreb » country

### LINKED ISSUE #1985: Ajouter la région Europe Maghreb
## :wrench: Problème

On aimerait avoir la région Europe Maghreb dans la liste des régions pour le food1.

## :cake: Solution

Ajouter une entrée au fichier `countries.json`

## :rotating_light: Points d'attention/commentaires

Attention, la région Europe Maghreb doit avoir une distance de 2000km parcourue par camion vers la France.
Le ratio terre/mer ne s'applique pas à cette distance.

## :desert_island: Comment tester

Une nouvelle région «  Europe et Maghreb » devrait être disponible dans le sélecteur de pays des ingrédients food1. La distance de cette région à la France devrait être 2660 km par route.

vjousse:
fixes #1985

### DIFF
--- a/public/data/countries.json
+++ b/public/data/countries.json
@@ -27,6 +27,15 @@
     "zone": "Europe"
   },
   {
+    "aquaticPollutionScenario": "Best",
+    "code": "REM",
+    "electricityProcessId": "ee114bf1-8c51-5c1e-8065-d762400447e2",
+    "heatProcessId": "f6ea2983-e024-5de2-b323-273f2436deba",
+    "name": "Région - Europe et Maghreb",
+    "scopes": ["food"],
+    "zone": "Europe"
+  },
+  {
     "aquaticPollutionScenario": "Worst",
     "code": "RAS",
     "electricityProcessId": "a2129ece-5dd9-5e66-969c-2603b3c97244",
--- a/public/data/countries.json
+++ b/public/data/countries.json
@@ -29,7 +29,8 @@
   {
     "aquaticPollutionScenario": "Best",
     "code": "REM",
-    "electricityProcessId": "ee114bf1-8c51-5c1e-8065-d762400447e2",
+    "comment": "Elec and head mixes are the same than 'Région - Europe de l'Est' but are not used by the 'food' simulator",
+    "electricityProcessId": "5ee392cf-954e-56c8-b555-427d55023548",
     "heatProcessId": "f6ea2983-e024-5de2-b323-273f2436deba",
     "name": "Région - Europe et Maghreb",
     "scopes": ["food"],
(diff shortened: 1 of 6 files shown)


public/data/transports.json
```diff
@@ -494,6 +495,9 @@
     "US": { "air": 8254, "road": null, "sea": 12246 },
     "VN": { "air": 8703, "road": 10866, "sea": 13893 }
   },
+  "REM": {
+    "FR": { "air": null, "road": 2000, "sea": null }
+  },
```
n1k0, line 500: This means the matrix can't resolve distances from/to REM with other countries than France, which may be fine for this very patch, but it feels awkward from a design perspective. More importantly, I thought the source of this table was [generated from ecobalyse-data](https://github.com/MTES-MCT/ecobalyse-data/tree/main/common/distances) by @paulboosz, so we may face future syncing issues if we decide maintaining the two datasources in parallel :/

camcoq-start (approved):
LGTM :) Well done
May be one thing should be fixed. In an other PR is also ok
The default origin Par défaut(Europe - Maghreb) has 3160km whereas Region- Europe et Maghreb has 2660.

Change the distance of 2500km for default origin Europeandmaghreb by 2000km could be solve the issue

public/data/transports.json
cedricr: Indeed ; I’ll generate the file properly on the data side.

cedricr: ok, that’s harder than I thought: @paulboosz when I regenerate the transports.json file (with `just export-transports`), the generated file have a loooot of diffs with the one in ecobalyse.
Any idea why ? It used to work not so long ago.

vjousse:
@camcoq-start fixed in [a0f9477](https://github.com/MTES-MCT/ecobalyse/pull/2085/commits/a0f94776bb7699a499ee9fd10143a7c3f40b40ae)

public/data/transports.json
cedricr: That being said, I’d say we should not block this PR for that, and address it separately on the data side later on.

cedricr: I’ll create the ticket about that.

public/data/countries.json
```diff
@@ -26,6 +26,16 @@
     "scopes": ["textile", "object", "veli", "food", "food2"],
     "zone": "Europe"
   },
+  {
+    "aquaticPollutionScenario": "Best",
+    "code": "REM",
+    "comment": "Elec and head mixes are the same than 'Région - Europe de l'Est' but are not used by the 'food' simulator",
```
cedricr:
```suggestion
    "comment": "Elec and heat mixes are the same than 'Région - Europe de l'Est' but are not used by the 'food' simulator",
```

PR #2092, opened 2026-04-15
fix(contrib): catch review assignment error

n1k0:
This patch fixes a poorly worded error message rendered to users after a successful contrib when the github API fails assigning a team as reviewers. Note that we'll eventually have to find a better way to assign reviewers because right now the gh permission mechanism gets in the way in a nasty fashion.


n1k0:
There are still issues I'll keep working on, hold on reviewing, Edit: solved, reassigning for review

n1k0:
Well, while I'm at it, I'm gonna add an examples.json file for food2. Shouldn't be much more work.

Edit: done in a2d93574f515c56af2f0ea51854f3ad8e9130e6d

PR #2082, opened 2026-04-15
chore: align displayName v2

cedricr:
## :wrench: Problem

Display names are not homogenous

See also: https://github.com/MTES-MCT/ecobalyse-data/pull/258

PR #2098, opened 2026-04-16
fix: batch commit in load_processes_fixtures

ccomb:
## :wrench: Problem

Deployment on Scalingo is stuck for ~2h in the postdeploy hook (load-fixtures.sh), during process loading (load_processes_fixtures).                                                                                                        
     
## :cake: Solution

This is an unconfirmed hypothesis:
                                                                                                                                                                                                                                         
  The function updates each existing process individually, with each one requiring: an UPDATE, a COMMIT, a REFRESH (re-read from DB) and eager loading (selectinload/joinedload). On a remote DB, each cycle takes ~2-3 seconds.              
                                                                                                                                                                                                                                              
  This code was introduced in #1453 (Oct 2025) when there were ~490 processes — it took ~8 minutes, which was tolerable. The process count has since grown steadily, and two recent PRs pushed it past the breaking point:                    
                                                                                                                                                                                                                                              
  - #2005 
  - #2082
                                                                                                                                                                                                                                              
  At 2287 processes, the deployment takes ~1h40 and times out.

So ;

- Set all operations (create, update, delete) to auto_commit=False
- Remove auto_refresh=True and the load=[...] from individual updates (we don't need to re-read data after each update)
- Perform a single await db_session.commit() at the end          
                                                                                 
Goes from ~2500 individual transactions to a single one. 

### Points to watch

See the previous comments on this topic on: #1453

- Run integration tests
- Deploy and verify the postdeploy hook completes in seconds instead of ~2h
- Check in the back-office that processes are present and up to date after deployment

### DIFF
--- a/backend/src/app/cli/commands.py
+++ b/backend/src/app/cli/commands.py
@@ -30,7 +30,6 @@ from app.domain.components.schemas import (
 )
 from app.domain.processes.deps import provide_processes_service
 from rich import get_console
-from sqlalchemy.orm import joinedload, selectinload
 from structlog import get_logger
 
 
@@ -383,7 +382,7 @@ async def load_processes_fixtures(
     if processes_to_add:
         await processes_service.create_many(
             data=processes_to_add,
-            auto_commit=True,
+            auto_commit=False,
         )
 
     if processes_to_update:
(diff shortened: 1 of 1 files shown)


vjousse (approved):
Not sure that the current deploy problem is due to that, but it makes sense from an optimization perspective.

PR #2103, opened 2026-04-19
chore: update dependencies

cedricr:
## :wrench: Problem

Updates python dependencies via `uv lock --upgrade`, and js dependencies via `npm audit fix`.

Should close Dependabots issues.


n1k0 (approved):
LGTM

PR #2110, opened 2026-04-20
chore: remove cookie auth

### LINKED ISSUE #2037: [backend] Remove cookie-based auth for request
Because this may be a csrf vector attack, and we don't use it.

vjousse:
fixes #2037 
Auth workflow should still work as before.


backend/src/app/lib/auth.py
n1k0, line 1: Argh, no other possible way to avoid duplicating the whole module? This is a major risk we're taking as we may easily miss security updates from core…

n1k0 (commented):
I'd like inputs from @cedricr here

backend/src/app/lib/auth.py
cedricr: @vjousse yeah, can’t you do something like
```
class CustomOAuth2PasswordBearerAuth(OAuth2PasswordBearerAuth)
```
and overload the minimal subset of fields possible ?

cedricr (commented):
Automatic tests pass
Manual login/logout, and checking the cookies/localstorage => everything looks ok

I’d like for OAuth2PasswordBearerAuth not to be completely duplicated if possible too !

backend/src/app/lib/auth.py
vjousse: I’ve just override the `create_response` to remove the cookies, it should do the trick!

cedricr (approved):
Works perfectly 🚀

n1k0 (approved):
🍪🔫

PR #2111, opened 2026-04-20
feat(generic): edit elements in their own modal window

n1k0:
Follow-up of #1886, implement progressive disclosure for component item elements edition, through a modal window.

image
&nbsp;
image
&nbsp;
image


vjousse (approved):
It looks good to me from a code point of view. Regarding the UI, I am more and more prone to think that we are doing the next Simapro here… but this modal looks like an improvement to me anyway, great work!

PR #2140, opened 2026-04-23
chore: perf improvements in explorer

vjousse:
## :wrench: Problem

Searching processes in the explorer is very slow, especially for `food2` where there are a lot of processes in the DB.

## :cake: Solution

Compute the word indexation once when loading the DB and don’t encode the CSV for every keystroke, do it only when downloading the file.

Open the `#/explore/food2` URL and try to search for terms, the display should be a lot faster than what we currently have here https://ecobalyse.beta.gouv.fr/#/explore/food2

Downloading the corresponding CSV should work the same as before.


vjousse:
Replaced by https://github.com/MTES-MCT/ecobalyse/pull/2154

PR #2243, opened 2026-05-11
chore: set defaultOrigin ot FranceOutreMer

paulboosz:
see also :https://github.com/MTES-MCT/ecobalyse-data/pull/297


camcoq-start (approved):
Good to me :)

PR #2256, opened 2026-05-12
perf: score history multiprocessing

paulboosz:
## :wrench: Problem

Improve score history


paulboosz:
doesn't improve perf :(

PR #2300, opened 2026-05-19
chore: exclude data folder from Scalingo image

cedricr:
## :wrench: Problem

After merging the `ecobalyse-data` repository, Scalingo image size increased without need.

## :cake: Solution

Add the `data` folder to the `.slugignore` file.

### DIFF
--- a/.slugignore
+++ b/.slugignore
@@ -4,6 +4,7 @@ src/
 tests/
 styles.scss
 review/
+data/
 
 node_modules/@parcel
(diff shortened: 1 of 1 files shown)


cedricr:
La PR “Monorepo” (34f70e6) faisait passer la taille de l’image à 909 MB.

image

Cette PR la fait repasser à 855

image

n1k0 (approved):
indeed 😅

PR #2301, opened 2026-05-19
feat(generic): update and apply intra-country distances

cedricr:
## :wrench: Problem

See https://github.com/MTES-MCT/ecobalyse/issues/1519
and https://github.com/MTES-MCT/ecobalyse/issues/1531

### DIFF
--- a/public/data/countries.json
+++ b/public/data/countries.json
@@ -2,459 +2,333 @@
   {
     "aquaticPollutionScenario": "Worst",
     "code": "---",
+    "distanceToHub": 0,
     "electricityProcessId": "ed6d177e-44bb-5ba4-beec-d683dc21be9f",
     "heatProcessId": "3561ace1-f710-50ce-a69c-9cf842e729e4",
     "name": "Pays inconnu (par défaut)",
-    "scopes": [
-      "food",
-      "textile"
-    ],
-    "distanceToHub": 0,
+    "scopes": ["food", "textile"],
     "zone": "Asia"
   },
   {
     "aquaticPollutionScenario": "Best",
     "code": "REO",
+    "distanceToHub": 500,
     "electricityProcessId": "ee114bf1-8c51-5c1e-8065-d762400447e2",
     "heatProcessId": "f6ea2983-e024-5de2-b323-273f2436deba",
     "name": "Région - Europe de l'Ouest",
-    "scopes": [
-      "textile",
-      "object",
-      "veli",
-      "food",
-      "food2"
-    ],
-    "distanceToHub": 500,
+    "scopes": ["textile", "object", "veli", "food", "food2"],
     "zone": "Europe"
   },
   {
     "aquaticPollutionScenario": "Best",
     "code": "REE",
+    "distanceToHub": 2000,
     "electricityProcessId": "5ee392cf-954e-56c8-b555-427d55023548",
     "heatProcessId": "f6ea2983-e024-5de2-b323-273f2436deba",
     "name": "Région - Europe de l'Est",
-    "scopes": [
-      "textile",
-      "object",
-      "veli",
-      "food2"
-    ],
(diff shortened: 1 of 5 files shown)


src/Data/Country.elm
```diff
@@ -76,7 +76,7 @@ decode processes =
         |> Pipe.required "aquaticPollutionScenario" decodeAquaticPollutionScenario
         |> Pipe.required "code" decodeCode
         |> DU.strictOptional "comment" Decode.string
-        |> DU.strictOptionalWithDefault "distanceToHub" (Decode.map Length.kilometers Decode.float) (Length.kilometers 0)
+        |> Pipe.required "distanceToHub" (Decode.map Length.kilometers Decode.float)
```
cedricr, line 79: @n1k0 is that a good idea to remove the optionality here ? And not sure I did that properly…

cedricr:
@n1k0 @camcoq-start I *think* I’ve implemented the distances as requested in https://github.com/MTES-MCT/ecobalyse/issues/1519, but I can’t manage to get the values expected in the tests described in https://github.com/MTES-MCT/ecobalyse/issues/1531.

src/Data/Country.elm
n1k0: @cedricr yes good call, I made it optional so it compiles but now it MUST be required :)

public/data/countries.json
```diff
@@ -295,7 +449,12 @@
     "electricityProcessId": "f3ad3ac5-731a-5ef5-b9e1-13ce71780444",
     "heatProcessId": "3561ace1-f710-50ce-a69c-9cf842e729e4",
     "name": "Vietnam",
-    "scopes": ["textile", "object", "veli"],
+    "scopes": [
+      "textile",
+      "object",
+      "veli"
+    ],
+    "distanceToHub": 500,
```
n1k0: Nit: as it's an established convention, please order the JSON keys alphabetically :)

n1k0 (approved):
Looks good to me from a code perspective! I'll let @camcoq-start validate the resulting computed impacts

camcoq-start (commented):
Data is ok :)
But it seems transport impact is not well calculated.
For example (see screenshot), between Turkey and France, i have sea distance without the road/sea ratio applied

Image

public/data/countries.json
cedricr: @n1k0 good catch, the .prettierignore file was incorrect, and prevented this file from being formatted properly. :pray:

cedricr:
@n1k0 handing it back to you ?

camcoq-start (approved):
Only 2 comments : 

- Case 4 is not ok: an element comes from an unknown region and goes to France.
In this case, we are supposed to sum 2000 km, as the default distance to hub for an unknown region, and 500km as France's average distance to hub.

- one rule about unknown case is not well applied. If origin and destination regions are both unknown then double 2000km

Neither of these comments is a blocker.

cedricr:
@camcoq-start I’m not clear on which of the 2 issues I’ve listed in the description should be closed by this PR, could you please check it out ?

camcoq-start:
@cedricr issues are: 
- https://github.com/MTES-MCT/ecobalyse/issues/1519
- https://github.com/MTES-MCT/ecobalyse/issues/1531

n1k0:
@cedricr following up with what we stated during the planning, I think you can safely land the patch now :)

@camcoq-start it may be useful to update [your comment](https://github.com/MTES-MCT/ecobalyse/pull/2301#pullrequestreview-4357690794) with links to created issues :)

camcoq-start:
created issue here: #2345

n1k0:
@camcoq-start thanks, so I'll start working on #2345!

@cedricr I need current this patch landed on master for building upon it, so I'm landing it :)

PR #2303, opened 2026-05-20
fix: reduce precision to remove computational diff

### LINKED ISSUE #2281: Impacts différents sur ma machine
## Explication du bug / Bug Explanation

Quand je run `just export-all` j'ai ce diff (juste à un seul endroit) sur le Radis Origine Inconnue  : 
Image

## Comportement attendu / Expected Behavior

Même impact

## Solution

- diminuer la précision d'une décimale pour effacer ce diff
-  ?

D'autres idées @cedricr ?

paulboosz:
## :wrench: Problem

Fix https://github.com/MTES-MCT/ecobalyse/issues/2281

## :cake: Solution

- Reduce precision 5 decimals -> 4
- fix tests

## Remarks

max diff : 9999.5 -> 1000
max relative diff :  (1000- 9999.5)/1000 = 0.5/1000 ~ 0,05%


cedricr:
@paulboosz no diff for me, and the impact seems very minimal on the final results (as displayed by the tests). I’d say let’s go for it.
Maybe the rounding precision could be a setting, defaulting to this new value, explaining why we choosed it.

cedricr (approved):
Perfect!

PR #2305, opened 2026-05-20
chore: check diff between machines

paulboosz:
## :wrench: Problem

test  #2281

## :cake: Solution

I'd like to confirm what diff we have between machines.

@cedricr @ccomb can you run `just export-all` and see if you have diff with this branch ?

If you have a diff with this branch, I made a PR reducing the float precision from 5 to 4 digits https://github.com/MTES-MCT/ecobalyse/pull/2303. Can you run `just export-all` on this branch and see if you have diff ?


ccomb:
You need to comment or remove the pypardiso line in data/pyproject.toml and `uv sync` to be sure there is no pardiso.

ccomb:
No diff on my side! (except uv.lock after removing pypardiso from pyproject)

cedricr:
@paulboosz I’m not sure I understand what you’re asking here : there are only the generated files in your PR, so of course I have differences (they revert the changes displayed here) :thinking: ?

PR #2365, opened 2026-05-29
chore: fix material type of PET fiber EOL recycling

nicolaspkandeel:
## :wrench: Problem

The material_type for PET fiber EOL recycling was not the correct one.
No consequence except information for users.

## :cake: Solution

fix material type


cedricr (approved):
👍

PR #2177, opened 2026-04-30
feat(data): HUE ingredients

ccomb:
HUE ingredients 
new export from https://github.com/MTES-MCT/ecobalyse-data/pull/287

PR #2075, opened 2026-04-14
feat(data): UE ingredients

ccomb:
UE raw ingredients

from  https://github.com/MTES-MCT/ecobalyse-data/pull/262

Issue #1872, opened 2026-03-03
PR ingrédients origine inconnue/FR/Bio


paulboosz:
https://github.com/MTES-MCT/ecobalyse-data/pull/215

PR #2372, opened 2026-05-29
feat(data): 504 transformed ingredients from variants

ccomb:
## :wrench: Problem

We need more transformed ingredients

See #2371

## :cake: Solution

We now have 5 ingredient variants (FR/BIO/EU/HUE/OI). (since #2075, #1872, #2177, #1904)

Methodology:

Starting with an existing processed ingredient in Agribalyse that contains one of these variants in the upstream processes, we replace this variant with one of the other 4 variants, resulting in 4 processed ingredients.

Example: Starting with carrot juice based on FR carrots, we replace the FR carrot with an organic, EU, HUE, or OI carrot, resulting in organic/EU/HUE/OI carrot juice.

Used a pyvolca script against all loaded databases. See [generate_transformed_ingredients.py](https://github.com/MTES-MCT/ecobalyse-method-tooling/blob/main/food/transformed-ingredients/generate_transformed_ingredients.py)

The CSV table of the new transformed ingredients is [here](https://github.com/MTES-MCT/ecobalyse-method-tooling/blob/163c093e96600d628aa8896ffc536caaa22ef0c5/food/transformed-ingredients/transformed_ingredients.csv).

- `cd data && just test`
- `just import-all && just export-all`


cedricr:
@ccomb I think you forgot to push some files? There are only the generated ones here.

ccomb:
@cedricr thanks for the reminder! I had pushed the wrong branch.

PR #2398, opened 2026-06-04
fix(generic): remove any air transport between transform steps

n1k0:
## :wrench: Problem

When air transport is enabled for shipping to distribution, plane transport was computed and added between element transforms (at the production stage): it shouldn't.

## :cake: Solution

Remove that

## 🔍 Comments, context

Interesting discusssion on how and why the patch fixes the current issue https://mattermost.incubateur.net/fabnum-mte/pl/48syggsgqtfm3ynnmna7x4hftc

Check that air transport is never added bewteen element transform steps.

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -539,7 +539,12 @@ applyTransforms requirements transportOptions initialCountry unit transforms mat
                         |> Result.andThen
                             (\mixes ->
                                 results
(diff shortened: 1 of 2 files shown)


camcoq-start (approved):
Bug is fixed.
Plane is not added anymore to the element score :) thanks and well done !

vjousse (approved):
Ship it! (Not by plane of course. Ahah. Hmmm. Sorry 🤓)

PR #2403, opened 2026-06-04
feat(data): add laminated glass

nicolaspkandeel:
## :wrench: Problem

Laminated glass is current in vehicules. This kind of glass cannont be recycled and should not be modelled ad a component

## :cake: Solution

Model laminated glass as a process with material type "other"

- ECS expected : 99.7.
- No recycling.

### DIFF
--- a/data/lci_catalog/ecobalyse/glass-laminated.json
+++ b/data/lci_catalog/ecobalyse/glass-laminated.json
@@ -1,5 +1,5 @@
 {
-  "activityName": "laminated glass",
+  "activityName": "Laminated glass",
   "categories": [
     "material",
     "material_type:Other"
--- a/data/lci_catalog/ecobalyse/glass-laminated.json
+++ b/data/lci_catalog/ecobalyse/glass-laminated.json
@@ -2,7 +2,7 @@
   "activityName": "Laminated glass",
   "categories": [
     "material",
-    "material_type:Other"
+    "material_type:other"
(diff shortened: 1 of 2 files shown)


nicolaspkandeel:
Ready for export and review @cedricr

data/lci_catalog/ecobalyse/glass-laminated.json
```diff
@@ -0,0 +1,28 @@
+{
+  "activityName": "laminated glass",
```
cedricr: Be careful to use exactly the same name here, it’s case-sensitive.
(I’m doing the change, nothing to do on your side)
```suggestion
  "activityName": "Laminated glass",
```

cedricr (approved):
ecs is “only” 91.37. I’m approving anyway, and let it to you to decide if it’s ok or not !

nicolaspkandeel:
Sorry, there was a mistake in the amount of PVC in activities_to_create. It requires a new export

data/lci_catalog/ecobalyse/glass-laminated.json
```diff
@@ -0,0 +1,28 @@
+{
+  "activityName": "Laminated glass",
+  "categories": [
+    "material",
+    "material_type:Other"
```
cedricr:
```suggestion
    "material_type:other"
```

(I’m doing it too)

PR #2405, opened 2026-06-05
chore: fix process in EoL landfill

nicolaspkandeel:
## :wrench: Problem

For lanfilling at EoL, RoW processes are used whereas :
- CH processes would be more representative
- CH processes are used for incineration
- CH processes are selected for packing material modeling

## :cake: Solution

change RoW process to CH process in EoL landfill

Low reduction of landfill impacts (on average -5%)


nicolaspkandeel:
Ready for export @cedricr

cedricr (approved):

PR #2417, opened 2026-06-08
refactor: allow missing, zeroed by default, impact keys

### LINKED ISSUE #1692: Enlever les impacts à 0 dans `processes.json`
:wrench: Problème

Actuellement dans le fichier `processes.json` lorsque les impacts détaillés sont cachés, ils sont à 0.

Image

- Cela peut induire en erreur.
- Cela ajoute aussi beaucoup de lignes inutiles.

:cake: Solution

Retirer les impacts à 0

n1k0:
## :wrench: Problem

Fixes #1692

## :cake: Solution

Smart-ass decoding!

- All data should be migrated according to the sample process I updated in `processes_generic.json` cc @cedricr @paulboosz 
- The JSON schema used to validate processes, which is duplicated in both frontend and data codebases, has been updated for the frontend only; it should be updated for the data pipeline as well (ideally, we should have only one json schema)
- I had to stop tackling tech debt but this work should be pursued in non-urgent, follow-up PRs

No regressions should happen, test suite should be green.


vjousse (approved):
😍

PR #2422, opened 2026-06-09
test(e2e): make the e2e test reliable without retries

ccomb:
## :wrench: Problem

The e2e tests in playwright could be rerun up to 10× on CI leading to timeout due to a race condition. This is a stateful test with a database and a retry would register the same alice again and again.

## :cake: Solution

A quick fix already landed here and is actually OK to keep
https://github.com/MTES-MCT/ecobalyse/commit/0542c490bc32b85dc6745d6a7a9659dd77814788

This PR is about to make a single run reliable (fix the race on the email wait) and **disable retries for this test**: a failure now surfaces in minutes instead of an hour.

See https://github.com/MTES-MCT/ecobalyse/issues/2408

- Retry several times the runner
- Check the following PRs


n1k0 (approved):
sounds reasonable 🙏

PR #2437, opened 2026-06-10
chore: merge processes files in data pipeline

### LINKED ISSUE #2319: chore: fusionner les fichiers de procédés
## :wrench: Problème

Les procédés sont exportés dans deux fichiers séparés (`processes.json` et `processes_generic.json`) alors que le front les merge pour n’en faire qu’un : ces deux fichiers sont inutiles.

## :cake: Solution

Fusionner `processes_generic` dans `processes`

## :desert_island: Comment tester

[ ] Exemple d'attente
[ ] Autre exemple d'attente

paulboosz:
Fix https://github.com/MTES-MCT/ecobalyse/issues/2319
- new file created by the data pipeline : `processes.json` (and `processes_impacts.json`) now the only processes file in `public/data`
- `processes.json` is created by concatenating intermediate files `processes_legacy.json` and `processes_generic.json` (both live in `data/public/data`)
- To simplify I removed the `LOCAL_EXPORT` boolean env variable. Now export needs 2 paths : output_dir (`public/data`) for final file and local_dir (`data/public/data`) for intermediate files. For test like `test_export.py`, both paths are set to the same `tmp_path`
- I updated the code each time I saw `processes_generic_impacts.json` or `processes_impacts.json` but maybe I forgot something


vjousse:
Why not keeping only `processes_impacts` and removing `processes_generic_impacts` instead of adding a new file?

n1k0:
Indeed, I was genuinely expecting to have only a single resulting `processes.json` file in the `public/data` folder (and a `processes_impacts.json` for detailed impacts, as previously). I'm certainely missing constraints here, but I just wanted to question that before we review the patch more deeply :)

paulboosz:
> Why not keeping only `processes_impacts` and removing `processes_generic_impacts` instead of adding a new file?

As a data team, keeping the 2 files `processes_impacts` and `processes_generic_impacts` is useful as these 2 files have the same processes under different formats (ie food1 and food2). Looking into the merged file can be confusing.

if we'd rename `processes_merged_impacts.json` -> `processes_impacts.json`
then we'd need to rename `processes_impacts.json` -> `processes_legacy_impacts.json` ?

vjousse:
Well, from a data point of view, do as you want 😅 But from a frontend POV, I would like to have only one file containing processes in `public/data/`: `processes.json` (and its companion `processes_impacts.json`)

n1k0:
> from a frontend POV, I would like to have only one file containing processes in public/data/: processes.json (and its companion processes_impacts.json)

I second that 💯

> As a data team, keeping the 2 files processes_impacts and processes_generic_impacts is useful as these 2 files have the same processes under different formats (ie food1 and food2). Looking into the merged file can be confusing.

But as a data team, maybe these two files should live elsewhere than in the public frontend data folder? I could totally see a need for these generic/legacy exports living in `/data`, and the merged one living in `/public/data`. 

What do you guys think?

paulboosz:
> But as a data team, maybe these two files should live elsewhere than in the public frontend data folder? I could totally see a need for these generic/legacy exports living in `/data`, and the merged one living in `/public/data`.
> 
> What do you guys think?

Sounds good I'll do that !

cedricr:
> > But as a data team, maybe these two files should live elsewhere than in the public frontend data folder? I could totally see a need for these generic/legacy exports living in `/data`, and the merged one living in `/public/data`.
> > What do you guys think?
> 
> Sounds good I'll do that !

Yep ; right now all files are duplicated in the /data/ folder, until I finalize the merge. The goal is indeed to keep the files in the format useful for us in /data/, and provide you guys with only the files you need in /public/data/

vjousse (approved):
I can’t tell from a data point of view, but from a frontend POV it looks good!

cedricr (approved):
LGTM 👍

cedricr (changes requested):
Need to fix the latest merge

cedricr (approved):
@paulboosz should be ok, but the rebase was problematic, maybe give it a quick look before merging !

PR #2439, opened 2026-06-10
fix(ui): route to the current URL when the component config is received

ccomb:
## :wrench: Problem

The auth e2e journey fails intermittently in CI (related to #2408), e.g. [this run](https://github.com/MTES-MCT/ecobalyse/actions/runs/27259767298/job/80502370611). When the component config is received, `Main` re-routes to the URL requested at boot time. After login the config lands late (behind the detailed processes download), so the user (or the test) gets redirected back to the landing page, and the elements the test reads are detached from the DOM.

## :cake: Solution

- **fix(ui)**: route to the current URL (`model.url`) instead of the boot-time one.
- **test(e2e)**: wait for the boot chain to settle after a reload before clicking a tab (`reloadAndSettle`), drop a redundant account re-click.
- **test(e2e)**: align timeouts with the slow post-login chain.

The page re-initialization on config arrival is kept as-is (pages must pick up the new db); removing it would be a deeper refactor.

- CI should be green without retries (the auth spec previously failed about one run out of two).
- Locally: `npx playwright test e2e/auth.spec.js` (about 22s).
- On master, with a throttled network: log in, then navigate while data is still loading, and you are redirected back to where you started. With this PR you stay on the current page.

### DIFF
--- a/e2e/auth.spec.js
+++ b/e2e/auth.spec.js
@@ -6,6 +6,7 @@ import {
   extractUrlsFromText,
   loginUser,
   registerAndLoginUser,
+  reloadAndSettle,
 } from "./lib";
 
 import impacts from "../public/data/impacts.json";
@@ -80,14 +81,15 @@ test.describe("auth", () => {
     });
 
     await test.step("browser reload and auto-login", async () => {
-      await page.reload();
+      await reloadAndSettle(page);
 
       await expect(page.getByRole("heading", { name: "Mon compte" })).toBeVisible();
     });
 
     await test.step("api tokens", async () => {
-      await page.getByRole("link", { name: "Mon compte" }).click();
-
+      // We're already on the account page: re-clicking the "Mon compte" link would
+      // re-route to the same page, and the async page re-init would race the tab
+      // selection below and reset it.
       await page.getByRole("button", { name: "Jetons d’API" }).click();
 
       await expect(page.getByText("Aucun jeton d’API actif")).toBeVisible();
@@ -123,7 +125,7 @@ test.describe("auth", () => {
       const apiResponseJson = await apiResponse.json();
       expect(apiResponseJson.impacts.cch).toBeGreaterThan(0);
 
-      await page.reload();
+      await reloadAndSettle(page);
       await page.getByRole("button", { name: "Jetons d’API" }).click();
 
       await expect(apiTokensTable.locator("tbody tr td:nth-child(2)")).not.toHaveText(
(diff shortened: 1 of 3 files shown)


ccomb:
Deployment seems stuck but it's actually deployed **https://ecobalyse-staging-pr2439.osc-fr1.scalingo.io** :roll_eyes:

vjousse:
I have to admit that I’m skeptical about this change, it looks good but, is it really? Does someone understands _for sure_ what the LLM is stating here? (For the future… please, avoid submitting AI written PRs like that, I feel like I’m discussing with a robot and not with you @ccomb, I really don’t like this feeling at all)

n1k0:
I'm gonna put here what I was writing this morning during the standup:

> pour ces histoires d'échec intermittents, le pb c'est de faire travailler un générateur stochastique sur un bug aléatoire non systématiquement reproductible, la probabilité de le faire halluciner est immense pour moi c'est le genre de truc impossible à reviewer ou lander sans faire jouer une centaine de fois la Cl afin de vérifier que sa solution a un impact statistique suffisamment significatif

ccomb:
So to be clear, I don't submit a PR I'm not able to fully understand and justify, or even support myself. The AI discovers issues based on a real CI run failure I encountered (link above), it provides suggestions, analysis, I read them, contradict, suggest alternatives, enforce some rules, try to get the smallest and well-defined changes.
But this is the kind of unpredictable problems that may arise from time to time, depending on the load of the CI, and that will set people in anger, waiting for a PR to be validated...

Here the playwright job may click faster than a human on the app, to check some results (such a <select> list) after navigating to another URL, but then the data asked for during the first load might arrive **after** the navigation on this second URL, and the expected <select> list won't provide the expected outcome.

ccomb:
I could reproduce it outside the CI, directly by clicking in the browser. Here are some screencasts using slight DSL throttling.
A few second after clicking on `Explorateur`, the file `config.json` arrives and I'm redirected to the previous  `/#/auth`. That's what the playwright tests don't like.

# without this PR :

https://github.com/user-attachments/assets/f2c84433-952d-45a4-a2f4-2687f31ef518

# same on the review app:

https://github.com/user-attachments/assets/1220a50f-24b1-4546-be06-81525194a6c4

vjousse:
Because we know have HTTP loading, I think this PR is not relevant anymore, what do you think @ccomb ?

ccomb:
Agreed, closing! Since #2416 the config is fetched after the app is already loaded, and that late receipt is now ignored instead of re-routing, so the redirect can't happen anymore. And clicking on the top menu does nothing during loading anyway

PR #2444, opened 2026-06-10
chore: display alias in explorer

### LINKED ISSUE #2443: l'alias n'est pas visible dans l'explorateur food2
## :wrench: Problème

Pour investiguer sur des incohérences, la méthode (notamment @GLestra15 ) a besoin de voir l'alias des procédés/ingrédients. 
Or cet alias n'est pas visible dans l'explorateur actuellement (notamment dans l'explorateur food2)

## :cake: Solution

Afficher l'alias dans l'explorateur

paulboosz:
Fix https://github.com/MTES-MCT/ecobalyse/issues/2443

- Alias is for now not present for every generic processes but could be in the future
- in old explorer, alias isn't present so its `N/A`

### DIFF
--- a/src/Page/Explore/Processes.elm
+++ b/src/Page/Explore/Processes.elm
@@ -61,6 +61,10 @@ baseColumns detailed scope =
                     a [ Route.href (Route.Explore scope (Dataset.Processes scope (Just process.id))) ]
                         [ code [] [ text (Process.idToString process.id) ] ]
       }
+    , { label = "Alias"
+      , toValue = Table.StringValue <| .alias >> Maybe.withDefault "N/A"
+      , toCell = .alias >> Maybe.map (\alias -> code [] [ text alias ]) >> Maybe.withDefault (text "N/A")
+      }
     , { label = "Nom"
       , toValue = Table.StringValue Process.getDisplayName
       , toCell = Process.getDisplayName >> tooltipedCell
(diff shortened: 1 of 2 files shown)


src/Page/Explore/Processes.elm
```diff
@@ -61,6 +61,10 @@ baseColumns detailed scope =
                     a [ Route.href (Route.Explore scope (Dataset.Processes scope (Just process.id))) ]
                         [ code [] [ text (Process.idToString process.id) ] ]
       }
+    , { label = "Alias"
+      , toValue = Table.StringValue <| .alias >> Maybe.withDefault "N/A"
+      , toCell = .alias >> Maybe.map (\alias -> code [] [ text alias ]) >> Maybe.withDefault (text "N/A")
+      }
```
n1k0: May I suggest we render the alias under the uuid in the same table cell instead? three columns for rendering names feel slightly overkill, especially when many processes don't have an alias and you get so many N/A (in such a case, we wouldn't render anything under the id)

Image

paulboosz: sounds good if its searchable, I'm not sure how to do that tough

n1k0: I can do it if you want :)

paulboosz: that'd be great yes :)

n1k0: done in 146aec2757fe05ffe16db449eab25fcaf759048a

paulboosz: very nice thanks !

n1k0 (approved):

PR #2501, opened 2026-06-18
chore: python is not needed on stable/textile

vjousse:
## :wrench: Problem

`dependabot` is spawning security PRs for the python deps on the stable branch… but we don’t use python at all anymore for the stable version. All the backend calls are routed to the current backend on the master branch.

## :cake: Solution

Remove all python related code

Tests and deployment should be green.


n1k0 (approved):
👍

PR #2535, opened 2026-06-23
refactor: resync transport file and generate offline

### LINKED ISSUE #2331: chore(data): resynchronisation du fichier transports.json
## :wrench: Problème

Le fichier public/data/transports.json n’est plus synchronisé avec le script servant à le générer (data/common/distances/transports.py) ; des éditions manuelles ont eu lieu, en particulier pour hardcoder certaines values (outre-mers, maghreb, etc.)

## :cake: Solution

?

paulboosz:
## :wrench: Problem

Fix https://github.com/MTES-MCT/ecobalyse/issues/2331 and https://github.com/MTES-MCT/ecobalyse/issues/2447 and should unblock https://github.com/MTES-MCT/ecobalyse/issues/2255

transports.json  is an incomplete state as we added manually some countries distances, but a country should have distance to all other countries and its not the case (eg "ROF")

Also regenerating transports.json was not possible as we API we used changed and was not functioning as before

## :cake: Solution

Compute all distances (air, sea, road) offline using documented heuristics (detailed in README)

- to avoid this kind of issues in the future, we should avoid editing `transports.json` manually and regenerate it using `transports.py` (which validate a coherent and complete internal state)

Should resolve
 https://github.com/MTES-MCT/ecobalyse/issues/2447 
  https://github.com/MTES-MCT/ecobalyse/issues/2255


cedricr:
@paulboosz the front-end tests are failing

data/common/distances/build_country_coordinates.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Any
+
+import country_set
+import geonamescache
+
+HERE = pathlib.Path(__file__).parent
+
+
+def largest_city_per_country() -> dict[str, Any]:
+    """Return {countrycode: city dict} for the most populous city of each country."""
```
cedricr: Is the “most populous city” a requirement ? I’d probably rather have used the country centroid and a static dataset (like https://www.data.gouv.fr/datasets/liste-des-pays-et-territoires) instead of a code dependency, but both works 🤷‍♂️

data/common/distances/README.md
```diff
@@ -1,6 +1,49 @@
...
+
+### 1. `build_country_coordinates.py` -> `country_coordinates.csv`
+Each country is represented by the coordinates of its most populous city, e.g. Brazil -> Sao Paulo
+
+### 2. `compute_distances.py` -> `distances_raw.json`
+For every pair of countries (derived from `countries.json` via `country_set`), computes three
+distances (km):
+- **air**: geodesic distance * AIR_CIRCUITY_FACTOR (1.05)
+- **sea**: via python package `searoute`
+- **road**: If two countries are road connected : geodesic distance * ROAD_CIRCUITY_FACTOR (1.25) else `null`
```
cedricr: @nicolaspkandeel was talking about a 1.21 factor according to https://www.plancarbonegeneral.com/approches-sectorielles/produit-generique/fret-amont-et-aval, althought I don’t really understand that link : 
it says that **except for road transportation** you can use the greater circle distance, with a corrective factor, but then gives a corrective factor even for road distance. But it doesn’t make sense to apply this factor to a “real” distance calculated by a routing engine.

And calculating distances between the most populated cities, as we are doing right now, or between the “center” of the country is not the same at all – I’m not sure what’s making the most sense from the methodological perspective.

data/public/data/processes_impacts.json
cedricr, line 1: careful here

data/public/data/processes_impacts_full.json
cedricr, line 1: …and here

tests/Data/Food/RecipeTest.elm
```diff
@@ -390,7 +390,7 @@ suite =
                           }
                             |> Recipe.compute db
                             |> Result.map (firstIngredientDistance .sea)
-                            |> Expect.equal (Ok (Just <| 8315))
+                            |> Expect.equal (Ok (Just <| 7035))
                             |> asTest "should have 8315 boat transport for overseas France ('ROF')"
```
cedricr:
```suggestion
                            |> asTest "should have 7035 boat transport for overseas France ('ROF')"
```

I guess ? that’s a huge change though.

public/data/transports.json
cedricr, line 1: It would be nice to keep the previous formatting on this file so that it is easier to check what has changed.
Maybe just by excluding it from prettier ?

cedricr: (the old format is also much nicer and readable in that case)

data/common/distances/country_set.py
```diff
@@ -0,0 +1,31 @@
...
+
+import json
+import pathlib
+
+REPO_ROOT = pathlib.Path(__file__).parents[3]
+DEFAULT_COUNTRIES_JSON = REPO_ROOT / "public" / "data" / "countries.json"
+
+# region code -> the real country used to represent it
+REGION_PROXIES = {
+    "---": "IN",  # unknown / default
```
cedricr, line 12: maybe add a TODO or FIXME here, indicating that "---” is only for textile/food1 (I think ?), so that we remember to clean it up one day.

data/common/distances/distances_raw.json
```diff
@@ -1,3045 +1,1677 @@
...
-    "CH": {
-      "air": 8857,
-      "road": null,
-      "sea": 8596
-    },
     "CN": {
-      "air": 16580,
+      "air": 19498,
       "road": null,
-      "sea": 25166
```
cedricr, line 669: the changes are huge… are those expected ?

data/common/distances/country_coordinates.csv
```diff
@@ -0,0 +1,27 @@
...
+MQ,Fort-de-France,14.6037,-61.0742,89995
+NL,Amsterdam,52.374,4.8897,741636
+NZ,Auckland,-36.8485,174.7635,1530500
+PE,Lima,-12.0432,-77.0282,7737002
+PK,Lahore,31.558,74.3507,13004135
+PL,Warsaw,52.2298,21.0118,1702139
+PT,Lisbon,38.7251,-9.1498,517802
+TN,Tunis,36.819,10.1658,693210
+TR,Istanbul,41.0138,28.9497,15701602
+US,New York City,40.7143,-74.006,8804190
```
cedricr: Country “center” would feel better in that case, if we need to calculate a US/China transport for ex.

data/common/distances/country_coordinates.csv
cedricr, line 1: We are not considering that many countries ; we could let @nicolaspkandeel for example choose the most pertinent city in each country, if that makes more sense ?

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+
+# Supplementary edges added to account because even if some countries are separated by water, transport is possible by short ferry
+CONNECTIVITY_ADD_EDGES = [
+    ("ES", "MA"),  # Gibraltar strait ferry
+]
+
+CONNECTIVITY_CUT_EDGES = [
+    ("FR", "BR"),  # French Guiana land border, not a mainland-France road route
+    ("FR", "SR"),  # French Guiana - Suriname, idem
+    ("PA", "CO"),  # Darien Gap: no road between North and South America
```
cedricr, line 41: Except for drugs and flesh eating flies, though! 😅

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+import csv
+import json
+import pathlib
+from datetime import datetime
+
+import country_set
+import geopy.distance
+import pycountry
+import searoute
+from countryinfo import CountryInfo
```
cedricr: I can be wrong, but it seems to me we are loading a lot of dependencies that all do more of less the same things ? 
For ex. why do we iterate on pycountry.countries if we have CountryInfo.all_countries, or CountryInfo.all() ?

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+    ("FR", "SR"),  # French Guiana - Suriname, idem
+    ("PA", "CO"),  # Darien Gap: no road between North and South America
+]
+
+
+def load_country_coords(csv_path):
+    """Return {alpha2_code: (lat, lon)} from country_coordinates.csv.
+
+    Each country is represented by its most populous city (see
+    build_country_coordinates.py), more representative of trade flows than the
```
cedricr: **If** the most populous city is the best representative (and I’m not really sure that makes sense), maybe we can use the capital city instead, and just use CountryInfo.capital_latlng instead, and get rid of all the "build_country_coordinates” stuff.
And if the center is best, CountryInfo.latlng will work.

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+    geometric centroid.
+    """
+    with open(csv_path, newline="") as csv_file:
+        reader = csv.DictReader(csv_file)
+        return {row["alpha2"]: (float(row["lat"]), float(row["lon"])) for row in reader}
+
+
+def _alpha3(alpha2_code):
+    """ISO alpha-2 -> alpha-3 (countryinfo borders are expressed in alpha-3)."""
+    country = pycountry.countries.get(alpha_2=alpha2_code)
```
cedricr: CountryInfo can do the conversion too

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+def compute_distances(codes, coords, component_of):
+    """Build {from: {to: {road, sea, air}}} for every unordered pair of codes."""
+    distances = {}
+    for i, code_from in enumerate(codes):
+        distances[code_from] = {}
+        for code_to in codes[i + 1 :]:
+            geodesic_distance = compute_geodesic_distance(code_from, code_to, coords)
+            connected = is_road_connected(code_from, code_to, component_of)
+            distances[code_from][code_to] = {
+                "road": road_distance(geodesic_distance, connected),
```
cedricr: if we know connected is false, there’s no point in calling this method


```suggestion
                "road":  road_distance(geodesic_distance) if connected else None,
```

data/common/distances/compute_distances.py
```diff
@@ -0,0 +1,166 @@
...
+def road_distance(geodesic_km, connected):
+    if not connected:
+        return None
+    return round(ROAD_CIRCUITY_FACTOR * geodesic_km)
+
+
+def sea_distance(code_from, code_to, coords):
+    """searoute sea distance in km, or None if it cannot be computed."""
+    try:
+        route = searoute.searoute(
```
cedricr, line 110: In the future, maybe we will have to define the preferred harbours at least for the biggest countries ; at the moment we have 20 000km between CN and US, that doesn’t seem reasonable at all
(@nicolaspkandeel)

cedricr (changes requested):
Huge work, thanks !
Here are a few comments already.

data/common/distances/compute_distances.py
paulboosz: indeed, i removed pycountry

data/common/distances/README.md
paulboosz: changed to country centroid

data/common/distances/compute_distances.py
paulboosz: changed to country centroid following method recommendation

public/data/transports.json
paulboosz: indeed, old format is prettier so I added this file as exception to .prettierignore

data/common/distances/distances_raw.json
paulboosz: CN-BR sea distance (20 000 ) seems good s
image

public/data/transports.json
cedricr: > old format is prettier so I added this file as exception to .prettierignore

Nice one :rofl:

data/common/distances/compute_distances.py
nicolaspkandeel: Indeed, but it is not a priority. We should wait for feedbacks from users before to work on this.

nicolaspkandeel (approved):

data/common/distances/country_set.py
```diff
@@ -0,0 +1,32 @@
...
+    "REM": "CZ",  # Europe and Maghreb
+}
+
+
+def distance_country_codes(countries_json_path):
+    """Real-country codes to compute distances for, derived from countries.json"""
+    with open(countries_json_path, encoding="utf-8") as f:
+        official_countries = json.load(f)
+    return sorted(
+        {REGION_PROXIES.get(c["code"], c["code"]) for c in official_countries}
```
cedricr, line 31: What are you trying to do here ? Not sure I understand the c["code"] duplicate.

cedricr (approved):
much simpler, nice ! :clap:

data/common/distances/country_set.py
paulboosz: if the code is in REGION_PROXIES (eg "REO"), get the real country (eg "ES“) else (eg "FR") keep the original code

PR #2393, opened 2026-06-03
fix: use JSON to store components

vjousse:
fixes #2386

PR #2536, opened 2026-06-24
chore: add a migration to resynchronise DB and models

cedricr:
## :wrench: Problem

DB and models are not synchronized anymore. 
The `component`.`comment` field which was removed from the model in https://github.com/MTES-MCT/ecobalyse/pull/2393 was not properly registered in the migration script.

## :cake: Solution

Run `uv run backend database make-migrations`


vjousse (approved):
Thanks!

PR #2561, opened 2026-06-28
chore: replace elecMJ by elecKwh

### LINKED ISSUE #1417: Elec is expressed in MJ instead of kWh
## :wrench: Problem

In processes.json, electricity is expressed in MJ through the elecMJ field.

In user facing context (API, UI) we display electricity in kWh, as expected.

While MJ is a valid energy unit,  by convention electricity should be expressed in kWh for better clarity and industry standard compliance.

## :cake: Solution

- rename `elecMJ` to `elecKWh`
- change MJ to kWh
- update all numerical values (divide by 3.6)
- update and adapt all code using this value
- update json schema 

- [ ] Tests should pass
- [ ] There should be no impact difference

cedricr:
## :wrench: Problem

Fixes #1417


data/ecobalyse_data/computation.py
```diff
@@ -336,7 +336,7 @@ def compute_impacts(
 
             unit = fix_unit(bw_activity.get("unit"))
 
-            # WARNING assume remote is in m3 or MJ (couldn't find unit from COM intf)
+            # WARNING assume remote is in m3 or kWh (couldn't find unit from COM intf)
```
cedricr, line 339: @paulboosz not sure about the meaning of this comment ; let me know if updating it was wrong

cedricr:
@vjousse or @n1k0: can you check the elm part ? :pray: 
@paulboosz @ccomb: the processes in data/lci_catalog were updated by running `./data/migrations/2026-06-22-use-kwh-instead-of-mj.py`. Its idempotent, so you might have to run it again on your branches if you’re adding new processes.

n1k0 (approved):
The Elm code looks absolutely fine to me, great work 🎉

data/ecobalyse_data/computation.py
paulboosz: comment is from @ccomb, not sure

paulboosz (approved):
No impact change, looks good !

PR #2600, opened 2026-07-06
fix: prevent forging auth token

n1k0:
## :wrench: Problem

This fixes a security report we'ge got on YesWeHack (see thread on Tchap)

## :cake: Solution

Ensure token email matches the one of the user in the db

I've added a regression test

### DIFF
--- a/backend/src/app/domain/accounts/controllers/access.py
+++ b/backend/src/app/domain/accounts/controllers/access.py
@@ -179,8 +179,6 @@ class AccessController(Controller):
     ) -> None:
         """Validate a token"""
 
-        user_email = None
-
         cache_duration = settings.app.DEFAULT_TOKEN_VALIDATION_CACHE_SECONDS
 
         if cache_duration:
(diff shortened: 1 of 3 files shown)


vjousse:
I would love more context about this patch please. What thread on Tchap are you talking about? I’m really surprised to discover this through a Github issue.

n1k0:
@vjousse this is an issue I was tagged on YesWeHack and confirmed for investigation by the betagouv account there. I reported about it on Tchap for security reason, and because [we're supposed to have conversation there now](https://mattermost.incubateur.net/fabnum-mte/pl/fpmuwy13w7nkz8wjqeg6fc6c9e) (Mattermost is deprecated). I can't really disclose the security report publicly in the comments here, and we don't have a configured way to have private conversations on Github about security (I don't have a clue if that's even possible, but I see that the "Security" tab and admin section isn't configured at all on this repo, so there's that.

Edit: link to the report on YesWeHack https://yeswehack.com/vulnerability-center/reports/825287

vjousse (approved):
Thanks for the fix 🙏

n1k0:
@vjousse I'm landing this, I'll port the patch to stable/textile next 👍

n1k0:
@vjousse scratch my last comment, stable/textile isn't impacted as it uses prod sso

PR #2607, opened 2026-07-06
chore: set 0.1% min diff for diff table

### LINKED ISSUE #2606: lors de l'export le tableau affiche des diffs de 0.0%
## :wrench: Problème

lors de l'export le tableau affiche des diffs de 0.0% ce qui rend le log difficilement lisible

## :cake: Solution

Augmenter le seuil de diff

paulboosz:
## :wrench: Problem

fix #2606 

## :cake: Solution

raise the diff min to 0.1%


ccomb (commented):
IMO we really don't care about changes lower than a few percents.

Just by changing the method version (adapted 1.0, 1.03, 1.05) or origin (adapted from simapro or ILCD from JRC) or by fixing biosphere mapping errors we already get significant changes such as 20% or up to 70% with regionalization tweaks.

However I admit I'm unable to define precisely  "a few percents" :-(

cedricr (approved):
Nice, thanks !

Note: you have to write “Fix #XXX” instead of just “XXX” in the description so that the PR is properly linked to its issue.

PR #2609, opened 2026-07-07
chore: move score_history to GH cron

vjousse:
## :wrench: Problem

Deploys are blocked by the failing score history action that takes too long.

## :cake: Solution

As a first step, remove score history from the required checks and run it every hour or on demand.

Merge and it should fix deploys

### DIFF
--- a/.github/workflows/score_history.yml
+++ b/.github/workflows/score_history.yml
@@ -6,6 +6,11 @@ on:
     - cron: "0 * * * *"
   workflow_dispatch:
 
+# prevent overlapping runs
+# see https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
 
 env:
   NODE_ENV: test
--- a/.github/workflows/score_history.yml
+++ b/.github/workflows/score_history.yml
@@ -1,9 +1,9 @@
 name: Score History
 
 on:
-  push:
-    branches:
-      - '*'
+  schedule:
+    # Every hour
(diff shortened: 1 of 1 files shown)


cedricr (approved):
Not sure how to test before merging, so let’s merge and then test ! 🎉

n1k0 (commented):
I suggest adding this to serialize runs so an in-flight computation run is never overlapped by the next one

```yaml
# prevent overlapping runs
# see https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
concurrency:
  group: ${{ github.workflow }}
  cancel-in-progress: false
```

n1k0 (approved):

cedricr:
> I suggest adding this to serialize runs so an in-flight computation run is never overlapped by the next one
> 
> ```yaml
> # prevent overlapping runs
> # see https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
> concurrency:
>   group: ${{ github.workflow }}
>   cancel-in-progress: false
> ```

Not fully sure about that; if we merge 3 PRs about at the same time, score_history will run for only one of them, won’t it?

@paulboosz I’m not sure how the results are used: do you need to have the score stored for each PR, or is one from time to time enough ?

n1k0:
> Not fully sure about that; if we merge 3 PRs about at the same time, score_history will run for only one of them, won’t it?

it's covering the case when a job is running for more than 1 hour

also this is a scheduled job and not dependent on the number of prs landed; so yes this patch implies we may not have a score entry for some prs if many have landed in a single hour tilmeframe (and that's independent from the change I suggest)

PR #2611, opened 2026-07-07
feat: add transport production cooling boolean tag

paulboosz:
## :wrench: Problem

Fix https://github.com/MTES-MCT/ecobalyse/issues/2002


n1k0:
The `transport_production_cooling:true` tag name and format is a little strange; how about simply `transported_cooled`?

Note: WRT bool vs tag, a tag can always be treated like a boolean by checking its presence or not; present: true, absent: false. So in code that would be `if (tags.includes("transported_cooled"))`

paulboosz:
> The `transport_production_cooling:true` tag name and format is a little strange; how about simply `transported_cooled`?

the issue talks about `transport_production_cooling`. I remember the `production` in `transport_production_cooling` was added following the workshop on  9/6/2026. Indeed this issue this only refer to the transport during the production step. In the later steps the user is free to select a normal transport or a cooling transport.

> Note: WRT bool vs tag, a tag can always be treated like a boolean by checking its presence or not; present: true, absent: false. So in code that would be `if (tags.includes("transported_cooled"))`

Indeed. It seems you know what you want so I suggest you just create the tag that suits you and I'll update the data pipeline accordingly

n1k0:
> Indeed. It seems you know what you want so I suggest you just create the tag that suits you and I'll update the data pipeline accordingly

Will do, thank you!

n1k0:
J'ai créé une PR qui utilise le tag `transported_cooled` ici #2616 :)

n1k0:
#2616 has landed onto master :)

PR #2633, opened 2026-07-13
chore: move e2e test suite to a scheduled job

### LINKED ISSUE #2408: Déplacer les tests e2e de la CI dans un cronjob GitHub
## :wrench: Problème

Les tests playwright sont très lents, et finissent très fréquement en timeout. => Il faut attendre (jusqu’à 1h !) pour pouvoir tester une PR, et parfois même les relancer manuellement.

## :cake: Solution

Les déplacer dans un cronjob de GitHub

n1k0:
## :wrench: Problem

CI runs & builds are very long, and it's inconvenient. Most of the CI build time is spent running the e2e test suite.

## :cake: Solution

Move the e2e test suite to a scheduled CI job, like we do for score history, so that CI builds on PRs are faster.

I don't have a clear idea how we could test the whole thing before landing… hints?

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,7 +3,7 @@ name: End-to-end tests
 on:
   schedule:
     # Every hour
-    - cron: "0 * * * *"
+    - cron: "40 * * * *"
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,7 +29,7 @@ jobs:
     services:
       postgres:
         # Same as Scalingo
-        image: postgres:14
+        image: postgres:17.9
         env:
           POSTGRES_DB: ecobalyse_test
           POSTGRES_USER: ecobalyse
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,7 +1,14 @@
-# This workflow will do a clean install of node dependencies, cache/restore them, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+name: End-to-end tests
 
-name: Node.js CI
+on:
+  schedule:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,106 @@
+name: End-to-end tests
+
+on:
+  schedule:
+    # Every hour
+    - cron: "0 * * * *"
```
cedricr: Maybe don’t run it right on the hour, as we already have score_history running at that time.

```suggestion
    - cron: "40 * * * *"
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,106 @@
...
+
+jobs:
+  build:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        # Same as Scalingo
+        image: postgres:14
```
cedricr: we are now on 17.9 on Scalingo (so we should probably update the docker-compose.yml file too 😅)

cedricr (approved):
👍 looks fine to me, that will help a lot!

The only way to test I can imagine would be to add a dummy test in this PR, for ex. checking that a given text appears on the home page, and create and merge a new PR that changes this text 🤷‍♂️

.github/workflows/e2e.yml
n1k0: I think we're kinda safe as they're ran in separate containers

n1k0: I've applied the suggested change, never too sure 😅

cedricr: yep I was thinking more in term of possible Github limits (too much CPU at a certain time, etc.).

n1k0:
@cedricr the [action runs as scheduled and reports indicate tests are played](https://github.com/MTES-MCT/ecobalyse/actions/workflows/e2e.yml), and PR builds are faster, I'm pretty confident this setup is the right one 🎉

PR #2636, opened 2026-07-13
feat(generic,ux): localize new transforms with sensible defaults

n1k0:
## :wrench: Problem

⚠️ there's no matching issue/card, this is an UX-oriented initiative from me

In the simulator UI, adding a new transformation process doesn't localize it as the material or previous transform step country; from a UX perspective, this slows down the process of creating simulations.

## :cake: Solution

When picking a transform process, automatically apply the country from either the material or the previous transform one, eg.:

1. I pick `Ail bio`, which has a default origin `France`
2. I add a transform and select `Cuisson des fruits et légumes`
3. The added transform is automatically localized in `France` by default

On a generic calculator (food2/object/veli):

Scenario 1:

- Add a component (=ingrédient/matériau)
- edit its first element
- either leave the default origin country if any, or pick one
- add a transform: the transform should be localized as the material

Scenario 2:

- Add a component (=ingrédient/matériau)
- edit its first element
- either leave the default origin country if any, or pick one
- add a transform, ensure it's localized
- add a new transform: the added transform should be localized the same as previous one

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -442,15 +442,15 @@ addElement targetItem material items =
             |> Ok
 
 
-addElementTransform : TargetElement -> Process -> List Item -> Result String (List Item)
-addElementTransform targetElement transform items =
+addElementTransform : TargetElement -> Maybe CountryCode.Code -> Process -> List Item -> Result String (List Item)
+addElementTransform targetElement maybeCountry transform items =
     if not <| List.member Category.Transform transform.categories then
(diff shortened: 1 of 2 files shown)


nicolaspkandeel (approved):
Good UX improvement !

PDagras:
Bien vu @n1k0 ! 👏 
Ca va effectivement simplifier la réalisation de simulations en gagnant quelques clics.

PR #2580, opened 2026-07-01
chore: use ecobalyse-data DB for score_history

vjousse:
fixes https://github.com/MTES-MCT/ecobalyse/issues/1790

Tested on this action run, it works as expected: https://github.com/MTES-MCT/ecobalyse/actions/runs/29958309025/job/89052860981

## Steps

Get dump from scalingo (`0260722193041_ecobalyse_9678.pgsql` here) and import it locally.

```bash
./bin/reset-docker-db.sh
export SCALINGO_DATABASE_URL="postgresql://ecobalyse@localhost:5433/ecobalyse_dev" 
pg_restore --clean --if-exists --no-owner --no-privileges --no-comments --dbname "${SCALINGO_DATABASE_URL}" 20260722193041_ecobalyse_9678.pgsql
```

Extract `score_history` table to a file `score_history.sql`

```bash
pg_dump --host localhost --port 5433 --username ecobalyse --format plain --verbose --file score_history.sql --table public.score_history --no-owner ecobalyse_dev
```

Open the tunnel to the `ecobalyse-data` PG

```bash
scalingo -a ecobalyse-data db-tunnel DATABASE_URL
```

Connect to the DB and drop the `score_history` table if it exists (password can be found in the connection string on the scalingo PG dashboard)

```
psql -h localhost -p 10000 -U ecobalyse_d_5689
drop table score_history;
```

Load the sql file:

```bash
psql -h localhost -p 10000 -U ecobalyse_d_5689 -1 -f score_history.sql 
```

## TODO

Once merged, remove `score_history` table from the `ecobalyse` PG database.

PR #2642, opened 2026-07-13
chore: speed up score history

paulboosz:
## :wrench: Problem

#1790 

- score_history was disabled because it kept timing out
- 
## :cake: Solution

- Speed up  `score_history.py` by 
  - don't download all the full table -> download only what's needed using sql queries
  - don't compute 1200+ ingredient scores using the API -> use processes_impacts.json directly (as a result we can't store transport/distribution impacts)
- add a README with directions to run `score_history` locally 
  - add a --dry-run flag

-> Duration goes from 30+min to 2min 30 🎉🎉

https://github.com/MTES-MCT/ecobalyse/actions/runs/29266417593

- add score_history back in CI 

- this doesnt move score_history table to ecobalyse-data db, this is https://github.com/MTES-MCT/ecobalyse/pull/2580

- you can test in local by following the README


bin/score_history/README.md
```diff
@@ -0,0 +1,65 @@
+# Score history
+
+Tracks the evolution of score over time. For every new commit to master
+scores are computed for textile examples and food ingredients.
+If the score changes, the new rows are appended to the score_history PostgreSQL
+table.
+
+
+## How to run in local
```
cedricr:
```suggestion
## How to run locally
```

bin/score_history/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+uv run --group score_history python bin/score_history/score_history.py \
+    http://localhost:8001 \
+    "$(git branch --show-current)" \
+    "$(git rev-parse HEAD)" \
+    "$SCORE_DB_URL"
+```
+
+Running in local will insert to the production database.
```
cedricr: I’d rather write the command line including the `--dry-run` flag. And let it at that ; there should be no reason, except for testing this PR to write to the prod db directly…

cedricr (approved):
Very nice speedup, congrats!

I’ve skimmed the changes in score_history.py, but the approach looks good.

PR #2653, opened 2026-07-15
feat(ui,api): provide authenticated api commands

n1k0:
## :wrench: Problem

When we recently made the food1 api requiring authentication to be accessed, we didn't update the generated sample commands in the "Share/Partager > API" section of the sidebar.

Even authenticated on the web app, copying and using the provided command would always fail with this message:

```
$ curl -sS -X POST https://ecobalyse.beta.gouv.fr/api/food \
  -H "accept: application/json" \
  -H "content-type: application/json" \
  -d '{"ingredients":[{"id":"f55d6582-d0c6-5987-8171-c6d3a9e448b3","mass":100}]}'
{"error":{"authorization":"Un token valide est requis pour utiliser l’API"},"documentation":"https://ecobalyse.beta.gouv.fr/#/api"}%
```

Frustrating, to say the least.

## :cake: Solution

Update all the provided API sample curl commands in the UI to use auth token when users when they're signed in. That way, commands they copy will work in their tooling.

### Before

image

### After

Note the added `Authorization` header:

image

I pondered if that could increase risks of leaking the token, but quickly realized a token can already be sent by email or over chat, so… I don't think this patch adds much to the existing risk (which is very low)

Check and test copyable API curl commands in all scopes, in anon and authenticated modes.

### DIFF
--- a/src/Data/Food/Query.elm
+++ b/src/Data/Food/Query.elm
@@ -35,6 +35,7 @@ import Data.Food.Ingredient as Ingredient
 import Data.Food.Preparation as Preparation
 import Data.Food.Retail as Retail
 import Data.Process as Process exposing (Process)
+import Data.Text as Text
 import Json.Decode as Decode exposing (Decoder)
 import Json.Decode.Pipeline as Pipe
 import Json.Encode as Encode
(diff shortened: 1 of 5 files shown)


paulboosz (approved):
Works great !

PR #2657, opened 2026-07-16
feat(data): add transported cooled tags

paulboosz:
## :wrench: Problem

#2656

## :cake: Solution

- add transported_cooled tag to relevant ingredients
- remove obsolete transport_cooling attribute from generic


data/common/infer_metadata.py
```diff
@@ -18,6 +18,37 @@
...
+def infer_transported_cooled(categories: List[str]) -> List[str]:
+    """add transported_cooled tag to materials with TRANSPORTED_COOLED_MATERIAL_TYPES"""
+    material_types = {
+        category[len(_MATERIAL_TYPE_PREFIX) :]
+        for category in categories
+        if category.startswith(_MATERIAL_TYPE_PREFIX)
+    }
+    is_ingredient = "ingredient" in categories
+    is_perishable = (
+        bool(material_types & TRANSPORTED_COOLED_MATERIAL_TYPES) & is_ingredient
```
n1k0: I'm a bit confused as to why we use the bitwise `&` operator here… EDIT: ok, the first one for the frozen set might make sense, but the second one would deserve to become a regular `and` :)

data/common/infer_metadata.py
```diff
@@ -18,6 +18,37 @@
...
 
 
+TRANSPORTED_COOLED_MATERIAL_TYPES = frozenset(
+    {
+        "fruits_and_vegetables",
+        "fish_and_shellfish",
+        "legumes",
+        "red_meats",
+        "poultry",
+        "offal",
```
n1k0, line 28: silly question but I prefer asking: no dairy products, like cheese or milk? do eggs fall under the `poultry` category? maybe these products which are usually transported cooled will be handled later? I'm asking just to be sure :)

data/common/infer_metadata.py
```diff
@@ -18,6 +18,37 @@
...
+    }
+)
+
+TRANSPORTED_COOLED_CATEGORY = "transported_cooled"
+_MATERIAL_TYPE_PREFIX = "material_type:"
+
+
+def infer_transported_cooled(categories: List[str]) -> List[str]:
```
n1k0, line 36: Nit: maybe this fn would deserve a unit test, especially as it's pure AFAICT :) your call, it's data land :)

n1k0 (approved):
The patch works great! just a few nits and remarks

data/common/infer_metadata.py
paulboosz: Thats a good point. The origin issue didn't deal with cheese and milk. I can see cheese and milk have `material_type:other_food_items` so we can't discriminate them right now but I think we should.

I created an issue : https://github.com/MTES-MCT/ecobalyse/issues/2661

data/common/infer_metadata.py
paulboosz: good idea ! done

PR #2663, opened 2026-07-20
feat(data): infer rawToCookedRatio

### LINKED ISSUE #2662: le rawToCookedRatio est dupliqué manuellement dans chaque `lci_catalog/*.json`
## :wrench: Problème

Le rawToCookedRatio est actuellement défini dans chaque`lci_catalog/*.json` :

```
   "rawToCookedRatio": 0.856,
      "rawToCookedRatioMatch": {
        "confidence": 0.95,
        "rule": "Matched with artichoke in cooked_to_raw.csv"
      },
``` 

Or cette valeur provient dans la majorité du temps d'un ensemble [de règles d'agribalyse ](https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire/impacts-consideres/rapport-cru-cuit).

Ces valeurs renseignés manuellements créent du bruit, ne rendent pas ces règles transparentes, facilement modifiables, et ajoute du travail lors de l'ajout d'ingrédients

## :cake: Solution

Déplacer cette logique dans la pipeline data. Inférer le rawToCookedRatio à partir du material_type (fruits,céréales,...) sauf quand cela ne correspond pas. Dans ce cas là surcharger la valeur par un ratio explicite.

paulboosz:
## :wrench: Problem

Fix #2662 

## :cake: Solution
- to re-migrate `lci_catalog/*.json` run `uv run --script data/migrations/2026-07-15-infer-raw-to-cooked-ratio.py`

### DIFF
--- a/data/common/infer_metadata.py
+++ b/data/common/infer_metadata.py
@@ -2,7 +2,7 @@
 
 import functools
 import json
-from typing import List, Optional, Tuple
+from typing import List, Optional, Set, Tuple
 
 from config import DATA_ROOT_DIR
 
@@ -15,6 +15,21 @@ _LEGACY_ZONE_TO_COUNTRY = {
     "EuropeAndMaghreb": "REM",
     "OutOfEuropeAndMaghreb": None,
     "OutOfEuropeAndMaghrebByPlane": None,
+}
+
+# Reference values from
+# https://fabrique-numerique.gitbook.io/ecobalyse/alimentaire/impacts-consideres/rapport-cru-cuit
+# `material_type:other_food_items` is absent cause it
+# heterogenous rawToCookedRatio values
+_MATERIAL_TYPE_TO_RAW_TO_COOKED_RATIO = {
+    "cereals": 2.259,
+    "eggs": 0.974,
+    "fish_and_shellfish": 0.819,
+    "fruits_and_vegetables": 0.856,
+    "legumes": 2.33,
+    "offal": 0.730,
+    "poultry": 0.755,
+    "red_meats": 0.792,
 }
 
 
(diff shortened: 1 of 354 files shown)


data/migrations/2026-07-15-infer-raw-to-cooked-ratio.py
```diff
@@ -0,0 +1,65 @@
...
+            if len(material_type_tags) == 1
+            else None
+        )
+
+        modified = False
+        for metadata in activity.get("metadata", []):
+            if "rawToCookedRatio" not in metadata:
+                continue
+            if metadata["rawToCookedRatio"] == inferred_ratio:
+                del metadata["rawToCookedRatio"]
```
cedricr, line 42: why `del` for the first one and `pop` for the second ? In case `rawToCookedRatio` exists but not `rawToCookedRatioMatch` ?

cedricr (approved):
Great !

data/common/infer_metadata.py
```diff
@@ -67,6 +86,25 @@ def infer_default_origin(
...
+    if explicit_ratio is not None:
+        return explicit_ratio
+
+    material_types = get_material_types(categories)
+    if len(material_types) == 1:
+        ratio = _MATERIAL_TYPE_TO_RAW_TO_COOKED_RATIO.get(next(iter(material_types)))
+        if ratio is not None:
+            return ratio
+
+    raise ValueError(
```
ccomb, line 102: When this raises, the message names the `material_type` tags but not the ingredient, so a contributor gets `Cannot infer rawToCookedRatio from material_type tags {'other_food_items'}` with no idea which catalog entry is at fault. You may pass the alias in (both call sites already have it) and add it to the message.

ccomb (changes requested):

data/migrations/2026-07-15-infer-raw-to-cooked-ratio.py
paulboosz: yes

data/common/infer_metadata.py
paulboosz: ok, reraised the error with alias in the parent function

ccomb (approved):
This PR goes in the right direction, but it holds a smaller category size than previous ones so it will need follow-up PRs to get finer/additional ones, and reduce the hardcoded (or ML-infered) metadata values. But still wondering how we can add thousands of new ingredients in the future and determine the material_type anyway 🤔

cedricr:
> This PR goes in the right direction, but it holds a smaller category size than previous ones so it will need follow-up PRs to get finer/additional ones, and reduce the hardcoded (or ML-infered) metadata values. But still wondering how we can add thousands of new ingredients in the future and determine the material_type anyway 🤔

BTW, where does `ingredientCategories` comes from ? 

Also : if needed, and if is very simple for a person to know the right category, we can always create a tool to help classifying manually. It won’t scale to 100s of thousands ingredients, but a few thousands can easily be done.

PR #2666, opened 2026-07-21
refactor(data): simpler, DRY processes detailed impacts

n1k0:
## Problem

The data in `processes_impacts.json` are mostly duplicated from `processes.json`, impact details are "simply" added on top of all contents from this source file. This is not DRY nor optimal from a performance and a maintenance perspective.

```
$ ls -la public/data/processes*
-rw-r--r--  1 niko  staff   8.2M Jul 21 18:03 public/data/processes_impacts.json
-rw-r--r--  1 niko  staff   6.9M Jul 20 17:11 public/data/processes.json
```

## Solution

This patch simplifies the format of `processes_impacts.json`, so we only have a list of process ids and theiur detailed impacts:

```json
[
  {
    "id": "007db9be-fbb0-5c2f-ac8a-4780bc4a5304",
    "impacts": {
      "acd": 0.1234,
      (…)
    }
  },
  {
    "id": "83a793c1-b053-5293-b5b4-94a52490f9bb",
    "impacts": {
      "acd": 1.2345,
      (…)
    }
  },
```

> ⚠️ note: impact details are still encrypted

Then we only load this file and merge it with base processes, in two different steps, so we don't download most of the data twice.

The weight reduction is encouraging:

image

```
$ ls -la public/data/processes*
-rw-r--r--  1 niko  staff   1.6M Jul 21 18:04 public/data/processes_impacts.json
-rw-r--r--  1 niko  staff   6.9M Jul 20 17:11 public/data/processes.json
```

I've also added integrity checks, and a script to migrate and slim down the initial data in the legacy `processes_impacts.json` file.

## Points to watch/comments

- I had to update the ingredient hierarchy check script to use the new format
- same thing for the backend fixtures loading script one

## How to test?

Use the web app and the API in both anonymous and authenticated mode, check for no regressions, especially when accessing detailed impacts (cch, acd, etc). A green build with tests passing is super reassuring though.

## What to do next

The data pipeline should be updated to export processes impact details in the new format. I can probably take care of this, though I prefer asking :) what do you think @cedricr @paulboosz @vjousse?


n1k0:
I'd like to gather feedback from @paulboosz @cedricr @vjousse as to if this approach looks viable or not :)

cedricr:
> I'd like to gather feedback from @paulboosz @cedricr @vjousse as to if this approach looks viable or not :)

I don’t know well enough how the front-end manages its startup to have an opinion on the approach. Naively, with the current state of the json files I would that say : if the users are authenticated, you only load the processes_impact.json file ; if they are not, you only load the processes.json file, so having both files complete would be a plus. The only problem with that is that the first time somebody access the app, they will first get the public file, until the moment they are authentified at which time they will have to request the second one. That’s a problem, granted, but the files are hopefully transmitted compressed, so the multi-MB sizes are only a few hundred ks over the wire.

But again, there are probably good reasons why this approach is not satisfying, and it isn’t a huge cost on the data side to change the format of the impacts output file, so up to you! This should probably wait on https://github.com/MTES-MCT/ecobalyse/issues/2410, on which I’m working on right now though.

vjousse (approved):
Great move! Nothing to say regarding the code and I tested it locally and couldn’t spot any regression.

paulboosz (commented):
Nice ! I think it makes sense to split the data this way

I have one concern, its about ease of reviewing impact diff. It's link to https://github.com/MTES-MCT/ecobalyse/issues/2410 so I'm not sure if its relevant here.

But I think its important that when we are reviewing impact change the whole process context (displayName, alias, metadata,...) is present as it's necessary to understand impact change. 

But I think we can export individual files per process with full context (https://github.com/MTES-MCT/ecobalyse/issues/2410) AND export another format like this one for the front, right @cedricr ?

cedricr:
> But I think we can export individual files per process with full context (#2410) AND export another format like this one for the front, right @cedricr ?

Yep, I think after #2410, the /public/data/processes_impact.json will not even be kept in version control, and will just be generated at build time. We’ll have the full diffs on atomic files, IF transcrypt can manage thousand of files.

n1k0:
I'm closing this PR, as folks have made fair points against landing it IMHO:

- json data files are loaded gz compressed over the wire, so the actual gain isn't as important as stated in the description
- also, the real loading happen the first time the user loads the app, cached files are served as 304 after that
- we were already loading the right processes file according to auth status, so the only case we were actually reloading most of the data is when logging in and out, which is not happening often
- the new format makes seeking detailed impact for a given process a pain, as one has to search for the id instead of just part of the name or any other meaningful piece of text as previously
- processes_details.json is a generated view over the data, it's not the actual source of truth, so we shouldn't apply the very best practices on a generated result, but rather on the actual db (currently the individual static files in `data/`)
- the patch introduces new complexity, with dataset merges happening here and there, while the cognitive burden is already high on this project

This was a "fun" spike to work on, but it's not worth landing as it brings more cons than pros.

PR #2680, opened 2026-07-27
chore: don’t duplicate processes schema

### LINKED ISSUE #2637: Unifier les json schemas front et data
## :wrench: Problème

Certains schemas de donnée sont dupliqués coté front et data (et backend ?)

- les inventorier
- les regrouper dans un répertoire commun à la racine

vjousse:
fixes #2637


cedricr (commented):
- I’d prefer to have all schemas moved to /data/schema (or directly in a /schema/ folder at the root of the project, to indicate that it’s shared by everything?)
- We could now standardize on the schemas testing tool; we can probably get rid of ajv, and replace it with `check-jsonschema` (see check-lci-catalog or check-processes in .justfile)
- /data//tests/processes-schema.json and /data/schemas/lci-schemas.json are very similar ; I don’t have the exact differences in mind, but I think one is just an array of the other, so maybe we can define it like that (I think we can import a schema in another one, but not sure all the tooling is able to read that)

package.json
```diff
@@ -19,7 +19,7 @@
     "encrypt": "./bin/encrypt",
     "db:build": "./bin/build-db && npm run db:check && npx elm-format --yes src/Static/Json.elm",
     "db:check": "elm make src/CheckDb.elm --optimize --output=check-db-app.js 1> /dev/null && node check-db.js",
-    "db:validate": "npx ajv validate --spec=draft2019 -c ajv-formats -s tests/processes-schema.json -d 'public/data/**/processes**.json'",
+    "db:validate": "just check-processes",
```
cedricr: I *think* that’s the good direction (slowly replacing the npm scripts by just), but if that’s “too fast”, we could just use the `uv` cmd here, and avoid the dependency on just inside npm.

cedricr:
```suggestion
    "db:validate": "just check-front-data",
```
no ?

schemas/lci-catalog-schema.json
cedricr, line 1: nit: I would have kept the original filename (lci-schema).
lci-catalog is a folder containing lci files ; this is the schema corresponding to one lci file format.

cedricr: metadata and children are missing in the validation, but that’s expected for the moment, right ?

schemas/lci-catalog-schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "$schema": "https://json-schema.org/draft/2019-09/schema#",
+  "properties": {
+    "categories": {
+      "items": {
+        "type": "string"
+      },
+      "type": "array"
+    },
+    "displayName": {
+      "type": ["string", "null"]
```
cedricr: that’s not new, but I think it should be only "string", and maybe just not required. I don’t think a “null” displayName makes sense

schemas/lci-catalog-schema.json
vjousse: Well, I kept what was already present in the files, no modification for now, just refactoring. I think we should improve the schemas coverage in another PR.

schemas/lci-catalog-schema.json
vjousse: Same here, I think we should improve the schemas quality and coverage in another PR

schemas/lci-catalog-schema.json
cedricr: But they were present in the original file, that’s what I meant.

schemas/lci-catalog-schema.json
cedricr: Indeed! Just talking out loud :grinning:

schemas/lci-catalog-schema.json
vjousse: Good catch, I’ve added `scopes` back too

n1k0 (approved):
Great move!

cedricr (approved):
🎉

PR #2675, opened 2026-07-23
chore: upgrade to elm 0.19.2

ccomb:
## :wrench: Problem

We ought to honor and celebrate the new release :-)

## :cake: Solution

upgrade to elm 0.19.2

## :rotating_light:  Points to watch/comments

The good thing is there are 5 platforms supported by default now

## :desert_island: How to test

Run at least client tests

PR #2705, opened 2026-08-04
chore: upgrade elm to 0.19.2

n1k0:
Supersedes #2675


n1k0:
Hmm nope… we're still blocked by https://github.com/parcel-bundler/parcel/pull/10340 😭

PR #2713, opened 2026-08-06
fix: json serialization for jsonb

vjousse:
## :wrench: Problem

There is an error in the component admin when trying to load components.

## :cake: Solution

When moving to Pydantic, I overlooked the serialization of the jsonb fields. When they are no Pydantic models to serialize too, we should still decode the json string to an object (that was the cause of the error: Elm was getting a string instead of a dict).

I’ve simplified some tests too and I’ve properly implemented the string serialization of the `JournalEntry` value (as it should be a dict but serialized as a string for the Elm frontend)

Go to the Admin in the components section : it should display the list of components as expected.

### DIFF
--- a/backend/src/app/config/base.py
+++ b/backend/src/app/config/base.py
@@ -39,7 +39,7 @@ class UUIDEncoder(json.JSONEncoder):
         return json.JSONEncoder.default(self, obj)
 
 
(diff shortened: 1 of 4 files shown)


cedricr (approved):
The changes look good (and the tests are much nicer that way!), and the admin doesn’t throw an error, nice one!

camcoq-start (approved):

PR #2721, opened 2026-08-07
refactor: explicit genericity wherever applicable

n1k0:
This is mainly a large architectural refactor, aiming at having consistency across the whole codebase handling generic scopes and associated features (simulator, route parsing, examples, explorer etc. for food2, object and veli). It cleans up a bunch of tech debt and todos we had laying around for months, maybe years 🥰

The main idea is to disambiguate *"Object"* as a scope name (currently used for amenity) from *"Object"* as generic simulation for basically anything, including amenity, food and vehicles, by introducing the *"Generic"* name/prefix for the latter.

The patch also adds an examples explorer for food2, and add a new computed column featuring the score per 100g of product per simulation for all generic scopes, as we had with food1.

### :rotating_light:  Points to watch/comments

- I resisted the urge to rename `Scope.Object` to `Scope.Furnishings` as I felt like this would require a kind of strategic approval first 

### :desert_island: How to test

- This is a rather large refactor, touching the fundations of the house, so… Apart from testing the whole app manually, I guess we'll have to heavily rely on our test suite here 😅
- Check the new added examples explorer for food2, and the cost per 100g impl.

### DIFF
--- a/src/Page/Explore/ProductCategories.elm
+++ b/src/Page/Explore/ProductCategories.elm
@@ -13,9 +13,15 @@ import Route
 
 table : Session -> { detailed : Bool, scope : Scope } -> Table ProductCategory String msg
 table { db } { scope } =
+    let
+        genericScope =
+            scope
+                |> Scope.toGenericScope
+                |> Maybe.withDefault Scope.Object
+    in
     { filename = Scope.toString scope ++ "-product-categories"
     , toId = .id >> ProductCategory.idToString
-    , toRoute = .id >> Just >> Dataset.ProductCategory scope >> Route.Explore scope
+    , toRoute = \{ id } -> Route.Explore scope (Dataset.ProductCategory genericScope (Just id))
     , toSearchableWords = ProductCategory.toSearchableString >> Text.toWords
     , facets =
         [ { key = "Transport réfrigéré"
--- a/src/Page/Explore/ProductCategories.elm
+++ b/src/Page/Explore/ProductCategories.elm
@@ -12,10 +12,10 @@ import Route
 
 
 table : Session -> Scope.GenericScope -> { detailed : Bool, scope : Scope } -> Table ProductCategory String msg
-table { db } genericScope { scope } =
-    { filename = Scope.toString scope ++ "-product-categories"
+table { db } genericScope _ =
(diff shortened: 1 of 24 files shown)


n1k0:
@camcoq-start @vjousse la review app est de nouveau testable :)

camcoq-start:
@n1k0 Alimentaire BETA is no longer in explorer 
image

src/Page/Explore/ProductCategories.elm
```diff
@@ -11,11 +11,11 @@ import Page.Explore.Table as Table exposing (Table)
 import Route
 
 
-table : Session -> { detailed : Bool, scope : Scope } -> Table ProductCategory String msg
-table { db } { scope } =
+table : Session -> Scope.GenericScope -> { detailed : Bool, scope : Scope } -> Table ProductCategory String msg
```
vjousse, line 14: (This is mostly a question to better understand the code base)
It looks strange here that we have to pass two scopes that I suppose should be coherent without a way to be sure that it’s the case. Am I missing something?

n1k0:
@camcoq-start c'était du au fait que les review-apps sont dérivées de l'app scalingo `ecpbalyse-staging` qui n'activait pas la variable d'environnement `ENABLE_FOOD2_SECTION` (en français: c'est réparé 😅)

vjousse (approved):
This looks good to me, tested locally and it works as expected (just a little question in my review for the sake of having a better understanding of the codebase)
I have an horizontal scrollbar in the explorer that may be removed with some width adjustments, what do you think?

Image

src/Page/Explore/ProductCategories.elm
n1k0: Good point! This is because `Page.Explore.Table.viewList` is built to handle both generic and legacy scopes, so we must pass a full scope to every call; I refactored the table to make this more explicit in 4c0b279ab2301798a0ef18c3aff4b7efa5b0f61e and 02d9a626c2ebade8584b1597bb76a50ae55531fb

camcoq-start (approved):
I tested these items. All look good :)
- [x] Log in
- [x] v.7
- [x] Bookmark and compare
- [x] Admin
- [x] Explorer
- [x] Object, veli and food2 simulations

PR #2728, opened 2026-08-17
chore: use `ty` typechecker

vjousse:
## :wrench: Problem

Python tooling now provides efficient type checkers like [ty](https://docs.astral.sh/ty/) that can improve the coding experience. We don’t use such tools on Ecobalyse.

## :cake: Solution

Run `ty` everywhere excluding the `data` dir for now as it has a lot of non production code in it.

I have added `ty` to the CI and to the pre-commit hooks. Some dead code has been removed as the same time.

Tests should be green, it should be a transparent move.


backend/src/app/cli/commands.py
```diff
@@ -399,8 +399,8 @@ async def load_processes_fixtures(
 @fixtures_management_group.command(
     name="load-processes", help="Load processes from JSON file."
 )
-@click.argument("json_files", type=click.File("rb"), nargs=-1)
-def load_processes_json(json_files: click.File) -> None:
+@click.argument("json_file", type=click.File("rb"), nargs=-1)
+def load_processes_json(json_file: click.File) -> None:
```
cedricr, line 403: Not sure about this change ; this method is supposed to accept several files I think  ? Otherwise, `nargs` shouldn’t be -1, and the iteration in line 414 doesn’t make sense.
Or, we can decide that it takes a single file.

```suggestion
@click.argument("json_files", type=click.File("rb"), nargs=-1)
def load_processes_json(json_files: list[click.File]) -> None:
```

cedricr (changes requested):
Finally ⭐⭐⭐

backend/src/app/cli/commands.py
vjousse: Good catch, as we only load one process file, I’ll change the nargs to one!

cedricr (approved):
🎉

PR #2731, opened 2026-08-18
chore: apply `ty` to `data/`

vjousse:
## :wrench: Problem

When using a system wide `ty` we need to tell it that the `data/` directory should be excluded.

## :cake: Solution

Add the corresponding configuration to the `pyproject.toml` file.

Using `ty` in your editor should not rise errors in `data/import_food.py` for example.


cedricr:
Not sure about that: I do want to have ty run in `data` ; obviously it raises a lot of errors at the moment, and as such it’s fine that it’s excluded in the CI, but the end goal is to fix `data` too, if we don’t display errors in the editor, this will keep getting worse.
Or, we could decide not to use types in `data` at all, as wrong types are worse than no types :sweat_smile:, but that would be a pity.

vjousse:
Agree! I’ll try to apply the suggestions for the data dir in this PR. If it’s too cumbersome, let’s merge it without it first and I’ll do the work in another PR.

vjousse:
Ok I did a full export and import after applying `ty` to the data dir and everything is working file. Do you see something else to test @ccomb @paulboosz @cedricr ?

paulboosz (approved):
LGTM

PR #2739, opened 2026-08-25
fix: downgrade highcharts to v13.0.0

n1k0:
Downgrades and pin Highcharts to v13.0.0 after v13.0.1 breaks the webapp, see https://github.com/highcharts/highcharts/issues/25072 (waiting for a proper upstream fix)

Also makes the prettier pre-commit hook command compatible with npm v12.0.2

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
         exclude: data
         types_or: [javascript, json]
         entry: |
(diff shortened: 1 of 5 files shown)


n1k0:
Updating targeted modules to load as [suggested in comment](https://github.com/highcharts/highcharts/issues/25072#issuecomment-5394804173) does not work, so I'm gonna revert Highcharts to v13.0.0 and wait for a proper patch version upstream

camcoq-start (approved):
LGTM :)

PR #2741, opened 2026-08-25
chore: fix e2e tests being cancelled with no error

vjousse:
## :wrench: Problem

When tests are in timeout or cancelled by the next run (every hour) they end up in a neutral state (no error, no success) and thus we are not aware that something is wrong.

## :cake: Solution

Fine tune the timeouts and retries values so that tests are running under 20 minutes even when they are timeouting (80 seconds * 11 tests, no retry). It prevents the tests to hang up for an hour and to be cancelled by Github Actions.

I’ve tested the values by reintroducing the regression with Highcharts and the tests are indeed failing this time (instead of being cancelled due to execution timeout before this patch) https://github.com/MTES-MCT/ecobalyse/actions/runs/32855695352

Current values allows the e2e tests to be successful https://github.com/MTES-MCT/ecobalyse/actions/runs/32867840464

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build:
-    timeout-minutes: 60
+    timeout-minutes: 55
     runs-on: ubuntu-latest
 
     services:
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -13,8 +13,7 @@ dotenv.config({ path: path.resolve(__dirname, ".env") });
  * @see https://playwright.dev/docs/test-configuration
  */
 export default defineConfig({
-  timeout: process.env.CI ? 120_000 : 10_000,
-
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -23,7 +23,7 @@ env:
 
 jobs:
   build:
-    timeout-minutes: 60
+    timeout-minutes: 55
```
n1k0, line 26: 💡

playwright.config.js
```diff
@@ -23,7 +22,7 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 10 : 0,
+  retries: 0,
```
n1k0, line 25: 💯

n1k0 (approved):
Great update!

PR #2745, opened 2026-08-25
refactor(data): share the SimaPro strategies and linking

ccomb:
## :wrench: Problem

brightway strategies may diverge between different databases, though they are all exported from simapro.

## :cake: Solution

refactor and share them in a common file

delete your $BRIGHTWAY2_DIR then run again a full import/export. `just import-all && just export-all` →no diff


vjousse (approved):
Tested locally, everything works as expected!

PR #2746, opened 2026-08-26
feat(data): introduce taxonomy.json, replace `base_ingredients.json`

paulboosz:
## :wrench: Problem

Begin #2733 

## :cake: Solution

- introduce `taxonomy.json` that replace `base_ingredients.json`. Each base_ingredient (eg `apple`) is classified under a material_type (eg `fruits_and_vegetables`). 
- validate taxonomy.json coherence with tests : each `base_ingredient` is under 1 and only 1 material_type (or class)
 - generate a `taxonomy_with_aliases.json` during the export to audit the quality of the taxonomy from a method perspective

- Some base_ingredient were under several material_type, in this case we put them under the majority material_type (eg `bean: mixed classes {'other_food_items': 1, 'cereals': 3}, keeping cereals`)
- in the next PRs ingredients material_type will be inferred from taxonomy.json and manual material_type in lci_catalog/*.json will be removed

- `just export-all` produce no diff
- generate the audit taxonomy file : `just generate-taxonomy-with-aliases`. There should be no diff

### Test the data migration 
`cd data`

bring back the input `base_ingredients.json`

`git show 17c6d195f~1:data/food/base_ingredients.json > food/base_ingredients.json`

run the migration

`uv run python migrations/2026-08-25-generate-taxonomy.py`

there should be no diff on taxonomy.json


data/bin/export.py
```diff
@@ -54,6 +55,10 @@ def metadata(
 
     activities = _get_lcias(root_dir)
 
+    if root_dir == DATA_ROOT_DIR:
+        # write the taxonomy_with_aliases() if it's not a test export
+        write_taxonomy_with_aliases()
+
```
vjousse: Depending on the directory to deduce what we should do looks strange to me. What about adding a flag to the command line instead that would allow the caller of the script to decide if the taxonomy should be written?

data/bin/generate_taxonomy_with_aliases.py
```diff
@@ -0,0 +1,51 @@
...
+            alias = metadata["alias"]
+            aliases_by_base[infer_base_ingredient(alias)].add(alias)
+
+    taxonomy_with_aliases = defaultdict(dict)
+    for base, material_type in material_type_by_base.items():
+        taxonomy_with_aliases[material_type][base] = sorted(aliases_by_base[base])
+    return {"food": dict(taxonomy_with_aliases)}
+
+
+def write_taxonomy_with_aliases():
```
vjousse: I would prefer if the `TAXONOMY_WITH_ALIASES_PATH` was passed as an argument to the method, defaulting to this value if you want, but the caller of this function should be able to chose where he wants to write the file.

data/bin/generate_taxonomy_with_aliases.py
```diff
@@ -0,0 +1,51 @@
...
+from config import DATA_ROOT_DIR
+
+TAXONOMY_WITH_ALIASES_PATH = DATA_ROOT_DIR / "food" / "taxonomy_with_aliases.json"
+
+
+def build_taxonomy_with_aliases() -> dict:
+    material_type_by_base = load_taxonomy()
+
+    aliases_by_base = defaultdict(set)
+    for lci_path in (DATA_ROOT_DIR / "lci_catalog").glob("*/*.json"):
```
vjousse: I would prefer if the `DATA_ROOT_DIR / "lci_catalog"` was passed as an argument to the method, defaulting to this value if you want, but the caller of this function should be able to chose where the json files are located.

data/common/infer_metadata.py
```diff
@@ -102,13 +102,44 @@ def infer_raw_to_cooked_ratio(
...
+}
+
+
+def parse_taxonomy(taxonomy: dict) -> dict[str, str]:
+    """Turn a taxonomy document into the baseIngredient -> material_type
+    mapping, rejecting unknown material_types and duplicated baseIngredients.
+    """
+    material_type_by_base = {}
+    for material_type, base_ingredients in taxonomy["food"].items():
+        if material_type not in KNOWN_MATERIAL_TYPES:
```
vjousse: Same here, you should avoid as much as you can to have functions calling global parameters when you can and instead put the required values as parameters to the function

vjousse (changes requested):

data/bin/export.py
paulboosz: done

data/bin/generate_taxonomy_with_aliases.py
paulboosz: done

data/bin/generate_taxonomy_with_aliases.py
paulboosz: done

data/common/infer_metadata.py
paulboosz: As KNOWN_MATERIAL_TYPES is a immutable frozenset I tought it was ok but done ✅

data/bin/export.py
```diff
@@ -54,6 +56,9 @@ def metadata(
 
     activities = _get_lcias(root_dir)
 
+    if write_taxonomy:
+        write_taxonomy_with_aliases()
```
vjousse: Could you explicitly pass the TAXONOMY_WITH_ALIASES_PATH here (so that we know where the taxonomy is written to when reading the export script), and maybe this path could be a setting config ?

vjousse (approved):

PR #2754, opened 2026-08-26
refactor: json formatting

### LINKED ISSUE #2639: Unifier le formatage json
## :wrench: Problème

Les fichiers json du front-end sont formatés par Prettier.
Les fichiers data sont formatés par le serialiser json de Python.

## :cake: Solution

Choisir un des deux mécanisme, et tout formater avec le même.
Se baser sur les listes d’exclusion (`/.prettierignore`, `/data/bin/json_formatter.py`) pour comprendre qui fait quoi.

## :rotating_light: Points d'attention/commentaires

Attention aux tris réalisés aujourd’hui ! (par ex. SORT_PATHS dans `/data/bin/json_formatter.py`)

vjousse:
## :wrench: Problem

fixes #2639 

## :cake: Solution

Use the python json formater for all the json files as we need the float precision number formatting for the data. I’ve enhanced the existing formater so that it allows small lists to be on the same line (like prettier was doing).

I’ve added a new file `.jsonformatignore` that follows the `.prettierignore` ignore format and removed the json files formatting from prettier.

```
npm run lint:json:all
npm run fix:json:all
```


.github/workflows/node.js.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     - name: Install Ubuntu dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y gettext
+        sudo apt-get install -y gettext ripgrep
```
n1k0: Not related to this change, but why do we need gettext at all already?

ccomb (approved):
Works after `uv sync --group data` and a right .env
However I'm wondering why we need to recurse in subdirs. Isn't there a way to lint and format only json files of the repository?

public/data/transports.json
cedricr, line 1: This file was excluded from formatting because it’s much more readable in its previous, more tabular, form (maybe that should be a csv rather than a json!)

package.json
```diff
@@ -22,21 +22,25 @@
...
     "lint:prettier:all": "npm run lint:prettier -- .",
     "lint:ruff:all": "npm run lint:ruff:check && npm run lint:ruff:format",
     "lint:ruff:check": "uv run ruff check --force-exclude --extend-select I",
     "lint:ruff:format": "npm run fix:ruff:format -- --check",
     "lint:ty:check": "uv run ty check",
-    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check",
+    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check && npm run lint:json:all",
     "fix:elm-format": "npx elm-format --yes",
+    "fix:json": "npm run lint:json -- --fix",
+    "fix:json:all": "npm run lint:json:all -- --fix",
```
n1k0: I'm still getting a weird error:

```
$ npm run fix:json:all
npm notice run ecobalyse@2.3.0 fix:json:all
npm notice run npm run lint:json:all -- --fix
npm notice run ecobalyse@2.3.0 lint:json:all
npm notice run rg --files --ignore-file .jsonformatignore -t json --hidden -0 | xargs -0 npm run lint:json --fix
npm warn "scalingo.json" is being parsed as a normal command line argument.
npm error code EUNKNOWNCONFIG
npm error Unknown cli flag:
npm error   - --fix
npm error Run `npm help config` for supported options.
npm error A complete log of this run can be found in: /Users/niko/.npm/_logs/2026-08-31T09_12_21_515Z-debug-0.log
```

This has started appearing since I've upgraded npm to v12.0.2…

```
➜  ecobalyse git:(chore/2639-unify-json-formatting) node --version
v22.22.2
➜  ecobalyse git:(chore/2639-unify-json-formatting) npm --version
12.0.2
```

If I downgrade npm to v11.13.0, the command works just fine. I don't know what we should do…

backend/src/app/db/fixtures/user.json
```diff
@@ -6,7 +6,9 @@
     "is_superuser": true,
     "last_name": "Dylan",
     "optin_email": false,
-    "organization": { "type": "individual" },
+    "organization": {
+      "type": "individual"
+    },
```
n1k0: Nit: now way to keep this on a single line? that would be consistent with how arrays are handled :)

data/ecobalyse_data/export/export_generic.py
```diff
@@ -248,12 +248,22 @@ def activities_to_processes_generic_json(
...
 
     without_detailed_impacts = remove_detailed_impacts(generic_dicts)
     for path in ecs_output_paths:
-        export_json(without_detailed_impacts, path)
+        export_json(
+            without_detailed_impacts,
+            path,
+            sort_fn=activities_processes_sort_key,
+            number_precision=settings.number_precision,
+        )
```
n1k0, line 250: Nit: it may be worth having a local `export_json` function

packages/python/ecobalyse/ecobalyse/json.py
```diff
@@ -0,0 +1,168 @@
...
+    def _format_value(self, o):
+        """Recursively apply number precision formatting and simple type coercions"""
+        # in python, bools are a subclass of int, so we should check explicitly
+        # if obj is not a bool, otherwise it will be converted to a float...
+        if isinstance(o, (int, float)) and not isinstance(o, bool):
+            if self.NUMBER_PRECISION is None:
+                return o
+            if o == 0:
+                return 0
+            return float(f"{o:.{self.NUMBER_PRECISION}g}")
```
cedricr, line 62: If we want to have some fun one day, it would be nice to use something like https://github.com/abolz/Drachennest 😆
That *might* even solve the mac/pc precision pb we have, and allow us to increase the precision again.
cc @paulboosz
Note: R uses grisu3 by default. The Python algorithm is apparently more naive (https://docs.python.org/3/library/string.html#format-specification-mini-language)

package.json
```diff
@@ -22,21 +22,25 @@
     "db:validate": "uv run --group data check-jsonschema --schemafile schemas/processes-schema.json public/data/processes*.json",
     "lint:openapi": "npx swagger-cli validate openapi.yaml",
     "lint:elm": "npx elm-format src review/src --validate",
+    "lint:json": "uv run bin/json_formatter.py",
```
paulboosz, line 25:
```suggestion
    "lint:json": "uv run --group data bin/json_formatter.py",
```
else I get an error as dynaconf is not in default group

data/ecobalyse_data/export/textile.py
```diff
@@ -15,7 +17,9 @@ def activities_to_materials_json(
 
     materials_dicts.sort(key=lambda x: x["id"])
 
-    export_json(materials_dicts, materials_path)
+    export_json(
+        materials_dicts, materials_path, number_precision=settings.number_precision
```
n1k0, line 21: It feels like `number_precision` is always `settings.number_precision`, shouldn't this be enforced in `export_json` directly?

data/bin/export.py
```diff
@@ -195,8 +197,18 @@ def merge_processes(
...
+        public_dir / settings.processes_merged_impacts_file,
+        sort_fn=activities_processes_sort_key,
+        number_precision=settings.number_precision,
+    )
+    export_json(
+        merged_ecs,
+        public_dir / settings.processes_merged_ecs_file,
+        sort_fn=activities_processes_sort_key,
+        number_precision=settings.number_precision,
+    )
```
n1k0, line 206: Nit: could use a single local shared fn

data/common/export.py
```diff
@@ -238,19 +224,29 @@ def export_processes_to_dir(
...
+    )
     exported_files.append(processes_impacts_absolute_path)
 
     # Also update the aggregated file
-    export_json(remove_detailed_impacts(filtered), processes_ecs_absolute_path)
+    export_json(
+        remove_detailed_impacts(filtered),
+        processes_ecs_absolute_path,
+        number_precision=settings.number_precision,
+    )
```
n1k0, line 260: Nit: could use a local fn (and the more I read the whol patch, maybe a broadly shared one)

pyproject.toml
```diff
@@ -36,7 +36,7 @@ requires = ["hatchling", "setuptools"]
 allow-direct-references = true
 
 [tool.hatch.build]
-dev-mode-dirs = ["./packages/python/ecobalyse", "./backend/src", "."]
+dev-mode-dirs = ["./packages/python/ecobalyse", "./backend/src", ".", "./data"]
```
n1k0, line 41: Nit: What does this do? May be worth a comment

n1k0 (commented):
Great work! This is a first batch of comments, I'll keep testing the patch :)

bin/filter_ignored.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Prend une liste de fichiers en arguments, et affiche ceux qui ne sont PAS
+# ignorés par .jsonignore un par ligne.
+# On optimise la recherche avec rg lui évitant de chercher à la racine du rep à chaque fois
+# en lui spécifiant les répertoires des fichiers passés en paramètre
+set -euo pipefail
+
+dirs=$(printf '%s\n' "$@" | xargs -n1 dirname | sort -u)
+allowed=$(rg --files --ignore-file .jsonformatignore --hidden -t json -- $dirs | sed 's#^\./##')
```
cedricr, line 9: I think it’s the job of the json_formatter command itself to manage its ignore file list.
Why not replace this script by https://docs.python.org/fr/3/library/os.html#os.walk ?

backend/src/app/db/fixtures/user.json
vjousse: Well the problem is that… it’s a classic json object, so putting it on one line would put every json object on one line. What could be the rule here, objects with only one key should be formatted as one liners?

data/ecobalyse_data/export/export_generic.py
vjousse: What do you mean by « local » ?

packages/python/ecobalyse/ecobalyse/json.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class CompactJSONEncoder(json.JSONEncoder):
+    """A JSON Encoder that puts small containers on single lines and formats
+    numbers with a reduced precision."""
+
+    CONTAINER_TYPES = (list, tuple, dict)
+    """Container datatypes include primitives or other containers."""
+
+    MAX_WIDTH = 70
```
cedricr: 80 is probably fine, as it looks it’s a hard limit, no ?

packages/python/ecobalyse/ecobalyse/json.py
```diff
@@ -0,0 +1,168 @@
...
+            return " " * (self.indentation_level * self.indent)
+        elif isinstance(self.indent, str):
+            return self.indentation_level * self.indent
+        else:
+            raise TypeError(
+                f"indent must either be of type int or str (is: {type(self.indent)})"
+            )
+
+
+def activities_processes_sort_key(entry: dict[str, Any]) -> tuple:
```
cedricr: I’d probably move this outside of this file, as it’s specific to Ecobalyse, whereas the rest is quite generic

data/ecobalyse_data/export/export_generic.py
n1k0: local to the module

backend/src/app/db/fixtures/user.json
n1k0: I think prettier handles this with the enforced max line length setting, I don't know if that's feasible with the python formatter… single key objects as one liner sounds good anyway!

.github/workflows/node.js.yml
vjousse: I think that `gettext` is needed for some pythons CLI scripts, but I may have to double check that

.jsonformatignore
```diff
@@ -0,0 +1,13 @@
...
+dist/
+elm.json
+package-lock.json
+package.json
+public/icomoon/*
+public/vendor/*
+tests/e2e-*-output.json
+tests/e2e-food.json
+tests/e2e-textile.json
+tests/processes-schema.json
```
cedricr: the file has moved to /schemas/processes-schema.json, but I see no reason not to format it.

.github/workflows/node.js.yml
n1k0: I remember it was a prerequisite for django but since we ditched it a long time ago… 😅

README.md
```diff
@@ -8,7 +8,7 @@ L’application est accessible [à cette adresse](https://ecobalyse.beta.gouv.fr
 
 ## Socle technique et prérequis
 
-Le frontend de cette application est écrit en [Elm](https://elm-lang.org/). Vous devez disposer d’un environnement [NodeJS](https://nodejs.org/fr/) 22+ et `npm`. Pour le backend vous devez disposer d’un environnement [python](https://www.python.org/) >=3.12, [uv](https://docs.astral.sh/uv/) et [gettext](https://www.gnu.org/software/gettext/) sur votre machine. Certains fichiers d’impacts détaillés nécessitent de configurer [`transcrypt`](https://github.com/elasticdog/transcrypt) pour les lire en local.
+Le frontend de cette application est écrit en [Elm](https://elm-lang.org/). Vous devez disposer d’un environnement [NodeJS](https://nodejs.org/fr/) 22+ et `npm`. Pour le backend vous devez disposer d’un environnement [python](https://www.python.org/) >=3.12, [uv](https://docs.astral.sh/uv/) et [gettext](https://www.gnu.org/software/gettext/) sur votre machine. Certains fichiers d’impacts détaillés nécessitent de configurer [`transcrypt`](https://github.com/elasticdog/transcrypt) pour les lire en local et le formattage des fichiers `json` requiert l’installation de [`ripgrep`](https://github.com/burntsushi/ripgrep).
```
n1k0:
```suggestion
Le frontend de cette application est écrit en [Elm](https://elm-lang.org/). Vous devez disposer d’un environnement [NodeJS](https://nodejs.org/fr/) 22+ et `npm`. Pour le backend vous devez disposer d’un environnement [python](https://www.python.org/) >=3.12, [uv](https://docs.astral.sh/uv/) et [gettext](https://www.gnu.org/software/gettext/) sur votre machine. Certains fichiers d’impacts détaillés nécessitent de configurer [`transcrypt`](https://github.com/elasticdog/transcrypt) pour les lire en local et le formatage des fichiers `json` requiert l’installation de [`ripgrep`](https://github.com/burntsushi/ripgrep).
```

.jsonformatignore
```diff
@@ -0,0 +1,13 @@
...
+
+!/data/common/distances/distances_raw.json
```
cedricr: Something was lost in translation ;->
It was listed in the prettierignore file, with a leading `!` to signify that we actually wanted this file formatted by prettier (prettier formating was more compact than the old json formatter).
Anyway :
- the `!` is not supported by your filter_ignored.sh file, so this ligne doesn’t do anything at the moment
- I’d like this file to stay formatted in a “tabular mode” if it’s possible, but see the comment on the file itself.

data/ecobalyse_data/export/food.py
```diff
@@ -165,7 +165,6 @@ def activities_to_ingredients_json(
     ecosystemic_factors_path: str,
     feed_file_path: str,
     raw_to_transformed_file_path: str,
-    cpu_count: int,
```
n1k0, line 168: just out of curiosity: what was that? 😅

.jsonformatignore
```diff
@@ -0,0 +1,13 @@
+.proxyrc.json
```
cedricr: 🤔 It’s not a generated file, is it ? So it could stay formatted (and I see it’s listed in prettierignore too, I’ll put a comment there too)

.prettierignore
cedricr: already covered by ligne 14.

.jsonformatignore
```diff
@@ -0,0 +1,13 @@
...
+dist/
+elm.json
+package-lock.json
+package.json
```
cedricr: It’s excluded from prettier now, so probably better to keep it formatted

.pre-commit-config.yaml
```diff
@@ -57,22 +54,16 @@ repos:
...
-        name: run ruff check
-        language: system
-        files: ^data/.*$
-        types_or: [python]
-        entry: |
-          bash -c 'just fix-python-data "${@}"' --
+          bash -c '
+            to_format=$(bin/filter_ignored.sh "$@")
+            [ -n "$to_format" ] && uv run bin/json_formatter.py --fix $to_format
+            exit 0
```
n1k0: Why do we silence error code here? I think it would be important to be notified when the formatter fails…

packages/python/ecobalyse/ecobalyse/json.py
```diff
@@ -0,0 +1,168 @@
...
+    NUMBER_PRECISION = None
+    """Number of significant digits to keep for floats (and ints). None disables formatting."""
+
+    def __init__(self, *args, number_precision=None, **kwargs):
+        # using this class without indentation is pointless
+        if kwargs.get("indent") is None:
+            kwargs["indent"] = 4
+        super().__init__(*args, **kwargs)
+        self.indentation_level = 0
+        if number_precision is not None:
```
cedricr: The test is probably useless, as None is anyway the default value

bin/json_formatter.py
```diff
@@ -32,38 +33,41 @@
...
+
+        assert src_data is not None
+        try:
+            input_data = json.loads(src_data)
+
+            number_precision = None
+            if path.name in SORT_PATHS:
+                input_data.sort(key=activities_processes_sort_key)
+                # Only apply precision number to processes files
+                number_precision = settings.number_precision
```
cedricr: note for later : dynaconf config is in the data directory, but it should probably become global

bin/json_formatter.py
n1k0, line 23: I'd naturally expect this list to exactly match the one in `.jsonformatignore`… what's the actual source of truth? why the diff with `/data/common/distances/distances_raw.json` and `!/data/common/distances/distances_raw.json`? maybe worth a comment if not a mistake

also `tests/activities-schema.json` and `tests/processes-schema.json` don't exist anymore so this should be updated

data/ecobalyse_data/export/export_generic.py
```diff
@@ -248,12 +248,22 @@ def activities_to_processes_generic_json(
     )
 
     for path in impacts_output_paths:
-        export_json(generic_dicts, path)
+        export_json(
+            generic_dicts,
+            path,
+            sort_fn=activities_processes_sort_key,
```
cedricr: there are lots of repetitions of this call ; I’d probably create a `export_ecobalyse_json` wrapper somewhere, and only call that.
Something like
```
def export_ecobalyse_json(self, json_data: dict[str, Any] | list[dict[str, Any]],
    filename)
    return export_json(json_data, filename, sort_fn=activities_processes_sort_key,
            number_precision=settings.number_precision)

File: bin/json_formatter.py [hunk 24]
```diff
@@ -6,9 +6,9 @@
 from typing import Annotated
 
 import typer
-
-from common import activities_processes_sort_key
-from ecobalyse_data.logging import logger
+from data.config import settings
```
n1k0: so importing this makes the json formatter stuff relying on dynaconf/env just to know about NUMBER_PRECISION… I don't see added value to make this configurable through env/conf, and it enforce the need for `S3*` being set in a non-CI env like a dev box… we should probably hardcode the NUMBER_PRECISION value to 4 or make it a cli flag…

.jsonformatignore
```diff
@@ -0,0 +1,13 @@
...
+public/icomoon/*
+public/vendor/*
+tests/e2e-*-output.json
```
cedricr, line 6: Do we really need these exclusions ?

.pre-commit-config.yaml
```diff
@@ -19,13 +19,12 @@ repos:
         exclude: data
         types_or: [javascript, json]
```
n1k0: we must remove `json` from this array as prettier will still load and attempt at formatting json files as it formats explicitly passed files even when they match the ignore file

.pre-commit-config.yaml
```diff
@@ -57,22 +54,16 @@ repos:
...
-          bash -c 'just fix-json-data "${@}"' --
-
-      - id: fix-python
-        name: run ruff check
-        language: system
-        files: ^data/.*$
-        types_or: [python]
-        entry: |
-          bash -c 'just fix-python-data "${@}"' --
+          bash -c '
```
cedricr, line 62: I’d probably add a `just` command that we can use everywhere, and avoid as much as possible any complexity here.

bin/json_formatter.py
n1k0: hmm we should definitely enforce a number_precision in export_json instead of falling back to what's expected to always be the same value across the project anyway

n1k0 (commented):
New batch of remarks and suggestions… I'll keep reviewing thoroughly as this is a super impactful change

cedricr (approved):
Great job ! 
I’ve listed a few possible ameliorations, especially if you want to make a dedicated modile.

packages/python/ecobalyse/ecobalyse/json.py
cedricr: (and maybe that would be enough so that the transport file is formatted more compactly)

package.json
cedricr: yep, I think we need to pull up dynaconf and its config, but that can be donne in another PR

package.json
```diff
@@ -22,21 +22,25 @@
...
     "lint:ruff:check": "uv run ruff check --force-exclude --extend-select I",
     "lint:ruff:format": "npm run fix:ruff:format -- --check",
     "lint:ty:check": "uv run ty check",
-    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check",
+    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check && npm run lint:json:all",
     "fix:elm-format": "npx elm-format --yes",
+    "fix:json": "npm run lint:json -- --fix",
+    "fix:json:all": "npm run lint:json:all -- --fix",
     "fix:prettier": "npm run lint:prettier -- --write",
     "fix:prettier:all": "npm run lint:prettier -- --write .",
```
n1k0, line 38: are these still necessary?

package.json
```diff
@@ -22,21 +22,25 @@
...
     "fix:ruff:all": "npm run fix:ruff:check && npm run fix:ruff:format",
     "fix:ruff:check": "npm run lint:ruff:check -- --fix",
     "fix:ruff:format": "uv run ruff format --force-exclude",
     "fix:ty:check": "npm run lint:ty:check -- --fix",
-    "fix:all": "npm run fix:elm-format . && npm run fix:ruff:all && npm run fix:prettier:all",
+    "fix:all": "npm run fix:elm-format . && npm run fix:ruff:all && npm run fix:prettier:all && npm run fix:json:all",
     "format:json": "npx prettier@3.0.3 --write . && uv run ruff check --select I --fix && uv run ruff format",
```
n1k0: pretty sure this should be removed (or rewired)

data/tests/test_json_format_encoder.py
```diff
@@ -1,43 +1,130 @@
...
+        test_7,
+        test_8,
+        test_9,
+        test_10,
+        test_0_are_kept_as_int,
     ],
 )
 def test_format_number_json_encoder(input_data, expected, test_id):
-    encoder = FormatNumberJsonEncoder()
+    encoder = CompactJSONEncoder(number_precision=4)
```
n1k0, line 96: How about passing `indent=2` so we test what's called in the formatter?

Also, a few more tests covering MAX_WIDTH / MAX_ITEMS and compact objects might be useful to add

n1k0 (approved):
Last batch of comments, I think most of what needs to be addressed before landing has been mentioned by other folks as well. I let that up to you though :)

data/ecobalyse_data/export/food.py
vjousse: It was using for multiprocessing export but it’s not relevant anymore!

backend/src/app/db/fixtures/user.json
vjousse: I’ve applied the `MAX_WIDTH` rule for objects too

.github/workflows/node.js.yml
vjousse: Nice catch it’s not used anymore!

bin/filter_ignored.sh
vjousse: This was for performance reason and because by default, `rg` reads the same format than prettier for its ignore file. I’ll see how I could improve that and maybe get rid on the `rg` dep.

packages/python/ecobalyse/ecobalyse/json.py
vjousse: Done!

package.json
vjousse: What do you mean, removing the whole `fix:all` command in npm?

n1k0: No I'm talking about `format:json` which calls prettier at a pinned version directly, line 44

package.json
vjousse: could you give it another try?

bin/json_formatter.py
```diff
@@ -6,9 +6,8 @@
 from typing import Annotated
 
 import typer
-
-from common import activities_processes_sort_key
-from ecobalyse_data.logging import logger
+from ecobalyse.json import CompactJSONEncoder, activities_processes_sort_key
+from ecobalyse.logging import logger
```
n1k0, line 11:
```suggestion
import json
import logging
from pathlib import Path
from typing import Annotated

import typer
from ecobalyse.json import CompactJSONEncoder, activities_processes_sort_key

logger = logging.getLogger(__name__)
```

n1k0: I think it’s safer to use stdlib logging here instead of `ecobalyse.logging`. That logger still imports Dynaconf whenever the package is present, so `lint:json` keeps requiring a valid `.env` after `uv sync --group data`.

Note: I did my best to provide an actionable patch using the gihub ui but failed miserably :D

the idea is to replace 

```python
from ecobalyse.logging import logger
```

with

```python
import logging

logger = logging.getLogger(__name__)
```

package.json
```diff
@@ -22,21 +22,25 @@
     "db:validate": "uv run --group data check-jsonschema --schemafile schemas/processes-schema.json public/data/processes*.json",
     "lint:openapi": "npx swagger-cli validate openapi.yaml",
     "lint:elm": "npx elm-format src review/src --validate",
+    "lint:json": "uv run bin/json_formatter.py",
+    "lint:json:all": "rg --files --ignore-file .jsonformatignore -t json --hidden -0 | xargs -0 npm run lint:json",
```
n1k0, line 26: how about avoiding passing xargs filename thourgh `npm run` which behavior has been unconsistent since a recent `npm` major version?

```json
    "lint:json:all": "rg --files --ignore-file .jsonformatignore -t json --hidden -0 | xargs -0 uv run bin/json_formatter.py",
```

package.json
```diff
@@ -22,21 +22,25 @@
...
     "lint:prettier:all": "npm run lint:prettier -- .",
     "lint:ruff:all": "npm run lint:ruff:check && npm run lint:ruff:format",
     "lint:ruff:check": "uv run ruff check --force-exclude --extend-select I",
     "lint:ruff:format": "npm run fix:ruff:format -- --check",
     "lint:ty:check": "uv run ty check",
-    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check",
+    "lint:all": "npm run lint:prettier:all && npm run lint:elm && npm run lint:ruff:all && npm run lint:openapi && npm run lint:ty:check && npm run lint:json:all",
     "fix:elm-format": "npx elm-format --yes",
+    "fix:json": "npm run lint:json -- --fix",
+    "fix:json:all": "rg --files --ignore-file .jsonformatignore -t json --hidden -0 | xargs -0 npm run fix:json",
```
n1k0, line 36: Same

```json
    "fix:json:all": "rg --files --ignore-file .jsonformatignore -t json --hidden -0 | xargs -0 uv run bin/json_formatter.py --fix",
```

.pre-commit-config.yaml
```diff
@@ -57,22 +54,16 @@ repos:
...
+            to_format=$(bin/filter_ignored.sh "$@")
+            [ -n "$to_format" ] && uv run bin/json_formatter.py --fix $to_format
+            exit 0
+          ' --
```
n1k0, line 65:
```suggestion
          bash -c '
            to_format=$(bin/filter_ignored.sh "$@")
            [ -n "$to_format" ] && uv run bin/json_formatter.py --fix $to_format
          ' --
```

I still think it's pretty important to avoid silencing formatter errors here, or maybe I'm missing a good reason to do so?

bin/json_formatter.py
cedricr: The problem is that we’ve already had to change this value twice in 6 months, because of floating point differences between Mac and PC. I’d like to avoid having it hardcoded is some places but not others.
We should probably make the S3 variables optional anyway (you cannot run the data pipeline without them, which means you can’t fix the occasional impact file conflict, but yes, they were made mandatory when the settings file was only scoped to the data repository. This makes less sense now). There’s no point having a configuration file, and not allow ourselves to use it when needed :sweat_smile:

package.json
paulboosz: yay it works :)

package.json
vjousse: I thought you were still using it for CSS or JS @n1k0 ?

.jsonformatignore
vjousse: Agree about `package.json` but as the others are generated, we should keep them in the ignore file

bin/json_formatter.py
vjousse: I need to think more about this one as the default logger really sucks. I’ve added code in the logging module to check if dynaconf is present (it’s just used to set the log level) and if note, we don’t require it (and don’t set the default level from the config file).
It should do the trick for now, but we will to decide what we do about dynaconf more globally

package.json
vjousse: I’ll try to manage the search and the excluded files directly in python, it should avoid having to call rg and xargs.

package.json
vjousse: Same as above https://github.com/MTES-MCT/ecobalyse/pull/2754#discussion_r3916005881

package.json
n1k0: to be honnest I'm using the prettier plugin in my editor to format css code, so I think this could be ditched anyway

vjousse: Ok but locally I don’t, so I’m afraid that if I have to edit JS or CSS files, we may end up having diffs on formatting. What do you think?

.jsonformatignore
vjousse: For package.json no, I’ve fixed it. For the other files, as they are generated by tools, should we really bother formatting them?

package.json
n1k0: Yeah you're right! Consistent formatting is probably more important than ditching dependencies here. We do want consistent js and css formatting, so let's keep it.

bin/json_formatter.py
```diff
@@ -0,0 +1,183 @@
...
+
+import json
+import os
+import posixpath
+from pathlib import Path
+from typing import Annotated
+
+import pathspec
+import typer
+from data.common.export import activities_processes_sort_key
```
n1k0: apart if I'm missing something in the way imports work, importing from data.common.export will import ecobalyse.logger (which uses dynaconf) and a bunch of other stuff unrelated to formating json (eg. impacts.json loading)

data/common/export.py
```diff
@@ -2,19 +2,41 @@
...
+
+
+def activities_processes_sort_key(entry: dict[str, Any]) -> tuple:
+    return (
+        entry.get("source", ""),
+        entry.get("activityName", ""),
+        entry.get("location"),
+        entry.get("alias") or "",
+        entry.get("displayName", ""),
+    )
```
n1k0: how about moving this outside of this specific module to solve the isse?

.pre-commit-config.yaml
```diff
@@ -57,22 +54,12 @@ repos:
...
-          bash -c 'just fix-json-data "${@}"' --
-
-      - id: fix-python
-        name: run ruff check
-        language: system
-        files: ^data/.*$
-        types_or: [python]
-        entry: |
-          bash -c 'just fix-python-data "${@}"' --
+          bash -c 'npm --prefix . run fix:json:all "${@}"' --
```
n1k0: wouldn't that reformat the whole repo? is this what we want? wouldn't it be slow wrt so many json files in the tree? I'd expect to format only staged paths in a pre-commit hook

vjousse: nice catch!

data/common/export.py
vjousse: Done!

n1k0 (approved):
Great work, let's land this behemoth 🎉

PR #2755, opened 2026-08-27
fix(generic): apply item qty to assembly transports

### LINKED ISSUE #2333: [BUG] Modifier la quantité du composant n'augmente pas le coût environnemental du transport vers la phase d'assemblage
## Explication du bug / Bug Explanation

Lorsque je positionne la quantité d'un composant à 2, le coût environnemental du transport vers la phase d'assemblage ne double pas:

Image

Image

Le problème semble isolé à la vue puisque le coût du transport est bien multiplié entre assemblage/emballage et distribution. Il faudrait regarder si c'est un bug d'affichage ou si le problème est plus profond au niveau du moteur de calcul.

## Comportement attendu / Expected Behavior

Lorsque je mets 2 dans un composant, le coût environnemental du transport vers la phase d'assemblage devrait doubler.

n1k0:
## :wrench: Problem

Fixes #2333

Item quantity wasn't applied when computing transports to assembly stage.

## :cake: Solution

Ensure multiplying element impacts according to parent element quantity.

Set the quantity of a component to 2: the transport impacts to assembly should be doubled.

### DIFF
--- a/src/Data/Component.elm
+++ b/src/Data/Component.elm
@@ -1229,8 +1229,10 @@ computeTransports ({ config, db } as requirements) ({ assembly, transportOptions
                     )
                     -- toAssembly
                     (transportElements <|
-                        \( expandedElement, elementResults ) ->
+                        \( quantity, expandedElement, elementResults ) ->
+                            -- element masses are per unit; scale by the parent item quantity
                             extractMass elementResults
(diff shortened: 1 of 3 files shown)


vjousse (approved):
Nice catch!

PR #2756, opened 2026-08-27
chore: import rapidoc from npm

n1k0:
We should load rapidoc from npm rather than vendoring it locally.

### How to test

Toy around with the OpenAPI explorer


vjousse (approved):
Tested locally, it works as expected, it’s way better like that 😍

PR #2761, opened 2026-08-27
fix(data): keep the method in sync with the biosphere it characterizes

ccomb:
## :wrench: Problem

`import_method.py` skipped the import when a method of that name already existed. Every database import adds elementary flows to the biosphere, and a method written before them characterizes none of them, so those emissions count zero in every score.

`add_unlinked_flows_to_biosphere_database` added new flows by rewriting the whole biosphere. `Database.write` reassigns an id to every node, so the datapackages built before the rewrite point at the wrong rows.

## :cake: Solution

The method is reimported on every run, and new biosphere flows are appended one at a time so existing flows keep their ids. Both paths call `process()` where it is needed, which leaves `sync_datapackages` with nothing to do, so it is removed from the pipeline. \o/

Two reports watch what the import silently discarded until now:

- factors naming a substance the biosphere carries, on a compartment that reaches none of its flows, are written to `data/output/dropped-cfs-<method>.csv` (28 substances on the current data);
- a flow ending up with two disagreeing factors stops the import and names the renames that merged them, because `bw2calc` sums factors sitting on one flow and would multiply that substance in every score.

The collision check finds nothing on the current data. It guards the renames in `data/simapro-biosphere.json`, where a future entry merging two names that carry different factors would otherwise pass unnoticed.

The factors keep the SimaPro name they arrived with, so a collision can name the rename responsible. That extra key has to be dropped before the deduplication that follows, otherwise identical factors no longer compare equal and get summed.

- `cd data && just test` runs the two new test modules (`tests/test_colliding_cfs.py`, `tests/test_dropped_cfs.py`).
- delete your BRIGHTWAY2_DIR then `just import-all && just export-all` → no diff


vjousse (approved):
Tested locally: everything’s looking good!

PR #2771, opened 2026-09-02
fix(security): upgrade maildev to v3.x

n1k0:
Supersedes #2770, upgrading and adapting the codebase to use maildev v3rc which ships with a security fix for nodemailer.

Also upgrades `browserslist` to 4.28.8 as per #2770 initial intention :)

- `maildev` was improperly called with the `-p` flags which was already ignored in v2 and rejected in v3. package.json commands have been refactored accordingly, and e2e tests updated to use the new v3 api
- I've also fixed and improved (I think?) e2e tests — which are failing on `main` atm — reliability by now waiting that all data are actually loaded over http before performing assertions the app renders the header menu, as well as a edge case where users weren't properly redirected to their account page when reclicking on a magic link, which is what the *auth* e2e test suite is doing **a lot** 😅

- [ ] Tests should be green on this PR
- [ ] Clone the branch locally and run the e2e tests with `npm run test:e2e`: they should now all pass 🤞
- [ ] Side-fixed bug: try reclicking a magic link once already signed in; you should be redirected to your account page instead of the "confirmer la connexion" screen (which was a source of intermittent e2e auth test failure)


n1k0:
@vjousse ping :)

vjousse (approved):
I feel your pain about e2e tests 🫂 Works as expected locally, great job!

n1k0:
I'm landing this so we'll know for sure if the e2e test suite is fixed in the ci cronjob 🤞

PR #2783, opened 2026-09-03
feat(data): volca comparison script

ccomb:
## :wrench: Problem

We publish the impacts of many processes, computed with Brightway from SimaPro exports. Nothing in this repository checks those numbers against a second engine reading the same files. The only comparison we previously had was using a remote SimaPro with a quick and dirty fastapi using the COM interface. This was unreliable, extremely slow (several seconds per process) and depending on the installed SimaPro and its current databases.

## :cake: Solution

`just compare-volca`:  a standalone command that recomputes the detailed impacts of the processes we use, then draw one diagonal cloud for each impact category :

image

and terminates by showing the biggest differences:

```
The 30 widest deviations, all categories together:
  htc-c    +2656%  Vanilla, cured and dried beans, at farm (WFLDB) [MG] (WFLDB)
  htc-c    +2481%  Apple {CN}| apple production | Cut-off, U [CN] (Ecoinvent 3.11)
  ior        -96%  Soft cheeses - cat E, 220g | Packaging System, N0, All, Plastic box with label (mono) {FR} U [FR] (Agribalyse 3.2)
  htc-c    +2387%  Pear {RoW}| pear production | Cut-off, U [RoW] (Ecoinvent 3.11)
  htc-c    +2349%  Lentils, dry, at farm (WFLDB) [CA] (WFLDB)
  htc      +1778%  Apple {CN}| apple production | Cut-off, U [CN] (Ecoinvent 3.11)
  htc      +1755%  Vanilla, cured and dried beans, at farm (WFLDB) [MG] (WFLDB)
  wtu      +1590%  Mineral water, still, PET one-way, 1.5 L, at water factory {FR} - Adapted from WFLDB U [FR] (Agribalyse 3.2)
(...)
written output/compare_volca.json and output/compare_volca.svg
```

- Only one script and two new dependencies : [pyvolca](https://pypi.org/project/pyvolca/) and `matplotlib` (in the pyproject `volca` group).
- Same prerequisite as `just import-all`: the S3 credentials in `.env`.
- The engine writes a cache next to each source file so the second run is twice as fast. You can delete them.
- Ginko's current refusal is a finding, not a failure: the run reports the three missing inputs by name and carries on with the other databases.
- The upcoming PRs will fix some divergences and change the scores.

- `just compare-volca`
- Open `output/compare_volca.svg` with a browser or inkscape
- Read the summary and the 30 widest deviations printed underneath
- Open `output/compare_volca.json` to see the numbers

### DIFF
--- a/.justfile
+++ b/.justfile
@@ -36,10 +36,6 @@ import-method:
 create-activities:
     {{ python-cmd-data }} data/create_activities.py
 
-# Compare the impacts we publish with the ones VoLCA computes on the same source files
-compare-volca:
-    uv run --group data --group volca python data/bin/compare_volca.py
-
 ################################################################################
 ### Data exports
 
@@ -128,7 +124,11 @@ fix-all-data: fix-json-data fix-python-data
 ci-data: check-all-data
 
 ################################################################################
-### Testing
+### Testing and analysing
 
 test-data:
     uv run --group data pytest data
+
+# Compare the impacts we publish with the ones VoLCA computes on the same source files
+compare-volca:
+    uv run --group data --group volca python data/bin/compare_volca.py
--- a/.justfile
+++ b/.justfile
@@ -36,6 +36,10 @@ import-method:
 create-activities:
     {{ python-cmd-data }} data/create_activities.py
 
+# Compare the impacts we publish with the ones VoLCA computes on the same source files
+compare-volca:
+    uv run --group data --group volca python data/bin/compare_volca.py
+
 ################################################################################
 ### Data exports
 
--- a/data/bin/compare_volca.py
+++ b/data/bin/compare_volca.py
@@ -1,0 +1,583 @@
+"""Compare the impacts we publish with the ones VoLCA computes from the same source files.
+
+One cloud per impact category: a point is one activity, x the impact published in
+processes_impacts.json, y what VoLCA computes on the same SimaPro export with the same
+method. The command stands on its own: it installs the engine, generates its configuration,
+starts a server on a free port, loads the seven databases, resolves every published activity
+that has a source file behind it, scores each database in one batch, draws, and says what
+each phase cost.
+
+It needs an engine that carries its own reference data, so VoLCA v0.12.0 or later.
+VOLCA_BINARY points it at a local build instead of the release pyvolca installs.
+"""
+
+import csv
+import io
+import json
+import os
+import statistics
+import sys
+import tempfile
+import time
+from collections.abc import Callable, Iterator, Sequence
+from contextlib import contextmanager, redirect_stderr
+from math import copysign, log
+from pathlib import Path
+from typing import NamedTuple, TypedDict
+
+import matplotlib.pyplot as plt
+import volca
+from volca import Client, Server, VoLCAError
+
+from common import (
+    calculate_aggregate,
+    correct_process_impacts,
+    get_normalization_weighting_factors,
+)
+from common.export import IMPACTS_JSON
+from common.impacts import impacts
+from config import DATA_ROOT_DIR, settings
+from ecobalyse_data import s3
+from ecobalyse_data.bw.migration import (
+    GINKO_MIGRATIONS,
+    PASTOECO_MIGRATIONS,
+    WOOLMARK_MIGRATIONS,
+)
+from ecobalyse_data.bw.strategy import use_unit_processes
+
+
+def db_name(file: str) -> str:
+    """The engine's name for one source file, ours alone so it never shares a cache."""
+    return f"eb-{file.lower()}"
+
+
+class Source(NamedTuple):
+    file: str  # settings.dbfiles key
+    dep: str | None = None  # the file its inputs link to, as in Ecobalyse's import
+    migrations: Sequence[dict] = ()  # Ecobalyse's renames, handed to VoLCA as aliases
+    unit_processes: bool = False  # also rename what strategy.use_unit_processes renames
+
+    @property
+    def db(self) -> str:
+        return db_name(self.file)
+
+
+SOURCES = {  # "source" in processes_impacts.json -> where VoLCA reads it
+    "Agribalyse 3.2": Source("AGRIBALYSE"),
+    "Ecoinvent 3.9.1": Source("EI391"),
+    "Ecoinvent 3.11": Source("EI311"),
+    "WFLDB": Source("WFLDB"),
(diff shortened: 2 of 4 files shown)


.justfile
```diff
@@ -35,6 +35,10 @@ import-method:
 create-activities:
     {{ python-cmd-data }} data/create_activities.py
 
+# Compare the impacts we publish with the ones VoLCA computes on the same source files
```
cedricr: Maybe move that somewhere else in the file (that’s the “import section”). Maybe in the last section “Testing”, which could become “Testing and analysing" ?

pyproject.toml
```diff
@@ -104,3 +108,10 @@ extend-select = ["I"]
 extra-paths = [
     "data/",
 ]
+
+# pyvolca is only installed by the optional "volca" group, so ty never resolves it
+[[tool.ty.overrides]]
```
cedricr, line 118: not a big fan of that if we can avoid it, but I gave it a quick look and didn’t find a good way to do it. @vjousse do you have an idea ?

pyproject.toml
```diff
@@ -104,3 +108,10 @@ extend-select = ["I"]
...
+include = ["data/bin/compare_volca.py"]
+
+[tool.ty.overrides.rules]
```
cedricr: not a big fan of that either… if there’s no other way, maybe ignore that with comments directly in the Volca script ?

data/bin/compare_volca.py
```diff
@@ -0,0 +1,583 @@
+"""Compare the impacts we publish with the ones VoLCA computes from the same source files.
```
cedricr, line 1: @vjousse @paulboosz we should adopt some conventions here…
We’ve got scripts in data/bin, others directly in data, others in some other places (transports…). Some have a shebang, some have not, some have uv dependencies directly in the headers, other have not ; some are executable, others aren’t.

data/bin/compare_volca.py
```diff
@@ -0,0 +1,583 @@
...
+
+One cloud per impact category: a point is one activity, x the impact published in
```
cedricr:
```suggestion
One graph per impact category: a point is one activity, x the impact published in
```
?

data/bin/compare_volca.py
```diff
@@ -0,0 +1,583 @@
...
+    """Read the engine's unit table: units.csv text in, unit name -> (dimension, factor) out."""
+    return {
+        row["name"].lower(): (row["dimension"], float(row["factor"]))
+        for row in csv.DictReader(io.StringIO(table))
+        if row["name"] and not row["name"].startswith("#")
+    }
+
+
+def alias_csv(source: Source, missing: list[str], dep: Catalogue) -> str:
+    """Hand VoLCA Ecobalyse's renames, so this database's inputs link where its own import links them.
```
cedricr:
```suggestion
    """Handles VoLCA Ecobalyse's renames, so this database's inputs link where its own import links them.
```
? I’ve got difficulties understanding this paragraph.

data/bin/compare_volca.py
```diff
@@ -0,0 +1,583 @@
...
+        now = time.monotonic()
+        spent[phase] = now - last
+        last = now
+        # pad over what the announcement wrote, in case this line is the shorter one
+        print(f"{phase} in {spent[phase]:.1f}s".ljust(width), flush=True)
+
+    return doing, done, spent
+
+
+def write(rows: list[dict], version: str, spent: dict[str, float]) -> None:
```
cedricr, line 512: @paulboosz could you give a look at that method ? I *think* you’re generating some matplotlib figures in your score_history method, it would be nice if there’s some consistency on how and where we do that, if that’s applicable (at least the output path, etc.)

cedricr (commented):
Very interesting results!
I’d tend to approve this PR once my comments are adressed, but the code is too complex for me in my current 40° hot brain state 😅
@vjousse @paulboosz the code works as advertised – do you want to give it a go, or is it good enough for you?

pyproject.toml
ccomb: An alternative: 
```
[[tool.ty.overrides]]
include = ["data/bin/compare_volca.py"]
rules = { unresolved-import = "ignore" }
```
Or we can just put it in the `data`group? `matplotlib` takes 25Mb and `pyvolca` is 250kb

data/bin/compare_volca.py
ccomb: A possible convention:
- end-user executable scripts in data/bin with an entry in `.justfile`.
- others are libraries

This is almost the case, except for import*py script and create_activities.
This may be done in a follow-up PR

data/bin/compare_volca.py
ccomb:
```
    """Before linking a database to it's upstream, Ecobalyse renames some of the input names.                                                                        
    VoLCA has to apply the same renames or the inputs stay unlinked.                               
    Returns a source,target CSV, each target spelled as the dependency spells it.                  
    """                                                                                            
```

pyproject.toml
ccomb: Tried that, but it doesn't work all the time: the behaviour differs between local and CI.
(pyvolca stays installed locally after `just compare-volca`, so ty reports the ignore as unused)

ccomb: tried `rules = { unresolved-import = "ignore" }`
Or we can put it in the `data` group which would remove some exceptions?

data/bin/compare_volca.py
ccomb: `score_history.py` doesn't draw anything with matplotlib.

The two places that do draw (PNGs) are `bin/check_ingredient_hierarchy.py` and `export/complements.py`
No one declares matplotlib in the pyproject (should it ?) : One with matplotlib added in the header (`# /// script`) and the other one with indirect dependency through bw2analyzer. Not sure how to align all that.

Good catch on the path though: I was using a relative `Path("output")`. Now a path anchored to the project root.

paulboosz: yep looks good to me

paulboosz (approved):

paulboosz (approved):
Work as expected, LGTM

pyproject.toml
cedricr: Let’s keep it that way, and we’ll probably get rid of the extra groups soon (cc @vjousse)

cedricr (approved):

PR #2785, opened 2026-09-07
fix: always reset processes

vjousse:
## :wrench: Problem

Updating processes on scalingo takes ages and prevent the app from being deployed.

## :cake: Solution

Always remove and then readd all the processes as we don’t need/use components anymore (that could have FK to the processes)

The review app should have processes in the admin. We will need to wait for the prod deploy to test in real conditions.

### DIFF
--- a/backend/src/app/cli/commands.py
+++ b/backend/src/app/cli/commands.py
@@ -420,3 +420,62 @@ def load_processes_json(json_file: click.File) -> None:
 
     console.rule("Loading processes file.")
     anyio.run(_load_processes_json, json_data)
+
+
+async def reset_processes_fixtures(
+    db_session, processes_service, processes_data: list[dict[str, Any]]
+) -> None:
+    """Import/Synchronize Database Fixtures."""
+
+    from structlog import get_logger
+
+    logger = get_logger()
+
+    user = await get_or_create_default_user(db_session)
+
+    for process in processes_data:
+        process["id"] = UUID(process["id"])
+        process["owner"] = user
+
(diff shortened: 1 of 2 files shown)


bin/load-fixtures.sh
```diff
@@ -17,8 +17,11 @@ if [ "$IS_REVIEW_APP" == "true" ]; then
    uv run backend database upgrade --no-prompt
 fi
 
-# Always update the processes with the latest data, even on staging or production
-uv run backend fixtures load-processes public/data/processes_impacts.json
+# Always reset the processes with the latest data, even on production
+# As we don’t use FK to processes anymore, we decided to stop updating
+# the processes and instead just delete and add again
+# (Updates were taking ages and was blocking scalingo deploy)
```
n1k0, line 23: Nit: maybe add a reference or link to this PR number for context

backend/src/app/cli/commands.py
```diff
@@ -420,3 +420,62 @@ async def _load_processes_json(components_data) -> None:
...
+        process["id"] = UUID(process["id"])
+        process["owner"] = user
+
+    existing_processes = await processes_service.get_many()
+    existing_processes_ids = [process.id for process in existing_processes]
+
+    await processes_service.delete_many(
+        item_ids=existing_processes_ids,
+        auto_commit=False,
+    )
```
n1k0: Nit: I'm surely missing good reasons not to do so but I'd naively go the `TRUNCATE TABLE process CASCADE` route here 😅

n1k0 (approved):
🎉

PR #2790, opened 2026-09-08
chore: use batch for inserts

vjousse:
## :wrench: Problem

`main` deploy hook is failing due to memory limitations

## :cake: Solution

Try to batch the inserts and see if it helps

In production? 🫠


n1k0 (approved):
This is really disappointing we have to deploy to production to investigate performance issues at app startup time because of a strange limitation of 512MB of RAM. If this doesn't work, I suggest we reach to support to request more memory at this stage.

PR #2794, opened 2026-09-09
chore: uv version >= 0.12

paulboosz:
## :wrench: Problem

issue #2789's cause is I had an older uv version (0.7)

## :cake: Solution

To avoid this kind of issue, constrain uv version to be at least >= 0.12

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,3 +107,6 @@ extend-select = ["I"]
 extra-paths = [
     "data/",
 ]
+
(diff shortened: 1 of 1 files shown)


vjousse:
Why not requiring the latest > `0.12`?

paulboosz:
> Why not requiring the latest > `0.12`?

indeed, done

vjousse (approved):

PR #2807, opened 2026-09-14
fix: remove token cache

vjousse:
## :wrench: Problem

Optimize security by note caching the email token verification process.

## :cake: Solution

Remove the cache

Nothing, it should be transparent


backend/src/app/lib/crypt.py
```diff
@@ -3,7 +3,6 @@
 import asyncio
 import base64
 
-from asyncstdlib.functools import cache
```
n1k0, line 6: it seems asyncstdlib is not used anywhere anymore, we should probably remove it from deps

backend/src/app/lib/crypt.py
```diff
@@ -38,8 +36,9 @@ async def get_password_hash(password: str | bytes) -> str:
     )
 
 
-@cache
-async def verify_password(plain_password: str | bytes, hashed_password: str) -> bool:
+async def verify_password(
```
n1k0, line 39: What would be the typical impact on performance? argon2 is cpu intensive and I fear that skipping caching entirely may create new scalability issues, especially with api call auth…

vjousse: Well, let’s see? Hard to know before testing it…

n1k0 (approved):
I'm gonna approve acknowledging I don't have a full understanding of all the implications with these changes 😅

PR #2831, opened 2026-09-21
chore: clean obsolete metadata files

paulboosz:
## :wrench: Problem

#2830


ccomb (approved):
Good idea to remove this data here

